PR #5, opened 2015-08-20
Fix broken corosync-shutdown-cleaner.service

nkrinner:
Service file needs absolute paths and also ExecStart has to be set to
something, even if it is not doing anything

### DIFF
--- a/chef/cookbooks/corosync/templates/default/corosync-shutdown-cleaner.service.erb
+++ b/chef/cookbooks/corosync/templates/default/corosync-shutdown-cleaner.service.erb
@@ -5,8 +5,9 @@ Before=<%= @service_name %>.service
 [Service]
 Type=oneshot
 RemainAfterExit=true
-ExecStart=
-ExecStop=rm -f <%= @block_corosync_file %>
+ExecStart=/usr/bin/true
+ExecStop=/usr/bin/rm -f <%= @block_corosync_file %>
+ExecStopPost=/usr/bin/echo "corosync-shutdown-cleaner.service cleaning up <%= @block_corosync_file %> done"
 
 [Install]
 WantedBy=multi-user.target


MaximilianMeister:
I have found a similar issue in https://github.com/crowbar/barclamp-crowbar/pull/1340.
Only the ExecStart i am not sure if it really needs to be specified with a dummy call.

see here:
http://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=

`When Type is not oneshot, only one command may and must be given. When Type=oneshot is used, zero or more commands may be specified`

nkrinner:
I have seen the same documentation piece and was suspecting a systemd bug for a while, as it did not work that way.  But in the end I  think it is just the  online document being wrong. 

The manpage for systemd.service says: `When Type is not oneshot, only one command may be given. When Type=oneshot is used, more  than one command may be specified.` No mention of not using any command. This is also what my testing showed.

MaximilianMeister:
okay. then it is maybe version specific. anyways the ExecStop change is more important here

+1

nkrinner:
For the record: After I made my change, I realized we have a similar fix for ExecStart already in another service: https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/provisioner/files/default/crowbar_notify_shutdown.service

tboerger:
Maybe you want to add an echo as well like @toabctl did for the other pull request?

nkrinner:
@tboerger Added some echo output

chef/cookbooks/corosync/templates/default/corosync-shutdown-cleaner.service.erb
```diff
@@ -5,8 +5,9 @@ Before=<%= @service_name %>.service
 [Service]
 Type=oneshot
 RemainAfterExit=true
-ExecStart=
-ExecStop=rm -f <%= @block_corosync_file %>
+ExecStart=/usr/bin/true
+ExecStop=/usr/bin/rm -f <%= @block_corosync_file %>
+ExecStopPost=/usr/bin/echo "corosync-shutdown-cleaner.service cleaning up <%= @block_corosync_file %>"
```
toabctl: maybe "... cleaning up <%= @block_corosync_file %> done" (or finished). Because it's a Post action the the cleanup is already done.

nkrinner:
@toabctl  updated the echo message

toabctl:
+1

tboerger:
No need to mention the service name, this should be already part of the journal!

PR #12, opened 2015-09-04
crowbar-pacemaker: Only set apache_params if RA is the OCF RA

vuntz:
This does two things:

  - reverse the logic (ie, not checking for systemd RA) since we know
    the params we're looking at are specific to the OCF RA
  - set the SSL-related param only for the OCF RA; it was useless for
    the systemd one, afaict.


aspiers:
+1

PR #18, opened 2015-09-07
[UNTESTED] prevent clusters with an even number of nodes

aspiers:
Clusters must have an odd number of nodes to ensure quorum during partitioning.  2-node clusters are allowed as a special exception for the typical DRBD setup.


crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -460,6 +460,12 @@ def validate_proposal_after_save proposal
     elements = proposal["deployment"]["pacemaker"]["elements"]
     members = elements["pacemaker-cluster-member"] || []
 
+    if members.length % 2 == 0 && members.length > 2
+      validation_error \
+        "Clusters must have an odd number of nodes to ensure quorum during " \
+        "partitioning.  2-node clusters are allowed as a special exception."
```
vuntz: You're old-school. We don't use two spaces at the end of a sentence anymore ;-)

aspiers: Who's "we"?  *I* use it all the time :-p  And it has the significant advantage of allowing my editor to jump forward/backward by one sentence at a time.

vuntz:
Anyway, +1 for the code.

aplanas:
+1 for me, but maybe this needs to be checked also in the UI part (javascript)?

crowbar_framework/app/models/pacemaker_service.rb
rsalevsky: We should use i18n here. ;)

tboerger: > We should use i18n here. ;)

:+1: that's how it should be done for all new code parts.

aspiers: Now uses i18n.  I think I originally submitted this before we used i18n :-)

aspiers:
Updated, but still haven't tested.  Any volunteers? ;-)

PR #19, opened 2015-09-11
haproxy: Decrease tune.maxrewrite from 8192 to 4096

vuntz:
We reached a case again where the headers were too long and haproxy
errored out with 400 without passing the request to the backends.

Instead of increasing tune.bufsize (which is already quite high), we can
decrease tune.maxrewrite to fix it. Upstream says that 1024 is generally
enough, and that big values do create issues with long requests (like
the issue that was hit here). So decrease it a bit to avoid the current
problem.


tboerger:
LGTM

rsalevsky:
+1

PR #20, opened 2015-09-14
corosync: Start corosync-shutdown-cleaner.service on install

vuntz:
If it's not started (just enabled), then the first shutdown will not
execute its stop command, so the block file will be there and the first
boot after installing HA will fail.


toabctl:
+1

aspiers:
+1

PR #22, opened 2015-09-21
Improvements for DRBD

vuntz:
Use variable sync rate config, and enable peer authentication.

### DIFF
--- a/chef/cookbooks/drbd/attributes/default.rb
+++ b/chef/cookbooks/drbd/attributes/default.rb
@@ -1,6 +1,19 @@
+# Comments for the c-* settings come from
+#   https://drbd.linbit.com/users-guide/s-configure-sync-rate.html
+#   https://blogs.linbit.com/p/128/drbd-sync-rate-controller/
+#   https://blogs.linbit.com/p/443/drbd-sync-rate-controller-2
+
+# Set c-plan-ahead to approximately 10 times the RTT; so if ping from one node
+# to the other says 200msec, configure 2 seconds (ie. a value of 20, as the
+# unit is tenths of a second).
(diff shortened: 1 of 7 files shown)


aspiers:
LGTM, why still WIP?

vuntz:
@aspiers because mkcloud fails. It seems the sync slows down everything, and applying neutron took too long. So probably need to tweak the defaults :/

toabctl:
I wonder why we have 20 GB as default size for the drbd block devices. We can speedup the sync by reducing the default size.

vuntz:
> I wonder why we have 20 GB as default size for the drbd block devices. We can speedup the sync by reducing the default size.

Very good point; fixed in https://github.com/SUSE-Cloud/automation/pull/560

toabctl:
> Very good point; fixed in SUSE-Cloud/automation#560

If that's not enough, we could increase the sync rate Temporary in mkcloud. See https://drbd.linbit.com/users-guide/s-configure-sync-rate.html section 6.10.2

vuntz:
Removed WIP; seems good enough after changing defaults.

toabctl:
+1

vuntz:
added link to bugzilla, no code change

toabctl:
still +1

vuntz:
Worked well in local testing.

vuntz:
Note that we can either wait for non-HA gating (which is useless) or HA gating (which will be red because of other issues).

aspiers:
@vuntz https://github.com/SUSE-Cloud/automation/pull/549 will ensure that the gate for this repo has `hacloud=1`.  However we know the HA gate is broken for different reasons, and anyway I am more inclined to trust your local testing. +1

PR #23, opened 2015-09-24
extract mutual ssh code into separate recipe

aspiers:
Aside from the good practice of avoiding overly long recipes, we may need this as a separate reusable recipe when we add `pacemaker_remote` nodes to the cluster.  Also add a comment about `hb_report`.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/mutual_ssh.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/mutual_ssh.rb
@@ -1,0 +1,43 @@
+#
+# Cookbook Name:: crowbar-pacemaker
+# Recipe:: mutual_ssh
+#
+# Copyright 2015, SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Allow mutual ssh connection among cluster members: go through the
+# cluster nodes and remember the public keys so the provisioner can save them.
+
+access_keys = {}
+
+node["provisioner"]["access_keys"].strip.split("\n").each do |key|
+  key.strip!
+  if !key.empty?
+    nodename = key.split(" ")[2]
+    access_keys[nodename] = key
+  end
+end
+
+CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
+  pkey = cluster_node[:crowbar][:ssh][:root_pub_key] rescue nil
+  if !pkey.nil? && cluster_node.name != node.name && !access_keys.values.include?(pkey)
+    access_keys[cluster_node.name] = pkey
+  end
+end
+
+if access_keys.size > 0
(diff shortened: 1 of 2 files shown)


chef/cookbooks/crowbar-pacemaker/recipes/mutual_ssh.rb
```diff
@@ -0,0 +1,48 @@
...
+node["provisioner"]["access_keys"].strip.split("\n").each do |key|
+  key.strip!
+  if !key.empty?
+    nodename = key.split(" ")[2]
+    access_keys[nodename] = key
+  end
+end
+
+CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
+  pkey = nil
```
tboerger: `pkey` is initially anyway `nil`. I think this is better:

```ruby
pkey = if cluster_node[:crowbar][:ssh]
  cluster_node[:crowbar][:ssh][:root_pub_key]
end
```

aspiers: That's a nice trick but slightly obfuscated.  Why is `pkey` initially `nil` anyway?

tboerger: That is just normal ruby behavior, a variable that is not defined before is always `nil`.

aspiers: Sorry, I don't understand your point.  A variable which is not defined is not `nil`, it's simply not defined:

```
$ irb
2.2.1 :001 > pkey
NameError: undefined local variable or method `pkey' for main:Object
        from (irb):1
        from /home/adam/.rvm/rubies/ruby-2.2.1/bin/irb:11:in `<main>'
2.2.1 :002 > 
```

With your `pkey = if ... end`, `pkey` gets assigned `nil` because that's what the `if ... end` block returns.  As I said, it's a nice trick, but it's slightly obfuscated - in 14 years of using Ruby I don't remember seeing that used before, so  think it's slightly friendlier to readers of the code to explicitly assign it to `nil`.

tboerger: > With your pkey = if ... end, pkey gets assigned nil because that's what the if ... end block returns. As I said, it's a nice trick, but it's slightly obfuscated - in 14 years of using Ruby I don't remember seeing that used before, so think it's slightly friendlier to readers of the code to explicitly assign it to nil.

It's not obfuscated, it's obvious and I have seen that quite often in my 8 or 9 years of ruby experience.

tboerger: Even rubocop knows that, Just try my example with an nil in the else branch.

```ruby
pkey = if false
else
  nil
end
```

aspiers: OK it's obvious to you, but it wasn't obvious to me and I'm not a Ruby newbie.  And I'm sure there are people on the team with much less Ruby experience than you or me.

aspiers:
Rebased.

vuntz:
+1

abel-navarro:
CI failed applying ceilometer barclamp, probably not related to this.

abel-navarro:
+1

aspiers:
Rebased again to reduce the risk of gating failures due to [sync drift](https://bugzilla.suse.com/show_bug.cgi?id=935462).

vuntz:
+1

aplanas:
+1

aspiers:
Thanks.  Waiting for the gate to successfully deploy some OpenStack services at least before merging, to make sure it doesn't break anything.

vuntz:
@aspiers it deployed up to ceilometer. Is that good enough? :-)

aspiers:
Yes it is.

PR #27, opened 2015-10-01
Drop cman cookbook

vuntz:
It's not used, so not sure why we ship it.

cc @aspiers


vuntz:
Note that gating fails because it's not able to remove files when building the package.

tboerger:
LGTM

rsalevsky:
+1

aplanas:
The gating fails because some OBS error (or I think so). Relaunched the job (or I think so :)

vuntz:
> The gating fails because some OBS error (or I think so). Relaunched the job (or I think so :)

See my earlier comment; gating doesnt seem to work when removing files.

aplanas:
+1 I do not see any reference remaining for cman

PR #28, opened 2015-10-01
Drop support for hb_gui

vuntz:
This tool doesn't exist anymore on modern systems.

### DIFF
--- a/chef/cookbooks/pacemaker/attributes/default.rb
+++ b/chef/cookbooks/pacemaker/attributes/default.rb
@@ -16,15 +16,6 @@
 case node[:platform_family]
 when "suse"
   default[:pacemaker][:platform][:packages] = %w(pacemaker crmsh)
-
-  # pacemaker-mgmt-client provides hb_gui, which it's useful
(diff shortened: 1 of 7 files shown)


vuntz:
Note that one setting in corosync.conf references it (use_mgmtd). Not sure if we should change it or not. @aspiers?

aspiers:
Not sure.  BTW I'm deeply uncomfortable with the fact that [Travis has been broken on this repo for so many weeks](https://trello.com/c/bsOyEwZZ) :-(

tboerger:
LGTM

aspiers:
Actually this PR isn't covered by the Pacemaker cookbook test suite anyway.

> Note that one setting in corosync.conf references it (use_mgmtd). Not sure if we should change it or not. @aspiers?

I just asked on `#sleha`.  On SLE12 `corosync.conf(5)` doesn't mention the `service {}` section at all.

vuntz:
@aspiers and I just read the code, and realized that the template I was talking about is SLE11 only anyway, so not relevant :-)

So this should be good.

rsalevsky:
LGTM

PR #29, opened 2015-10-04
crowbar-pacemaker: Support String for ordering of order_only_existing

vuntz:
It can be useful to have non-sequential sets of dependencies in
crowbar_pacemaker_order_only_existing, so on top of supporting
definition of ordering as Array, we also need support as String which
lets us use the full syntax of crm.

This requires some pseudo-parsing of the crm syntax to make sure we
don't drop bits that shouldn't be dropped, but also to drop empty sets
after non-existing resources have been removed.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/providers/order_only_existing.rb
+++ b/chef/cookbooks/crowbar-pacemaker/providers/order_only_existing.rb
@@ -24,10 +24,33 @@ def delete_order(name)
 end
 
 action :create do
+  ordering = new_resource.ordering
   # evil command line; there must be a better way to fetch the list of resources
   # unfortunately, "crm_resource --list-raw" doesn't list groups/clones/etc.
   all_resources = %x{crm --display=plain configure show | awk '/^(primitive|group|clone|ms)/ {print $2}'}.split("\n")
-  ordering_for_existing_resources = new_resource.ordering.select { |r| all_resources.include?(r) }
+  case ordering
+  when Array
+    ordering_for_existing_resources = ordering.select { |r| all_resources.include?(r) }
+  when String
+    # Try to ensure the syntax makes sense
+    raise "Sets in ordering cannot be nested." if ordering =~ /\([^\)]*[\(\[\]]/ || ordering =~ /\[[^\]]*[\[\(\)]/
+    # Only keep valid items, including what's valid in the crm syntax, which
+    # is:
+    # - foo ( bar foobar ) xyz
+    # - foo [ bar foobar ] xyz
+    # - foo [ bar foobar sequantial=true ] xyz
+    # - foo [ bar foobar require-all=true ] xyz
+    ordering_array = ordering.split(" ")
+    existing_ordering_array = ordering_array.select do |r|
+      all_resources.include?(r) || %w{( ) [ ]}.include?(r) || r =~ /sequential=/ || r =~ /require-all=/
+    end
+    # Drop empty sets; we don't want something like:
+    #  order Mandatory: foo ( ) bar
+    # It should become:
+    #  order Mandatory: foo bar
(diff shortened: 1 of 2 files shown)


chef/cookbooks/crowbar-pacemaker/providers/order_only_existing.rb
```diff
@@ -24,10 +24,26 @@ def delete_order(name)
...
   # unfortunately, "crm_resource --list-raw" doesn't list groups/clones/etc.
   all_resources = %x{crm --display=plain configure show | awk '/^(primitive|group|clone|ms)/ {print $2}'}.split("\n")
-  ordering_for_existing_resources = new_resource.ordering.select { |r| all_resources.include?(r) }
+  case ordering
+  when Array
+    ordering_for_existing_resources = ordering.select { |r| all_resources.include?(r) }
+  when String
+    # Try to ensure the syntax makes sense
+    raise "Sets in ordering cannot be nested." if ordering =~ /\([^\)]*[\(\[\]]/
+    raise "Sets in ordering cannot be nested." if ordering =~ /\[[^\]]*[\[\(\)]/
```
aspiers: Just use `or` here, no need to duplicate the `raise` statement.

chef/cookbooks/crowbar-pacemaker/providers/order_only_existing.rb
```diff
@@ -24,10 +24,26 @@ def delete_order(name)
 end
 
 action :create do
+  ordering = new_resource.ordering
   # evil command line; there must be a better way to fetch the list of resources
   # unfortunately, "crm_resource --list-raw" doesn't list groups/clones/etc.
   all_resources = %x{crm --display=plain configure show | awk '/^(primitive|group|clone|ms)/ {print $2}'}.split("\n")
-  ordering_for_existing_resources = new_resource.ordering.select { |r| all_resources.include?(r) }
+  case ordering
```
aspiers: Why not

```
ordering_for_existing_resources =
  case ordering
    ...
```

?  Bit more idiomatic Ruby.

chef/cookbooks/crowbar-pacemaker/providers/order_only_existing.rb
```diff
@@ -24,10 +24,26 @@ def delete_order(name)
...
+  when String
+    # Try to ensure the syntax makes sense
+    raise "Sets in ordering cannot be nested." if ordering =~ /\([^\)]*[\(\[\]]/
+    raise "Sets in ordering cannot be nested." if ordering =~ /\[[^\]]*[\[\(\)]/
+    # Only keep valid items, including what's valid in the crm syntax
+    ordering_array = ordering.split(" ")
+    existing_ordering_array = ordering_array.select { |r|
+      all_resources.include?(r) || %w{( ) [ ]}.include?(r) || r =~ /sequential=/ || r =~ /require-all=/
+    }
+    # Drop empty sets
```
aspiers: Without consulting the `crm` man page for exact `order` syntax, I can kind of guess what empty sets means, but it would be nice to make this explicit e.g. by extending the comment so that it gives an example.

aspiers:
Basically LGTM.

chef/cookbooks/crowbar-pacemaker/providers/order_only_existing.rb
vuntz: Well, I find it less idiomatic because of the multiple lines in the `when String`case.

vuntz:
@aspiers other comments addressed

chef/cookbooks/crowbar-pacemaker/providers/order_only_existing.rb
aspiers: OK, just an idea :)

rsalevsky:
+1

PR #30, opened 2015-10-05
reenable Pacemaker cookbook testing in Travis CI

aspiers:
When barclamp-pacemaker was migrated to crowbar-ha, there was a
regression in the Travis CI config which meant that the pacemaker
cookbook's test suite no longer got tested.  This should fix that.


dirkmueller:
+1

PR #32, opened 2015-10-16
Increase default timeout for sync_mark resource

rhafer:
When deploying some of the later barclamps (e.g. heat) the chef-client run on
the founder can be significantly slower than on the non-founder nodes. The
default timeout of 60 seconds for the sync_mark resource can be easily reached
in that case, especially on slower machines.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/resources/sync_mark.rb
+++ b/chef/cookbooks/crowbar-pacemaker/resources/sync_mark.rb
@@ -59,4 +59,4 @@ attribute :mark,      kind_of: String,  default: nil
 attribute :revision,  kind_of: Integer, default: nil
 
 attribute :fatal,     kind_of: [TrueClass, FalseClass], default: true
-attribute :timeout,   kind_of: Integer, default: 60
+attribute :timeout,   kind_of: Integer, default: 120


rhafer:
This should fix failures like: https://ci.suse.de/job/openstack-mkcloud/9852/ (once backported to tex)

rhafer:
Hm, I just saw https://github.com/crowbar/crowbar-openstack/pull/54, but I am not really sure this is only a heat related problem. AFAICS it's not that creating the resource takes so long, but that the chef-client run on the founder node needs much longer to reach the point to actually start creating the resource, because it does a lot more things than the non-founder node(s) before that. I.e. the non-founder not already times out before the founder even reaches the point where it would start creating the heat resources.

So this might not be a heat specific problem. (Though I've yet to see a run where it's not heat that got hit by this. I guess for now I'll close this and backport the other PR to tex.

aspiers:
@rhafer [bsc#935462 (missing synchronization means chef-client runs on cluster nodes drift apart)](https://bugzilla.suse.com/show_bug.cgi?id=935462) explains why this happens.

aspiers:
And it also lists the potential solutions.  Increasing the default timeout might help a bit, but it certainly won't fix the underlying problem, and the downside is that it also means we (and all `mkcloud` jobs) have to wait for longer for the timeout if another problem causes a synchronization failure ([bsc#947044: Install of openstack-dashboard takes way too long](https://bugzilla.suse.com/show_bug.cgi?id=947044) is just a recent one of countless examples of a synchronization failure unrelated to drift).

PR #36, opened 2015-10-27
Added option to setup watchdog with SBD

jsuchome:
Tested: surprisingly worked on the first try

### DIFF
--- a/chef/cookbooks/pacemaker/recipes/stonith.rb
+++ b/chef/cookbooks/pacemaker/recipes/stonith.rb
@@ -38,6 +38,21 @@ when "sbd"
     sbd_cmd += " -d #{Shellwords.shellescape(sbd_device)}"
   end
 
+  watchdog_module = node[:pacemaker][:stonith][:sbd][:watchdog_module]
+
+  execute "Load watchdog module #{watchdog_module}" do
+    command "/sbin/modprobe #{watchdog_module}"
+    not_if { watchdog_module.empty? }
+  end
+
+  file "/etc/modules-load.d/crowbar-watchdog.conf" do
+    content watchdog_module
+    mode 0644
+    owner "root"
+    group "root"
+    not_if { watchdog_module.empty? }
+  end
+
   execute "Check if watchdog is present" do
     command "test -c /dev/watchdog"
   end
@@ -177,4 +192,11 @@ else
   message = "Unknown STONITH mode: #{node[:pacemaker][:stonith][:mode]}."
   Chef::Log.fatal(message)
   raise message
+end
+
+file "delete crowbar-watchdog.conf if not needed" do
+  path "/etc/modules-load.d/crowbar-watchdog.conf"
+  action :delete
+  only_if { node[:pacemaker][:stonith][:mode] != "sbd" ||
+            node[:pacemaker][:stonith][:sbd][:watchdog_module].empty? }
 end
--- a/crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
@@ -18,6 +18,11 @@
       #stonith_sbd_container{ "data-nodes" => node_aliases.to_json }
         .alert.alert-info
           = t(".stonith.sbd.info_html")
+
+        = string_field %w(stonith sbd watchdog_module)
+        %span.help-block
+         = t('.stonith.sbd.watchdog_module_hint')
+
         %table.table.table-middle
           %thead
             %tr
--- a/crowbar_framework/config/locales/pacemaker/en.yml
+++ b/crowbar_framework/config/locales/pacemaker/en.yml
@@ -75,8 +75,10 @@ en:
           mode: 'Configuration mode for STONITH'
           no_nodes: 'No nodes have been assigned a pacemaker role.'
           sbd:
-            info_html: 'Manual configuration is required for SBD: before applying the proposal, you will have to ensure that the devices are available and initialized for SBD; you will also need to setup a watchdog. Refer to the <a href="http://www.linux-ha.org/wiki/SBD_Fencing">Linux HA documentation</a> for details.'
+            info_html: 'Manual configuration is required for SBD: before applying the proposal, you will have to ensure that the devices are available and initialized for SBD. You will also need to manually setup a watchdog if not all nodes use the same watchdog kernel module. Refer to the <a href="http://www.linux-ha.org/wiki/SBD_Fencing">Linux HA documentation</a> for details.'
             name: 'Node name'
+            watchdog_module: 'Kernel module for watchdog'
+            watchdog_module_hint: 'Leave the watchdog module attribute empty if the nodes need different watchdog modules.'
             devices: 'Block devices for node'
             devices_hint: 'Multiple devices can be specified in the comma-separated list. It is advised to use a stable path for devices (with /dev/disk/by-id/, for instance).'
           shared:
--- a/chef/cookbooks/pacemaker/attributes/default.rb
+++ b/chef/cookbooks/pacemaker/attributes/default.rb
@@ -54,6 +54,9 @@ default[:pacemaker][:stonith][:mode] = "disabled"
 #  default[:pacemaker][:stonith][:sbd][:nodes][$node][:devices] = ['/dev/disk/by-id/foo-part1', '/dev/disk/by-id/bar-part1']
 default[:pacemaker][:stonith][:sbd][:nodes] = {}
 
+# kernel module to use for watchdog
+default[:pacemaker][:stonith][:sbd][:watchdog_module] = ""
+
 default[:pacemaker][:stonith][:shared][:agent] = ""
 # This can be either a string (containing a list of parameters) or a hash.
 # For instance:
--- a/chef/data_bags/crowbar/migrate/pacemaker/032_add_sbd_watchdog.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/032_add_sbd_watchdog.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["stonith"]["sbd"]["watchdog_module"] = ta["stonith"]["sbd"]["watchdog_module"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["stonith"]["sbd"].delete("watchdog_module")
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -16,6 +16,7 @@
       "stonith": {
         "mode": "per_node",
         "sbd": {
+          "watchdog_module" : "",
           "nodes": {}
         },
         "shared": {
@@ -53,7 +54,7 @@
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 31,
+      "schema-revision": 32,
       "element_states": {
         "pacemaker-cluster-member" : [ "readying", "ready", "applying" ],
         "hawk-server"              : [ "readying", "ready", "applying" ]
--- a/chef/data_bags/crowbar/template-pacemaker.schema
+++ b/chef/data_bags/crowbar/template-pacemaker.schema
@@ -39,6 +39,7 @@
                   "type": "map",
                   "required": true,
                   "mapping": {
+                    "watchdog_module": { "type": "str", "required": true },
                     "nodes": {
                       "type": "map",
                       "required": true,


jsuchome:
I thought this might have a sense to create from barclamp and not require user to setup before... @aspiers ?

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -38,6 +38,21 @@
...
+    execute "Load watchdog module #{watchdog_module}" do
+      command "/sbin/modprobe #{watchdog_module}"
+    end
+
+    file "/etc/modules-load.d/watchdog.conf" do
+      content watchdog_module
+      mode 0644
+      owner "root"
+      group "root"
+    end
```
vuntz: We probably need to delete that file in `else` (and for other stonith cases).

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -38,6 +38,21 @@
...
   end
 
+  watchdog_module = node[:pacemaker][:stonith][:sbd][:watchdog_module] || ""
+
+  if watchdog_module
+    execute "Load watchdog module #{watchdog_module}" do
+      command "/sbin/modprobe #{watchdog_module}"
+    end
+
+    file "/etc/modules-load.d/watchdog.conf" do
```
vuntz: Maybe name it `crowbar-watchdog.conf` to be sure it won't overwrite anything?

chef/cookbooks/pacemaker/recipes/stonith.rb
jsuchome: done

vuntz: Not sure it matters, but what about doing:

```ruby
file "/etc/modules-load.d/crowbar-watchdog.conf" do
  action :delete
end unless node[:pacemaker][:stonith][:mode] == "sbd" || node[:pacemaker][:stonith][:sbd][:watchdog_module]
```

outside of the big `case`?

vuntz:
+1; there's just one comment about the code duplication (where I'm mostly worried about the fact that we would forget to copy that bit if we add a new stonith device type).

chef/cookbooks/pacemaker/recipes/stonith.rb
jsuchome: Yes, I've thought about that as well, might be better...

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -38,6 +42,25 @@
     sbd_cmd += " -d #{Shellwords.shellescape(sbd_device)}"
   end
 
+  watchdog_module = node[:pacemaker][:stonith][:sbd][:watchdog_module] || ""
+
+  if watchdog_module
```
tboerger: In general it's more conventional to add the `only_if` and `not_if` arguments to the resources instead of wrapping it with `if` or `else`. This is something that would get a complain with `foodcritic`.

tboerger:
LGTM beside my general comment.

chef/cookbooks/pacemaker/recipes/stonith.rb
jsuchome: Done

chef/cookbooks/pacemaker/recipes/stonith.rb
jsuchome: OK

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -167,3 +182,8 @@
   Chef::Log.fatal(message)
   raise message
 end
+
+# delete crowbar-watchdog.conf if not needed
+file "/etc/modules-load.d/crowbar-watchdog.conf" do
+  action :delete
+end unless node[:pacemaker][:stonith][:mode] == "sbd" || node[:pacemaker][:stonith][:sbd][:watchdog_module]
```
tboerger: I think that would fit better into `not_if` :)

tboerger:
Again a single comment, afterwards it's :+1: for me :)

chef/cookbooks/pacemaker/recipes/stonith.rb
aspiers: I agree with @tboerger.

aspiers:
Same for me as what @tboerger said.  Embarrassing that we didn't already do this!

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -167,3 +182,9 @@
   Chef::Log.fatal(message)
   raise message
 end
+
+# delete crowbar-watchdog.conf if not needed
+file "/etc/modules-load.d/crowbar-watchdog.conf" do
```
vuntz: This redefines the resource for this file in case we use sbd, which is not really nice. So we have two options: either rename the resource (and use a `path` attribute), or do not use `not_if` (which goes against what was suggested, but I think it's fine).

crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
```diff
@@ -18,6 +18,9 @@
       #stonith_sbd_container{ "data-nodes" => node_aliases.to_json }
         .alert.alert-info
           = t(".stonith.sbd.info_html")
+
+        = string_field %w(stonith sbd watchdog_module)
```
vuntz: For the record, I remember the reason we didn't have this attribute in the past: in case the module is different for some nodes. We have two options to deal with that here: either change the attribute to be per-node (so we should instead move it to the table), or add some hint that if this the case, people should leave the value empty and configure it themselves.

chef/cookbooks/pacemaker/recipes/stonith.rb
jsuchome: I do not have preference, so maybe that'a a question for chef purists...

crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
jsuchome: Or as a 3rd option, we could just write it in the relevant section of Deployment Guide...?

jsuchome: I've adapted the locale strings, see en.yml

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -38,6 +38,21 @@
     sbd_cmd += " -d #{Shellwords.shellescape(sbd_device)}"
   end
 
+  watchdog_module = node[:pacemaker][:stonith][:sbd][:watchdog_module] || ""
+
+  execute "Load watchdog module #{watchdog_module}" do
+    command "/sbin/modprobe #{watchdog_module}"
+    only_if { watchdog_module }
```
tboerger: I think `not_if { watchdog_module.empty? }` should be safer.

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -38,6 +38,21 @@
...
+  end
+
+  file "/etc/modules-load.d/crowbar-watchdog.conf" do
+    content watchdog_module
+    mode 0644
+    owner "root"
+    group "root"
+    only_if { watchdog_module }
```
tboerger: I think `not_if { watchdog_module.empty? }` should be safer.

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -167,3 +182,9 @@
   Chef::Log.fatal(message)
   raise message
 end
+
+file "delete crowbar-watchdog.conf if not needed" do
+  path "/etc/modules-load.d/crowbar-watchdog.conf"
+  action :delete
+  not_if { node[:pacemaker][:stonith][:mode] == "sbd" && node[:pacemaker][:stonith][:sbd][:watchdog_module] }
```
tboerger: Maybe you want to check the presence of `node[:pacemaker][:stonith][:sbd][:watchdog_module]`.

```ruby
[1] pry(main)> if ""                                                                                                                                                                                                                           
[1] pry(main)*   puts "true"
[1] pry(main)* end  
(pry):6: warning: string literal in condition
true
```

tboerger:
Some tweaks for the `not_if` and `only_if` blocks, ortherwise I'm fine with it.

chef/cookbooks/pacemaker/recipes/stonith.rb
jsuchome: Hm, I guess `only_if { node[:pacemaker][:stonith][:sbd][:watchdog_module].empty? }` should be actually enough at this place

crowbar_framework/config/locales/pacemaker/en.yml
```diff
@@ -75,8 +75,9 @@ en:
           mode: 'Configuration mode for STONITH'
           no_nodes: 'No nodes have been assigned a pacemaker role.'
           sbd:
-            info_html: 'Manual configuration is required for SBD: before applying the proposal, you will have to ensure that the devices are available and initialized for SBD; you will also need to setup a watchdog. Refer to the <a href="http://www.linux-ha.org/wiki/SBD_Fencing">Linux HA documentation</a> for details.'
+            info_html: 'Manual configuration is required for SBD: before applying the proposal, you will have to ensure that the devices are available and initialized for SBD. You will also need to setup a watchdog manually if cluster nodes should use different watchdog module. Refer to the <a href="http://www.linux-ha.org/wiki/SBD_Fencing">Linux HA documentation</a> for details.'
```
vuntz: I suggest `You will also need to setup a watchdog manually if cluster nodes should use different watchdog module` -> `You will also need to manually setup a watchdog if not all nodes use the same watchdog kernel module.`

crowbar_framework/config/locales/pacemaker/en.yml
```diff
@@ -75,8 +75,9 @@ en:
...
             name: 'Node name'
+            watchdog_module: 'Kernel module for watchdog (same for all nodes)'
```
vuntz: With your text above, I don't think we need `same for all nodes`

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -38,6 +38,21 @@
     sbd_cmd += " -d #{Shellwords.shellescape(sbd_device)}"
   end
 
+  watchdog_module = node[:pacemaker][:stonith][:sbd][:watchdog_module] || ""
```
vuntz: Maybe define the variable outside of the `case` and reuse it in the delete action at the end of the file. Right now, you're missing `|| ""` there.

jsuchome: But shouldn't it be always defined? It's added into schema and has `""` as default value

crowbar_framework/config/locales/pacemaker/en.yml
jsuchome: True, I just thought pointing it out again does not hurt

chef/cookbooks/pacemaker/recipes/stonith.rb
vuntz: You're probably right, then we don't need the `|| ""` here ;-)

tboerger:
:+1:

aplanas:
Can you rebase (if possible) and force push to relaunch the gating process?

dirkmueller:
@aplanas actually I retriggered it via https://ci.suse.de/job/cloud-crowbar-testbuild-pr/1650/parameters/

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -167,3 +182,9 @@
   Chef::Log.fatal(message)
   raise message
 end
+
+file "delete crowbar-watchdog.conf if not needed" do
+  path "/etc/modules-load.d/crowbar-watchdog.conf"
+  action :delete
+  only_if { node[:pacemaker][:stonith][:sbd][:watchdog_module].empty? }
```
vuntz: `node[:pacemaker][:stonith][:mode] != "sbd" || ...`

crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
vuntz: Thanks. With the new string, I wonder if we need an explicit "Leave watchdog attribute empty if the nodes need different watchdog modules" hint for the watchdog attribute? At least it feels like it's easy to not see the bit about the watchdog in the long paragraph.

chef/cookbooks/pacemaker/recipes/stonith.rb
jsuchome: You mean for the cases when there could be some `watchdog_module` left after changing stonith mode to something else? (I wonder if changing existing stonith would be that easy...)

crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
jsuchome: Hm, I think I had something like that there and you thought it is excesive when it's mentioned by info_html... OK, I had it as part of the watchdog_module label, not as hint

chef/cookbooks/pacemaker/recipes/stonith.rb
jsuchome: But it doesn't hurt, ok.

chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -167,3 +182,9 @@
   Chef::Log.fatal(message)
   raise message
 end
+
+file "delete crowbar-watchdog.conf if not needed" do
+  path "/etc/modules-load.d/crowbar-watchdog.conf"
+  action :delete
+  only_if { node[:pacemaker][:stonith][:mode] != "sbd" || node[:pacemaker][:stonith][:sbd][:watchdog_module].empty? }
```
jsuchome: I would ignore this, line looks better this way than broken...

bmwiedemann:
test https://ci.suse.de/view/Cloud/view/Builds/job/openstack-mkcloud/14566/
failed with
Committing the crowbar 'default' proposal for 'ceph' failed
and
/opt/dell/bin/barclamp_lib.rb:532:in `eval': Net::ReadTimeout (Net::ReadTimeout)

jsuchome:
Gating is green! @vuntz , what do you think about this?

vuntz:
@jsuchome I like it, but there's one bit (the part where we delete the file if it's not the sbd case) that kind of conflicts with the code move I did in the stonith pacemake-remote PR (not conflicts as in "the two patches conflict" -- that's already the case, but in the "it doesn't look as beautiful anymore" sense). I'm not really sure how to deal with that yet.

But yeah, +1.

jsuchome:
You mean that first part (creating) could be moved to new `sbd.rb` recipe, but deleting would need to state in `stonith.rb` ?

vuntz:
> You mean that first part (creating) could be moved to new sbd.rb recipe, but deleting would need to state in stonith.rb ?

Yes. Which is no big deal, except that it means hard-coding a path in two different files :/

aspiers:
Somehow you confused the hound dog ;-)  I'm still +1

PR #38, opened 2015-11-06
During validation if cluster member is present, look for member, really

jsuchome:


aspiers:
+1

vwallfahrer:
+1

jsuchome:
gating fails in neutron proposal; as this is about pacemaker, could we merge?

PR #39, opened 2015-11-20
Avoid HoundCI bitching about has_key?

dirkmueller:
According to our now master of style, Hash#has_key? is deprecated in
favor of Hash#key?. (https://github.com/bbatsov/ruby-style-guide#hash-key)

### DIFF
--- a/chef/cookbooks/pacemaker/recipes/stonith.rb
+++ b/chef/cookbooks/pacemaker/recipes/stonith.rb
@@ -100,7 +100,7 @@ when "shared"
     raise message
   end
 
-  unless primitive_params.has_key?("hostlist")
+  unless primitive_params.key?("hostlist")
     message = "Missing hostlist parameter for shared fencing agent!"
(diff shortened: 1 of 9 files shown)


chef/cookbooks/pacemaker/recipes/stonith.rb
```diff
@@ -144,7 +144,7 @@
     # Only set one of hostname / hostlist param if none of them are present; we
     # do not overwrite it as the user might have passed more information than
     # just the hostname (some agents accept hostname:data in hostlist)
-    unless primitive_params.has_key?("hostname") || primitive_params.has_key?("hostlist")
+    unless primitive_params.key?("hostname") || primitive_params.has_key?("hostlist")
```
tboerger: You missed one here :)

tboerger:
Some other offenses but :+1:

rsalevsky:
:+1:

PR #41, opened 2015-11-21
Add pacemaker_transaction LWRP

aspiers:
Currently this is only tested via rspec, but submitted for early review in lieu of the originally planned design meeting.


vuntz:
@aspiers cool. I suggest you split the first 4 commits (and maybe some others) in another PR that we could merge relatively quickly; this would simplify the review of the transaction bit...

aspiers:
@vuntz done in #42.

chef/cookbooks/pacemaker/providers/transaction.rb
```diff
@@ -0,0 +1,57 @@
...
+action :commit_new do
+  chef_resource_names = new_resource.cib_objects
+
+  chef_resources = chef_resource_names.map do |chef_resource_name|
+    run_context.resource_collection.lookup(chef_resource_name)
+  end
+
+  # The transaction should only create CIB objects which don't already exist.
+  chef_resources_to_create = chef_resources.select do |chef_resource|
+    ! ::Pacemaker::CIBObject.exists?(chef_resource.name)
```
vuntz: How would we update a resource if it's part of a transaction? Because here, we're filtering the resource out.

(Maybe this is obvious if you have an example of how this will work in a concrete example :-))

vuntz:
Btw, it feels like most of the changes are unrelated cleanups; we can probably merge these faster if you split them in another PR.

chef/cookbooks/pacemaker/providers/transaction.rb
aspiers: That's why I'm adding `action :update` to all the other LWRPs, as mentioned in the last 2 standups ;-)

```ruby
pacemaker_primitive "keystone" do
  ...
  action :update
end

pacemaker_clone "cl-keystone" do
  rsc "keystone"
  action :update
end

pacemaker_transaction "keystone clone" do
  cib_objects [
    "pacemaker_primitive[keystone]",
    "pacemaker_clone[cl-keystone]",
  ]
  action :commit_new  # this is the default and only action, so this line is optional
end
```

aspiers:
OK then, let's do #44 first.

chef/cookbooks/pacemaker/providers/transaction.rb
vuntz: Thanks, now I get it. And I was listening in the standups, but was still not clear :-)

aspiers:
Superseded by #46.

PR #45, opened 2015-12-02
Increased sync_mark timeout from 60 to 180 seconds

abel-navarro:
It has been observed that non-founder members of the cluster do timeout
when waiting for founder members to complete their tasks. This patch
increases that timeout.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
@@ -57,7 +57,7 @@ require "timeout"
 module CrowbarPacemakerSynchronization
 
   # See "Synchronization helpers" documentation
-  def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout = 60)
+  def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout = 180)
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
     return if CrowbarPacemakerHelper.is_cluster_founder?(node)
 


jsuchome:
So just increasing timeout is good enough? I'd be fine with that, although, isn't there a way which we could make the whole process better?

abel-navarro:
This is a bit new to me but I'd say maybe with an event driven orchestration, but for now we just have active waits.

aspiers:
Please can you alter the commit message to reference [bsc#935462 (missing synchronization means chef-client runs on cluster nodes drift apart)](https://bugzilla.suse.com/show_bug.cgi?id=935462) (and BTW I generated that hyperlink using [my userscripts hack](https://github.com/aspiers/desktop-config/blob/master/lib/browser/userscripts/page-id-helpers/README.md) ;-) )

Also I wonder if it might be better if we change the default to 120 first, but I'm not sure.  The longer the timeout, the longer our CI jobs will take to fail when any Chef operation fails.  Having said that, this only affects HA clouds which are all failing anyway, so I guess 180 is OK.  @vuntz what do you think?

The medium-term solution is already planned by @vuntz - the idea is to use the new role recipes to only run the recipes which are required for applying the current proposal.  Then each chef-client run will be much *much* shorter and there will be no opportunity for node synchronization to drift apart like it does currently.

abel-navarro:
@aspiers the timeout I observed yesterday was about 82 seconds. 120 seconds sound reasonable. Anyway, this PR sets the timeout to 180 and yet the CI fails with the same error. I'm doubting the patch was applied correctly in the CI but cannot see anything wrong in CI output. So I may need to look somewhere else for this timeout problem.

abel-navarro:
updated the commit message to include a reference to the bug. This also retriggers the CI, so I may a a chance to look into it for debugging purposes.

aspiers:
This is failing due to a race condition in neutron server startup.  @vuntz @dirkmueller I propose we merge this since we know for sure it eliminates or at least greatly reduces the occurrence of a common problem.

vuntz:
+1

aplanas:
So stay in 180 instead of 120?

+1

dirkmueller:
+1

aspiers:
I just realised this PR doesn't actually make any difference :-(  It needs to be changed [here](https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/resources/sync_mark.rb#L62), since the default comes from the resource definition and the default values in the method definitions never get used.

PR #46, opened 2015-12-03
Add pacemaker_transaction LWRP and :update action

aspiers:
Add support for creating multiple objections in the Pacemaker CIB via a single transaction, and a corresponding new `:update` action for the other existing Pacemaker LWRPs which only updates existing resource and never creates new ones.  The combination of these two new features allows us to implement automatic deployment of compute node HA, and also makes it possible to clean up some really ugly long-standing hacks in the way we set up `clone`, `ms`, and similar objects.

Here is an example usage:

```ruby
pacemaker_primitive "keystone" do
  ...
  # this will never create anything, only converge an existing primitive to the right parameters
  action :update 
end

pacemaker_clone "cl-keystone" do
  rsc "keystone"
  # this will never create anything, only converge an existing primitive to the right parameters
  action :update
end

pacemaker_transaction "keystone clone" do
  cib_objects [
    "pacemaker_primitive[keystone]",
    "pacemaker_clone[cl-keystone]",
  ]
  # Create any of the objects in the above Array which don't yet exist.
  # :commit_new is the default and only action, so this line is optional.
  action :commit_new
end
```


chef/cookbooks/pacemaker/providers/transaction.rb
```diff
@@ -0,0 +1,57 @@
...
+  chef_resource_names = new_resource.cib_objects
+
+  chef_resources = chef_resource_names.map do |chef_resource_name|
+    run_context.resource_collection.lookup(chef_resource_name)
+  end
+
+  # The transaction should only create CIB objects which don't already exist.
+  chef_resources_to_create = chef_resources.select do |chef_resource|
+    ! ::Pacemaker::CIBObject.exists?(chef_resource.name)
+  end
```
vuntz: I'd maybe change the `select` to a `reject` to avoid the negation in the test. But only change this if you agree it's better (since I don't really care much ;-))

chef/cookbooks/pacemaker/providers/transaction.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  cib_objects_to_create = chef_resources_to_create.map do |chef_resource|
+    cib_object_class =
+      chef_resource.provider_for_action(:nothing).cib_object_class
+    cib_object_class.from_chef_resource(chef_resource)
+  end
+
+  if chef_resources_to_create.any?
```
vuntz: `return if chef_resources_to_create.empty?` (to avoid the block)

But again, no big deal, feel free t keep the current way.

vuntz: Actually, if we do the change, we could move it before the block that you have before (which would lead to amazing performance improvements ;-))

chef/cookbooks/pacemaker/providers/location.rb
```diff
@@ -33,6 +33,10 @@
   end
 end
 
+action :update do
+  maybe_modify_resource(new_resource.name)
```
vuntz: Why not `standard_update_action`?

chef/cookbooks/pacemaker/providers/colocation.rb
```diff
@@ -33,6 +33,10 @@
   end
 end
 
+action :update do
+  maybe_modify_resource(new_resource.name)
```
vuntz: Same question: why not `standard_update_action`?

chef/cookbooks/pacemaker/providers/transaction.rb
aspiers: I agree, changed.

chef/cookbooks/pacemaker/providers/transaction.rb
aspiers: Oh wow awesome, now we can deploy all our clouds in -10 minutes! ;-)

Good idea - I've changed it, although it had to be `next` not `return` since it was in a block.

chef/cookbooks/pacemaker/providers/location.rb
aspiers: At a guess I was probably coding it at 2.30am or something.  Changed.

chef/cookbooks/pacemaker/providers/colocation.rb
aspiers: Ditto.

vuntz:
+1

vuntz:
Gating failed while deploying nova.

aspiers:
Rebased to include longer nova_database timeout from https://github.com/crowbar/crowbar-openstack/issues/159

vuntz:
Confirming that this works well (at least up to ceilometer).

vuntz:
(so still +1)

aspiers:
@vuntz Are we realistically going to get any other reviews? I think you are the only other person who understands this code area ...

dirkmueller:
LGTM

PR #47, opened 2015-12-07
pacemaker: Symlink the suse templates as opensuse

vuntz:
We want the same set of templates for both SUSE and openSUSE.


dirkmueller:
:+1:

aspiers:
+1 once gating passes.

tboerger:
It should be easier to put it simply into default ;)

PR #49, opened 2015-12-16
really increase sync_mark timeout to 300s (bsc#935462)

aspiers:
d2c1a01 from #45 aimed to do this but it only changed one method's default value, and both methods' defaults are actually overridden by the default value from the LWRP resource definition.  However we make them all 180s for consistency, in the unlikely case anything else should call the methods directly in future.

https://bugzilla.suse.com/show_bug.cgi?id=935462

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
@@ -57,7 +57,7 @@ require "timeout"
 module CrowbarPacemakerSynchronization
 
   # See "Synchronization helpers" documentation
-  def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout = 180)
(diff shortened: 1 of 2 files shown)


aplanas:
Is https://github.com/crowbar/crowbar-ha/pull/45 going to be closed in favor of this?

vuntz:
+1

aspiers:
@aplanas #45 was merged 2 days ago.

tboerger:
Gating seems to be broken with the new client. Looks like we have to increase the timeout value of the client.

jsuchome:
+1

aplanas:
+1

vuntz:
@aspiers with the move to transactions, I see a single chef-client taking 330s here. Should we maybe bump this to 300s directly?

aspiers:
@vuntz done.

abel-navarro:
tried manually with a 180s timeout and mkcloud ha-deployment succeeded, while with the previous value (60s) was failing.

PR #52, opened 2015-12-18
Add new pacemaker-remote role to proposal

vuntz:
Depends on @aspiers' pacemaker-remote branch (no PR yet).


aspiers:
Thanks - there are actually 2 new roles but I incorporated your changes into #53 which now supersedes this.

PR #53, opened 2015-12-23
automate pacemaker_remote setup

aspiers:
This is moderately tested and seems to work but more testing wouldn't hurt.  Supersedes #52.  Notice that there are two new roles:

- `pacemaker-remote`, which needs to be assigned to remote nodes
- `pacemaker-remote-delegator`, which needs to be assigned to corosync nodes (although this should happen invisibly via `pacemaker_service.rb`)

A fuller explanation is included in the commit messge which introduces them - highly recommended reading.

Here's a recipe call tree diagram I scribbled in order to ensure I was making the right changes.  Good luck trying to understand that ... Here are a few hints though:

- `c-p` refers to the `crowbar-pacemaker` cookbook.  Similarly `p::foo` refers to the `foo` recipe in the `pacemaker` cookbook.
- Chef roles are in green.
- Known issues or warnings are in red.
- Orchestration / synchronization / ordering is in orange.
- I have since renamed the `remote-resources` role to `remote-delegator`.

pacemaker_recipes_-_page_1

### DIFF
--- a/chef/cookbooks/corosync/recipes/authkey.rb
+++ b/chef/cookbooks/corosync/recipes/authkey.rb
@@ -25,7 +25,7 @@ if Chef::Config[:solo]
 end
 
 cluster_name = node[:corosync][:cluster_name]
-unless cluster_name and ! cluster_name.empty?
+unless cluster_name && !cluster_name.empty?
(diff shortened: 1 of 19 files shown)


vuntz:
Moved to https://github.com/crowbar/crowbar-ha/pull/55

PR #54, opened 2015-12-30
allow CIB objects with arbitrary unparsed definition strings

aspiers:
Until now, CIB objects were only instantiated by supplying individual fine-grained attributes, e.g. in the case of `location` constraints, the attributes are `name`, `rsc`, `score`, and `node`.

However CIB objects can actually have more complicated structures, and we need to support some of them, such as rule-based constraints like:

    location rulelocation1 primitive1 resource-discovery=exclusive \
        rule 0: OpenStack-role eq controller

In fact `crmsh` even supports `location` constraints with far more complicated syntax than this.  So rather than attempting to reimplement the entire `crmsh` parser, we just add support for objects which are defined by arbitrary definition strings.  The only bit we actually need to parse is the first two words, i.e. the object's type and name respectively.

To maintain backwards compatibility with callers which still want to construct objects in the fine-grained manner, we add an internal `@authority` instance variable which determines whether the source of authority for the object's definition is the fine-grained attributes or the arbitrary unparsed string.  When a subclass's `#parse_definition` method succeeds in parsing the definition string into fine-grained attribute values, it must invoke a `#attrs_authoritative` method newly defined in the base class which sets `@authority` to `:attributes`. Otherwise the definition will be retrieved directly from the raw definition string.  This new mechanism replaces `#definition_string`, which was a semantic duplication of `#definition`.

We also remove `#load_definition`, which was a hack only used by tests, and instead do tests using the more real-world `.from_name` API.

### DIFF
--- a/chef/cookbooks/pacemaker/libraries/pacemaker/constraint/colocation.rb
+++ b/chef/cookbooks/pacemaker/libraries/pacemaker/constraint/colocation.rb
@@ -26,16 +26,19 @@ class Pacemaker::Constraint::Colocation < Pacemaker::Constraint
     # FIXME: this is incomplete.  It probably doesn't handle resource
     # sets correctly, and certainly doesn't handle node attributes.
     # See the crm(8) man page for the official BNF grammar.
-    unless definition =~ /^#{self.class.object_type} (\S+) (\d+|[-+]?inf): (.+?)\s*$/
+    # It can be fixed in the same way location has been fixed.
+    unless @definition =~
+        /^#{self.class.object_type} (\S+) (\d+|[-+]?inf): (.+?)\s*$/
       raise Pacemaker::CIBObject::DefinitionParseError, \
             "Couldn't parse definition '#{definition}'"
     end
-    self.name  = $1
+    self.name = $1
     self.score = $2
     self.resources = $3
+    attrs_authoritative
   end
 
-  def definition_string
+  def definition_from_attributes
     "#{self.class.object_type} #{name} #{score}: #{resources}"
   end
 end
--- a/chef/cookbooks/pacemaker/libraries/pacemaker/constraint/colocation.rb
+++ b/chef/cookbooks/pacemaker/libraries/pacemaker/constraint/colocation.rb
@@ -13,7 +13,7 @@ class Pacemaker::Constraint::Colocation < Pacemaker::Constraint
     when String
       @resources = val
     else
-      raise "Tried to set resources attribute for colocation '#{name}' " +
+      raise "Tried to set resources attribute for colocation '#{name}' " \
         "to invalid type #{val.class} (#{val.inspect})"
     end
   end
--- a/chef/cookbooks/pacemaker/libraries/pacemaker/resource/primitive.rb
+++ b/chef/cookbooks/pacemaker/libraries/pacemaker/resource/primitive.rb
@@ -21,24 +21,26 @@ class Pacemaker::Resource::Primitive < Pacemaker::Resource
   end
 
   def parse_definition
-    unless definition =~ /\A#{self.class.object_type} (\S+) (\S+)/
+    unless @definition =~ /\A#{self.class.object_type} (\S+) (\S+)/
       raise Pacemaker::CIBObject::DefinitionParseError, \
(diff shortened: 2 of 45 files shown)


chef/cookbooks/pacemaker/spec/helpers/cib_object.rb
```diff
@@ -21,14 +21,24 @@ def expect_to_match_fixture(obj)
...
-      raise_error(Pacemaker::CIBObject::TypeMismatch,
-                  "Expected #{object_type} type but loaded definition was type sometype")
+    expect { Pacemaker::CIBObject.from_name(fixture.name) }.to raise_error(
+      RuntimeError,
+      "No subclass of Pacemaker::CIBObject was registered with type 'sometype'"
+    )
+  end
+
+  it "should barf if the definition's type is not right" do
+    expect {
```
aspiers: Reported as https://github.com/bbatsov/ruby-style-guide/issues/522

chef/cookbooks/pacemaker/resources/location.rb
```diff
@@ -24,4 +24,5 @@
 attribute :name,  kind_of: String, name_attribute: true
 attribute :rsc,   kind_of: String
 attribute :score, kind_of: String
-attribute :node,  kind_of: String
+attribute :lnode, kind_of: String
```
vuntz: Because of this change, you need to update chef/cookbooks/pacemaker/recipes/stonith.rb to use `lnode`.

chef/cookbooks/pacemaker/libraries/pacemaker/constraint/colocation.rb
```diff
@@ -30,12 +30,12 @@ def parse_definition
       raise Pacemaker::CIBObject::DefinitionParseError, \
             "Couldn't parse definition '#{definition}'"
     end
-    self.name  = $1
+    self.name = $1
     self.score = $2
     self.resources = $3
   end
```
vuntz: Missing call to `attrs_authoritative` here?

chef/cookbooks/pacemaker/libraries/pacemaker/resource/primitive.rb
```diff
@@ -21,24 +21,26 @@ def self.attrs_to_copy_from_chef
   end
 
   def parse_definition
-    unless definition =~ /\A#{self.class.object_type} (\S+) (\S+)/
+    unless @definition =~ /\A#{self.class.object_type} (\S+) (\S+)/
```
vuntz: I'm a bit confused: why do we change to `@definition` here and not in the other resources / constraints?

chef/cookbooks/pacemaker/resources/location.rb
aspiers: Yeah, probably no easy way to make this backwards-compatible :-(

chef/cookbooks/pacemaker/libraries/pacemaker/constraint/colocation.rb
aspiers: Good catch, will fix.

chef/cookbooks/pacemaker/libraries/pacemaker/resource/primitive.rb
aspiers: No good reason; just because I didn't add support for arbitrary unparsed definition strings to anything yet except for `location` (doing the others will require some fixture additions/modifications).  It's more correct to use `@definition` to avoid accidentally parsing a string which is auto-constructed from attributes originating from an already parsed definition, but I'm not sure this corner case is actually possible.  However I'll change for consistency.

chef/cookbooks/pacemaker/resources/location.rb
aspiers: Oh, you mean in this repo.  Gotcha, will fix.  `crowbar-openstack` repo will need corresponding fixes too though :-(

vuntz: There's nothing in the other repos using this, iirc, so we're lucky ;-)

vuntz:
Moved to https://github.com/crowbar/crowbar-ha/pull/56

PR #55, opened 2016-01-05
automate pacemaker_remote setup (updated version)

vuntz:
This is https://github.com/crowbar/crowbar-ha/pull/53 with some fixes.

### DIFF
--- a/chef/cookbooks/corosync/recipes/authkey.rb
+++ b/chef/cookbooks/corosync/recipes/authkey.rb
@@ -25,7 +25,7 @@ if Chef::Config[:solo]
 end
 
 cluster_name = node[:corosync][:cluster_name]
-unless cluster_name and ! cluster_name.empty?
+unless cluster_name && !cluster_name.empty?
   Chef::Application.fatal! "Couldn't figure out corosync cluster name"
   return
 end
--- a/chef/cookbooks/corosync/recipes/authkey.rb
+++ b/chef/cookbooks/corosync/recipes/authkey.rb
@@ -34,6 +34,7 @@ end
 query  = "chef_environment:#{node.chef_environment}"
 query += " AND corosync_cluster_name:#{cluster_name}"
 
+# FIXME: move this Crowbar-specific code into the crowbar-pacemaker cookbook
 is_crowbar = !(node[:crowbar].nil?)
 authkey_node = nil
 
--- a/chef/cookbooks/corosync/recipes/authkey.rb
+++ b/chef/cookbooks/corosync/recipes/authkey.rb
@@ -17,15 +17,13 @@
 # limitations under the License.
 #
 
-require "base64"
-
 if Chef::Config[:solo]
   Chef::Application.fatal! "This recipe uses search. Chef Solo does not support search."
   return
 end
 
 cluster_name = node[:corosync][:cluster_name]
-unless cluster_name && !cluster_name.empty?
+if cluster_name.nil? || cluster_name.empty?
   Chef::Application.fatal! "Couldn't figure out corosync cluster name"
   return
 end
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -126,7 +126,7 @@ module CrowbarPacemakerHelper
   #
   # This call signature only makes sense because it is not possible
   # for a node to be in multiple clusters.
-  def self.cluster_nodes(node, role = "*")
+  def self.cluster_nodes(node, role = "pacemaker-cluster-member")
     if cluster_enabled?(node)
       server_nodes = []
       # Sometimes, chef-server is a little bit outdated and doesn't have the
@@ -134,7 +134,7 @@ module CrowbarPacemakerHelper
       # actually part of the cluster; we also want to make sure that we include
       # the latest bits with latest attributes for this node, so we always
       # manually add it, instead of relying on the search for this one.
-      Chef::Search::Query.new.search(:node, "roles:#{role || "*"} AND pacemaker_config_environment:#{node[:pacemaker][:config][:environment]}") do |o|
+      Chef::Search::Query.new.search(:node, "roles:#{role || "pacemaker-cluster-member"} AND pacemaker_config_environment:#{node[:pacemaker][:config][:environment]}") do |o|
         server_nodes << o if o.name != node.name
       end
       server_nodes << node if (role.nil? || role == "*" || node.roles.include?(role))
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -126,15 +126,20 @@ module CrowbarPacemakerHelper
   #
   # This call signature only makes sense because it is not possible
   # for a node to be in multiple clusters.
-  def self.cluster_nodes(node, role = "pacemaker-cluster-member")
+  def self.cluster_nodes(node, role = nil)
     if cluster_enabled?(node)
+      role ||= "pacemaker-cluster-member"
       server_nodes = []
+      env = node[:pacemaker][:config][:environment]
       # Sometimes, chef-server is a little bit outdated and doesn't have the
       # latest information, including the fact that the current node is
       # actually part of the cluster; we also want to make sure that we include
       # the latest bits with latest attributes for this node, so we always
       # manually add it, instead of relying on the search for this one.
-      Chef::Search::Query.new.search(:node, "roles:#{role || "pacemaker-cluster-member"} AND pacemaker_config_environment:#{node[:pacemaker][:config][:environment]}") do |o|
+      Chef::Search::Query.new.search(
+        :node,
+        "roles:#{role} AND pacemaker_config_environment:#{env}"
+      ) do |o|
         server_nodes << o if o.name != node.name
       end
       server_nodes << node if (role.nil? || role == "*" || node.roles.include?(role))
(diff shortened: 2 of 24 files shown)


chef/cookbooks/corosync/recipes/authkey.rb
```diff
@@ -25,7 +25,7 @@
 end
 
 cluster_name = node[:corosync][:cluster_name]
-unless cluster_name and ! cluster_name.empty?
+unless cluster_name && !cluster_name.empty?
```
jsuchome: `if cluster_name.nil? || cluster_name.empty?` looks nicer

chef/cookbooks/crowbar-pacemaker/recipes/pacemaker_authkey.rb
```diff
@@ -0,0 +1,72 @@
...
+#
+
+# This recipe is intended for use on *all* cluster nodes, i.e. both
+# corosync and remote nodes.  Only the founder generates an authkey
+# for communication with remote nodes, and all the others copy it;
+# thefore it must be run on the corosync nodes first.  Crowbar
+# achieves this by first running it via the pacemaker-cluster-member
+# role, and then later by the pacemaker-remote role.
+
+require "base64"
```
jsuchome: Maybe move this require to both authkey_generator.rb and authkey_writer.rb? 
I understand that they are included in this one but here it looks bit confusing when not used in the recipe...

chef/cookbooks/crowbar-pacemaker/recipes/remote.rb
```diff
@@ -0,0 +1,47 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# The parent recipe to be run on all remote nodes.
+
+# Figure out which cluster we're tied to, so that we can find the founder
+# and hence the authkey.
+node[:corosync][:cluster_name] = CrowbarPacemakerHelper.cluster_name(node)
```
jsuchome: Isn't `node.save` needed?

chef/cookbooks/pacemaker/recipes/remote.rb
```diff
@@ -0,0 +1,66 @@
...
+if node[:pacemaker][:platform][:remote_packages].nil?
+  Chef::Application.fatal! "FIXME: #{node.platform} platform not supported yet"
+end
+
+node[:pacemaker][:platform][:remote_packages].each do |pkg|
+  package pkg
+end
+
+service "pacemaker_remote" do
+  action [:disable, :start]
```
jsuchome: Is it correct to have the service disabled? 
(Maybe because we are only handling it manually? Comment here would help...)

chef/cookbooks/crowbar-pacemaker/recipes/remote_delegator.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  params_hash = remote[:pacemaker][:remote][:params].to_hash
+  params_hash["server"] = remote[:fqdn]
+
+  pacemaker_primitive "remote-#{remote[:hostname]}" do
+    agent remote[:pacemaker][:remote][:agent]
+    op remote[:pacemaker][:remote][:op].to_hash
+    params params_hash
+    action [:create, :start]
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
jsuchome: This should not be needed, see the first non-comment line of the recipe...

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -116,7 +116,7 @@ def self.cluster_nodes(node, role = "*")
       # actually part of the cluster; we also want to make sure that we include
       # the latest bits with latest attributes for this node, so we always
       # manually add it, instead of relying on the search for this one.
-      Chef::Search::Query.new.search(:node, "roles:#{role || "*"} AND pacemaker_config_environment:#{node[:pacemaker][:config][:environment]}") do |o|
+      Chef::Search::Query.new.search(:node, "roles:#{role || "pacemaker-cluster-member"} AND pacemaker_config_environment:#{node[:pacemaker][:config][:environment]}") do |o|
```
jsuchome: Do we need to provide this backup value when the default role value is set above?

vuntz: It's just in case someone sets the role to `nil` (which should work like the default, imho)

chef/cookbooks/crowbar-pacemaker/recipes/remote.rb
vuntz: Actually, no: it's only needed while we do this run (so, not needed in search).

That being said: there's a save later on in the file (in the `ruby_block`) which will save anyway.

chef/cookbooks/pacemaker/recipes/remote.rb
vuntz: @aspiers wrote this in his docs:

> Later we'll probably want to disable this and start it up via crowbar_join. Actually, even now it's maybe worth disabling, in order to avoid problems with the compute node recovering too quickly after fencing

Although I'm a bit unclea about it, because there's no known issue for this (afaics). Let's wait for @aspiers

chef/cookbooks/crowbar-pacemaker/recipes/remote_delegator.rb
vuntz: You're right. However, we're doing a lot of copy & paste for the HA bits, and I feel a bit safer keeping it so that it's not forgotten in case this is pasted elsewhere.

Do you have a strong feeling about this?

vuntz:
Added some commits to address comments.

chef/cookbooks/crowbar-pacemaker/recipes/remote_delegator.rb
jsuchome: No, I do not mind

jsuchome:
+1 (not tested, based on the code)

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
aspiers: This line is WAY too long, requiring use of the horizontal scroll bar to review in github.  Please fix!

aspiers: > It's just in case someone sets the role to nil (which should work like the default, imho)

In that case you could avoid duplicating the hardcoded default value by changing the signature to `role = nil` and then doing `role ||= "pacemaker-cluster-member"` at the top of the method.  This would also help with shrinking the size of this monster line ;-)

chef/cookbooks/crowbar-pacemaker/recipes/remote_delegator.rb
aspiers: The amount of copy & paste worries me.  What can we do to avoid that?

chef/cookbooks/pacemaker/recipes/remote.rb
aspiers: I realised after writing the ["recovering too quickly" thing](https://github.com/SUSE/cloud/blob/master/docs/dev/HA-for-compute-nodes/manual-setup.md#on-all-compute-nodes-1) that should not be an issue, because the `NovaCompute` RA is supposed to wait for fencing to complete before it starts `nova-compute`.  However there is (or at least was) [a known issue with this](https://github.com/SUSE/cloud/blob/master/docs/dev/HA-for-compute-nodes/known-issues.md).

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
vuntz: @aspiers better now?

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -117,15 +117,18 @@ def self.cluster_haproxy_vpublic_name(node)
...
       server_nodes = []
+      env = node[:pacemaker][:config][:environment]
       # Sometimes, chef-server is a little bit outdated and doesn't have the
       # latest information, including the fact that the current node is
       # actually part of the cluster; we also want to make sure that we include
       # the latest bits with latest attributes for this node, so we always
       # manually add it, instead of relying on the search for this one.
-      Chef::Search::Query.new.search(:node, "roles:#{role || "*"} AND pacemaker_config_environment:#{node[:pacemaker][:config][:environment]}") do |o|
+      Chef::Search::Query.new.search(:node,
+                                     "roles:#{role} AND pacemaker_config_environment:#{env}") do |o|
```
aspiers: Well at least I can see all the code now without using the horizontal scroll bar :-)  But IMHO this is less ugly:

    Chef::Search::Query.new.search(
      :node, "roles:#{role} AND pacemaker_config_environment:#{env}"
    ) do |o|

because it doesn't require your eye to travel so far right in order to find the `do |o|`.

You could even also split after `:node,`.  But I'll leave it up to you ...

vuntz: There you go :-)

chef/cookbooks/crowbar-pacemaker/recipes/remote_delegator.rb
vuntz: I'm talking about copy & paste of pacemaker resources when we create them. I can't think of anything to reduce this...

vuntz:
Had to rebase due to merge conflict, hopefully gating will be green...

jsuchome:
Still +1 after later commits.

nkrinner:
+1

aspiers:
In an ideal world this would be squashed, since 26e2f4f is just undoing what 556744d and e9f7811 previously did.  But this PR has already dragged on too long, and it's good enough for me, so merging.

PR #57, opened 2016-01-05
crowbar-pacemaker: Add num_corosync_nodes helper

vuntz:
This can be used when creating clones to set the clone-max metadata.


aplanas:
+1

jsuchome:
+1

vuntz:
Gating fails during testing, which is totally unrelated to this PR (since it's just adding a new method that is not used yet). So merging.

PR #60, opened 2016-01-07
crowbar: Workaround to avoid the drift caused by sync marks (bsc#935462)

vuntz:
Because each chef-client runs executes everything in all cookbooks, and because
the crm calls slow things down considerably, we have a drift issue where the
founder (which does much more work since it's the only one doing the crm calls)
goes much slower than the other nodes. This can impact things badly when it's
so slow that it triggers the timeout in the sync marks.

The workaround here is to reset the sync marks on the founder, so that all
nodes wait on the sync marks and the drift is only happening between sync
marks.

The goal is to only do this when applying a proposal, so that other chef-client
runs are not blocked waiting for sync marks.

Unfortunately, there are cases where we cannot do that: for swift, it's
likely that swift-ring-compute would be on a cluster node, and since
that role is run first, the cluster node would be alone in running chef
and blocking forever. This means that this small helper here will have
to be called explicitly by all barclamps for which there's no such issue
-- afaict, swift is the only one with an issue.

All of this should be removed when we get proper crowbar orchestration.

https://bugzilla.suse.com/show_bug.cgi?id=935462

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service_object.rb
+++ b/crowbar_framework/app/models/pacemaker_service_object.rb
@@ -133,6 +133,37 @@ class PacemakerServiceObject < ServiceObject
     [elements, all_nodes, has_expanded]
   end
 
+  # !!! Horrible workaround until we fix crowbar orchestration !!!
+  # Because each chef-client runs executes everything in all cookbooks, and
+  # because the crm calls slow things down considerably, we have a drift issue
+  # where the founder (which does much more work since it's the only one doing
+  # the crm calls) goes much slower than the other nodes. This can impact
+  # things badly when it's so slow that it triggers the timeout in the sync
+  # marks.
+  # The workaround here is to reset the sync marks on the founder, so that all
+  # nodes wait on the sync marks and the drift is only happening between sync
+  # marks.
+  # The goal is to only do this when applying a proposal, so that other
+  # chef-client runs are not blocked waiting for sync marks.
+  def reset_sync_marks_on_clusters_founders(elements)
+    elements.each do |element|
+      next unless PacemakerServiceObject.is_cluster? element
+
+      cluster = cluster_name(element)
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -133,6 +133,37 @@ def role_expand_elements(role, role_name)
...
+  # where the founder (which does much more work since it's the only one doing
+  # the crm calls) goes much slower than the other nodes. This can impact
+  # things badly when it's so slow that it triggers the timeout in the sync
+  # marks.
+  # The workaround here is to reset the sync marks on the founder, so that all
+  # nodes wait on the sync marks and the drift is only happening between sync
+  # marks.
+  # The goal is to only do this when applying a proposal, so that other
+  # chef-client runs are not blocked waiting for sync marks.
+  def reset_sync_marks_on_clusters_founders(elements)
```
aplanas: Ah here it is

jsuchome:
+1 (not tested, but the code looks correct)

aplanas:
+1

aplanas:
+1

aplanas:
Gating fails with: `Error: Committing the crowbar 'default' proposal for 'dns' failed`, too early. But looks unrelated

vuntz:
Merging as gating now deploys (up to nova as this point in time), and this is just a new method that can't cause failures in the remaining part of the gating.

aspiers:
Retrospective -1 purely due to the newly introduced long lines :-p  (otherwise +1)

vuntz:
> due to the newly introduced long lines :-p

I love long lines ;-) We can fix that up now if you want.

aspiers:
Because they annoy the English guy, no doubt ;-p  Yes please.

vuntz:
> Because they annoy the English guy, no doubt ;-p Yes please.

Actually, that's fixed in https://github.com/crowbar/crowbar-ha/pull/61

PR #61, opened 2016-01-07
Reset the sync marks on founder when applying proposal (bsc#935462)

vuntz:
~~Depends and contains https://github.com/crowbar/crowbar-ha/pull/60 (we need to merge separately, as gating can't work for the full patch until https://github.com/crowbar/crowbar-openstack/pull/181 is merged too).~~

~~Depends on https://github.com/crowbar/crowbar-openstack/pull/181~~

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
@@ -59,9 +59,14 @@ end
 include_recipe "crowbar-pacemaker::stonith"
 
 # We let the founder go first, so it can generate the authkey and some other
-# initial pacemaker configuration bits; we do it in the first phase of the chef
-# run because the non-founder nodes will look in the compile phase for the
-# attribute
+# initial pacemaker configuration bits; we do it in the compile phase of the
+# chef run because the non-founder nodes will look during the compile phase for
+# the attribute, while the attribute is set during the convergence phase of the
+# founder node.
+# Note that resetting this sync mark should be avoided after the initial setup
+# is done: since wait is done during compile phase and create is done during
+# convergence phase, it can create a drift between the founder node and the
+# non-founder nodes.
 crowbar_pacemaker_sync_mark "wait-pacemaker_setup" do
   revision node[:pacemaker]["crowbar-revision"]
   # we use a longer timeout because the wait / create are in two different
@@ -72,8 +77,10 @@ end.run_action(:guess)
 
 include_recipe "pacemaker::default"
 
-# This is not done in the compile phase, because saving the authkey attribute
-# is done in a ruby_block
+# This part of the synchronization is *not* done in the compile phase, because
+# saving the corosync authkey attribute is done in convergence phase for
+# founder (but reading the attribute is done in compile phase for non-founder
+# nodes) -- see the corosync::authkey_generator recipe.
 crowbar_pacemaker_sync_mark "create-pacemaker_setup" do
   revision node[:pacemaker]["crowbar-revision"]
 end
--- a/crowbar_framework/app/models/pacemaker_service_object.rb
+++ b/crowbar_framework/app/models/pacemaker_service_object.rb
@@ -145,22 +145,30 @@ class PacemakerServiceObject < ServiceObject
   # marks.
   # The goal is to only do this when applying a proposal, so that other
   # chef-client runs are not blocked waiting for sync marks.
+  def self.reset_sync_marks_on_cluster_founder(founder, cluster)
+    return if founder.nil? ||
+        founder[:pacemaker].nil? ||
+        founder[:pacemaker][:sync_marks].nil? ||
+        founder[:pacemaker][:sync_marks][cluster].nil?
+
+    founder[:pacemaker][:sync_marks][cluster].keys.each do |sync_mark|
+      # this sync mark is special, only required for initial pacemaker
+      # setup and running in compile phase, so we don't want to reset it
+      next if sync_mark == "pacemaker_setup"
+      founder[:pacemaker][:sync_marks][cluster].delete(sync_mark)
+    end
+
+    founder.save
+  end
+
   def reset_sync_marks_on_clusters_founders(elements)
     elements.each do |element|
       next unless PacemakerServiceObject.is_cluster? element
 
       cluster = cluster_name(element)
       founder = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:pacemaker-config-#{cluster}").first
-      next if founder.nil? || founder[:pacemaker][:sync_marks].nil? || founder[:pacemaker][:sync_marks][cluster].nil?
 
-      founder[:pacemaker][:sync_marks][cluster].keys.each do |k|
-        # this sync mark is special, only required for initial pacemaker
-        # setup and running in compile phase, so we don't want to reset it
-        next if k == "pacemaker_setup"
-        founder[:pacemaker][:sync_marks][cluster].delete(k)
-      end
-
-      founder.save
+      PacemakerServiceObject.reset_sync_marks_on_cluster_founder(founder, cluster)
     end
   end
 
--- a/crowbar_framework/app/models/pacemaker_service_object.rb
+++ b/crowbar_framework/app/models/pacemaker_service_object.rb
@@ -162,8 +162,20 @@ class PacemakerServiceObject < ServiceObject
         founder[:pacemaker][:sync_marks][cluster].nil?
 
     founder[:pacemaker][:sync_marks][cluster].keys.each do |sync_mark|
-      # this sync mark is special, only required for initial pacemaker
-      # setup and running in compile phase, so we don't want to reset it
+      # The pacemaker_setup sync mark (see the crowbar-pacemaker::default
+      # recipe) requires special handling: it is created by the founder in
+      # Chef's convergence phase, but waited for by all non-founders in Chef's
+      # compile phase, because they need it in order to copy the authkey
+      # attribute from the founder to themselves and then invoke the
+      # corosync::authkey_writer recipe.
+      #
+      # This is only required for initial pacemaker setup and running in
+      # compile phase, so we don't want to reset it.  If we were to reset it,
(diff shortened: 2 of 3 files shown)


jsuchome:
Looks good. +1

aplanas:
As I understand #60 needs to have the gating green, but #61 can't until crowbar/crowbar-openstack#181 is merged, but not sure that `Error: Committing the crowbar 'data' proposal for 'pacemaker' failed` is related with this or not.

vuntz:
@aplanas I think it's a bug in my changes. Will investigate.

vuntz:
In gating, we deployed up to ceilometer and then the sync mark failed due to the missing https://github.com/crowbar/crowbar-openstack/pull/181

aplanas:
+1

aspiers:
Is the PR description now out of date?  Doesn't seem to contain #60 so I guess it was rebased?

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -145,22 +155,30 @@ def role_expand_elements(role, role_name)
...
   # chef-client runs are not blocked waiting for sync marks.
+  def self.reset_sync_marks_on_cluster_founder(founder, cluster)
+    return if founder.nil? ||
+        founder[:pacemaker].nil? ||
+        founder[:pacemaker][:sync_marks].nil? ||
+        founder[:pacemaker][:sync_marks][cluster].nil?
+
+    founder[:pacemaker][:sync_marks][cluster].keys.each do |k|
+      # this sync mark is special, only required for initial pacemaker
+      # setup and running in compile phase, so we don't want to reset it
```
aspiers: Blegh, hardcoding this is ugly even relative to the rest of this hack.  It's not immediately obvious to me from this comment what would be wrong with resetting it?  Although I appreciate this commit doesn't introduce this code, merely moves it.

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -145,22 +155,30 @@ def role_expand_elements(role, role_name)
   # marks.
   # The goal is to only do this when applying a proposal, so that other
   # chef-client runs are not blocked waiting for sync marks.
+  def self.reset_sync_marks_on_cluster_founder(founder, cluster)
+    return if founder.nil? ||
+        founder[:pacemaker].nil? ||
+        founder[:pacemaker][:sync_marks].nil? ||
+        founder[:pacemaker][:sync_marks][cluster].nil?
+
+    founder[:pacemaker][:sync_marks][cluster].keys.each do |k|
```
aspiers: Can we call it something like `sync_mark` rather than `k`, to make it clear what the keys represent?

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -89,6 +89,16 @@ def cluster_vhostname_from_element(element)
       end
     end
 
+    # Get the founder of a cluster, based on the element
+    def cluster_founder(element)
+      if is_cluster?(element)
+        cluster = cluster_name(element)
+        NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:pacemaker-config-#{cluster}").first
+      else
```
aspiers: Yeah, apparently `if false_expression ... end` returns `nil` in Ruby.  I prefer the explicit version, but Hound doesn't...

vuntz: I also prefer the explicit version, which is why I didn't fix that warning :-)

vuntz:
> Is the PR description now out of date? Doesn't seem to contain #60 so I guess it was rebased?

Correct, fixed.

crowbar_framework/app/models/pacemaker_service_object.rb
vuntz: I expanded on this in the comment. Does that help?

crowbar_framework/app/models/pacemaker_service_object.rb
vuntz: Fixed.

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -145,22 +155,33 @@ def role_expand_elements(role, role_name)
...
+        founder[:pacemaker].nil? ||
+        founder[:pacemaker][:sync_marks].nil? ||
+        founder[:pacemaker][:sync_marks][cluster].nil?
+
+    founder[:pacemaker][:sync_marks][cluster].keys.each do |sync_mark|
+      # This sync mark is special, only required for initial pacemaker
+      # setup and running in compile phase, so we don't want to reset it.
+      # If we were to reset it, then the founder would go to convergence while
+      # the non-founder nodes would finally be unblocked in compile phase,
+      # which would result in a huge drift.
```
aspiers: Thanks - this expanded comment is definitely an improvement ... but I still don't get it ;-/

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -145,22 +155,33 @@ def role_expand_elements(role, role_name)
...
   # The goal is to only do this when applying a proposal, so that other
   # chef-client runs are not blocked waiting for sync marks.
+  def self.reset_sync_marks_on_cluster_founder(founder, cluster)
+    return if founder.nil? ||
+        founder[:pacemaker].nil? ||
+        founder[:pacemaker][:sync_marks].nil? ||
+        founder[:pacemaker][:sync_marks][cluster].nil?
+
+    founder[:pacemaker][:sync_marks][cluster].keys.each do |sync_mark|
+      # This sync mark is special, only required for initial pacemaker
```
aspiers: s/This/The pacemaker_setup/

vuntz: done

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -145,22 +155,34 @@ def role_expand_elements(role, role_name)
...
+        founder[:pacemaker][:sync_marks].nil? ||
+        founder[:pacemaker][:sync_marks][cluster].nil?
+
+    founder[:pacemaker][:sync_marks][cluster].keys.each do |sync_mark|
+      # The pacemaker_setup sync mark is special, only required for initial
+      # pacemaker setup and running in compile phase, so we don't want to reset
+      # it.
+      # If we were to reset it, then the founder would go to convergence phase
+      # while the non-founder nodes would finally be unblocked in compile
+      # phase, which would result in a huge drift.
```
aspiers: I think this is clearer:

```
# The pacemaker_setup sync mark (see the crowbar-pacemaker::default
# recipe) requires special handling: it is created by the founder in
# Chef's convergence phase, but waited for by all non-founders in Chef's
# compile phase, because they need it in order to copy the authkey
# attribute from the founder to themselves and then invoke the
# corosync::authkey_writer recipe.
# 
# This is only required for initial pacemaker setup and running in
# compile phase, so we don't want to reset it.  If we were to reset it,
# then every time a proposal was applied, the non-founder nodes would be
# blocked in their compile phase until the founder reached the point in
# its convergence phase where it creates the pacemaker_setup sync mark,
# and this would be too long for the non-founders to wait when the run
# list is long.
```

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -72,8 +77,10 @@
 
 include_recipe "pacemaker::default"
 
-# This is not done in the compile phase, because saving the authkey attribute
-# is done in a ruby_block
+# This part of the synchronization is *not* done in the compile phase, because
+# saving the corosync authkey attribute is done in convergence phase for
+# founder (but reading the attribute is done in compile phase for non-founder
```
aspiers: Can you insert "(see the corosync::authkey_generator recipe)" here?

vuntz: done

aplanas:
Even more +1

bmwiedemann:
https://ci.suse.de/view/Cloud/view/Builds/job/openstack-mkcloud/14557/
gating failed with
```
crowbar ceilometer proposal commit default 
Failed to talk to service proposal commit: 404: {"status":"404","error":"Not Found"}
```

vuntz:
Issue is that starting ceilometer-polling took 40s, which basically killed the sync mark. We probably need to improve that sync mark.

vuntz:
I moved the revert bits to https://github.com/crowbar/crowbar-ha/pull/66 as they are causing the gating failure (we need to tweak things a bit in crowbar-openstack to avoid that), and are not 100% related to the main goal of this PR.

aspiers:
Why do we need to tweak stuff (presumably timeouts) in crowbar-openstack to avoid the gating failure?

vuntz:
> Why do we need to tweak stuff (presumably timeouts) in crowbar-openstack to avoid the gating failure?

See earlier comment: ceilometer-polling is very slow to start, so the 60s default timeout is too short for that one.

aspiers:
Ah OK, thanks.

aspiers:
Still slightly confused though, you said "starting ceilometer-polling took 40s" which is under 60s, so why did that "kill the sync mark"?  (What does "kill" mean in this context anyway?)

vuntz:
> Still slightly confused though, you said "starting ceilometer-polling took 40s" which is under 60s, so why did that "kill the sync mark"? (What does "kill" mean in this context anyway?)

"kill" means that it triggered a timeout on the other node ;-)

So starting ceilometer-polling took 40s, but then the rest of the work happening inside the sync mark also takes some time (since it's create in a transaction), and in the end it was a bit more than 1 minute. But the bit that really triggers the whole issue is the 40s.

aspiers:
Got it now, thanks!

vuntz:
Rebased, no code change.

vuntz:
@aspiers you didn't give a +1 after your review. Were you waiting for gating, or do you have additional comments?

aspiers:
Wow, look at that beautiful green HA gate!!!  No, this looks more than ready to me now.

PR #58, opened 2016-01-05
crowbar-pacemaker: Add ability to set custom attributes on nodes

vuntz:
This will be useful to set the OpenStack-role attribute.

Note that this contains https://github.com/crowbar/crowbar-ha/pull/55 -- only the last commit is different here.

PR #62, opened 2016-01-07
crowbar-pacemaker: Add ability to set custom attributes on nodes

vuntz:
This will be useful to set the OpenStack-role attribute.

This is the bit from https://github.com/crowbar/crowbar-ha/pull/58 that can be merged right now.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
@@ -94,6 +94,15 @@ if node[:pacemaker][:drbd][:enabled]
   include_recipe "crowbar-pacemaker::drbd"
 end
 
+node[:pacemaker][:attributes].each do |attr, value|
+  execute "set pacemaker attribute \"#{attr}\" to \"#{value}\"" do
+    command "crm node attribute #{node[:hostname]} set \"#{attr}\" \"#{value}\""
+    # The cluster only does a transition if the attribute value changes,
+    # so checking the value before setting would only slow things down
+    # for no benefit.
+  end
+end
+
 include_recipe "crowbar-pacemaker::haproxy"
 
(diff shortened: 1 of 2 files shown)


jsuchome:
Repeating my +1 here.

aplanas:
+1

vuntz:
This failed while deploying manila with `Errno::ENOMEM: package[bind9] (bind9::default line 19) had an error: Errno::ENOMEM: Cannot allocate memory - fork(2)`. Totally unrelated, and the deployment wasn't impacted by the change, so merging.

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -94,6 +94,15 @@
   include_recipe "crowbar-pacemaker::drbd"
 end
 
+node[:pacemaker][:attributes].each do |attr, value|
+  execute "set pacemaker attribute \"#{attr}\" to \"#{value}\"" do
+    command "crm node attribute #{node[:hostname]} set \"#{attr}\" \"#{value}\""
```
aspiers: I know you love backslash-quoting double quotes inside double quotes, but I still have no clue why :-)  Something like

    command %'crm node attribute #{node[:hostname]} set "#{attr}" "#{value}"'

avoids the requirement for the reader's brain to decode the quoting themselves, because they can see exactly what the intended string is without imagining any changes to it.

aspiers:
Other than your weird backslash fetish this looks fine :)

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
vuntz: Done in https://github.com/crowbar/crowbar-ha/pull/64

PR #63, opened 2016-01-08
crowbar-pacemaker: Add num_remote_nodes helper

vuntz:
This will be useful in the same way as num_corosync_nodes.


jsuchome:
+1

aplanas:
+1

aspiers:
Tempest failure in gate is unrelated so I've filed [a card for that](https://trello.com/c/t7G9D2S7) and am now merging this non-intrusive change which will make testing of compute node HA easier.

PR #67, opened 2016-01-14
pacemaker: Use crm_resource instead of crm when checking for existence

vuntz:
crm is relatively slow (especially in the CI), while crm_resource is
fast (10x faster). This can have quite an impact on the duration of
chef-client for a cluster with many resources.

### DIFF
--- a/chef/cookbooks/pacemaker/libraries/pacemaker/resource.rb
+++ b/chef/cookbooks/pacemaker/libraries/pacemaker/resource.rb
@@ -11,6 +11,21 @@ module Pacemaker
       "#{type} resource"
     end
 
+    def exists?(name)
+      # Directly use crm_resource, as it's faster than crm. This cannot be done
+      # for all CIBObjects as this works only for resources, and not for
+      # constraints :/
+      cmd = Mixlib::ShellOut.new("crm_resource --resource #{name} --query-xml-raw")
+      cmd.environment["HOME"] = ENV.fetch("HOME", "/root")
+      cmd.run_command
+      begin
+        cmd.error!
+        true
+      rescue
+        false
+      end
+    end
+
     def running?
       cmd = shell_out! "crm", "resource", "status", name
       Chef::Log.info cmd.stdout


jsuchome:
Works manually, and I do not see any problems with the code. +1

chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
```diff
@@ -27,7 +27,16 @@ def get_definition(name)
       end
 
       def exists?(name)
-        get_definition(name) ? true : false
+        # Directly use crm_resource, as it's faster than crm
+        cmd = Mixlib::ShellOut.new("crm_resource --resource #{name} --query-xml-raw")
```
aplanas: can this be a new function like `exists_definition?()`?

aplanas:
+1

chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
jsuchome: Well, if the point is replace current calls, than it would need replacing all callers. If anything, we can have extra function for the current behavior...

bmwiedemann:
https://ci.suse.de/view/Cloud/view/Builds/job/openstack-mkcloud/14655/ failed in rabbitmq proposal
```
STDERR: ERROR: 2: rsc_colocation (1): Invalid attribute id for element rsc_colocation
       rsc_location (0): Element constraints has extra content: rsc_location
   error: main:	CIB did not pass DTD/schema validation
Errors found during check: config not valid
```

aspiers:
-1 - you didn't update the test suite, so Travis breaks.  Please run `cd chef/cookbooks/pacemaker; bundle exec guard` while you are hacking on this code :)

aspiers:
(but +1 for the approach)

vuntz:
> -1 - you didn't update the test suite, so Travis breaks.

I know; it was just a quick test in the middle of the night; will try to come back to this when I have time.

vuntz:
Hrm, the failure actually shows that `crm_resource` can't be used like this to query location constraints :/

aspiers:
In that case shouldn't we give up on this approach and close this?  Or would fixing the Travis failures give this a better chance of succeeding? :)

aspiers:
Oh you moved the method from the base class to the resource class.  OK that makes sense.  But it's still not the best solution IMHO.

vuntz:
@aspiers it's still WIP, but I'm not convinced that works well in the end, indeed. Didn't see that much improvement in the timing :/

aspiers:
To clarify: a better solution would be to implement `Pacemaker::CIBObject.all` which retrieves all objects from a single `crm configure show`, and then use this in the `pacemaker_transaction` existence test loop, rather than invoking a command for each existence test.

aspiers:
@vuntz roughly followed my suggestion in #79 which is now merged - should we close this?

aspiers:
Closing - @vuntz please reopen if you disagree.

PR #70, opened 2016-01-20
Added required new methods for ha setups with remote nodes

tboerger:
**Depends on https://github.com/crowbar/crowbar-core/pull/193**

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service_object.rb
+++ b/crowbar_framework/app/models/pacemaker_service_object.rb
@@ -28,13 +28,6 @@ class PacemakerServiceObject < ServiceObject
   #
 
   class << self
-    # This is the key that allows to find out that an element item is a
-    # pacemaker cluster: if the element name is $cluster_key:$foo, then it's
-    # one. Otherwise, it's not.
-    def cluster_key
-      "cluster"
-    end
-
     # Note that we cannot cache the list of clusters here as we're in a
     # eigenclass, and so the cache will be longer term than a single request
     # (hence we won't notice new clusters). It's therefore up to the callers to
@@ -50,26 +43,32 @@ class PacemakerServiceObject < ServiceObject
       clusters
     end
 
-    # Returns: name of the barclamp and of the proposal for this cluster
-    def cluster_get_barclamp_and_proposal(element)
-      if is_cluster?(element)
-        ["pacemaker", cluster_name(element)]
-      else
-        [nil, nil]
-      end
+    # Returns: List of available clusters excluding remotes
+    def available_clusters_excluding_remotes
+      # TODO(must): Fetch only clusters excluding remotes
+      {
+        "cluster:services" => available_clusters["cluster:services"]
+      }
+    end
+
+    # Returns: List of available clusters including remotes
+    def available_clusters_including_remotes
+      # TODO(must): Fetch only clusters including remotes
+      {
+        "remotes:test_1" => available_clusters["cluster:test_1"],
+        "remotes:test_2" => available_clusters["cluster:test_2"]
+      }
+    end
+
+    # This is the key that allows to find out that an element item is a
+    # pacemaker cluster excluding remote nodes: if the element name is
+    # $cluster_key:$foo, then it's one. Otherwise, it's not.
+    def cluster_key
+      "cluster"
     end
 
     def is_cluster?(element)
       element.start_with? "#{cluster_key}:"
-    end
-
-    # Returns: name of the cluster, or nil if it's not a cluster
-    def cluster_name(element)
-      if is_cluster? element
-        element.gsub("#{cluster_key}:", "")
-      else
-        nil
-      end
     end
 
     def cluster_vhostname_from_name(name)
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -172,7 +218,7 @@ def allocate_virtual_ips_for_any_cluster_in_networks(elements, networks)
     new_allocation = false
 
     elements.each do |element|
-      if PacemakerServiceObject.is_cluster? element
+      if PacemakerServiceObject.is_cluster?(element) || PacemakerServiceObject.is_remotes?(element)
```
vuntz: Actually I don't think we want this to work on remotes.

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -208,7 +254,7 @@ def prepare_role_for_cluster_vip_networks(role, elements, networks)
     dirty = false
 
     elements.each do |element|
-      next unless PacemakerServiceObject.is_cluster? element
+      next unless PacemakerServiceObject.is_cluster?(element) && PacemakerServiceObject.is_remotes?(element)
```
vuntz: Same here, my gut feeling is that we don't want this on remotes.

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -142,7 +188,7 @@ def role_expand_elements(role, role_name)
   #     taken). This information can be used to know when a DNS sync is
   #     required or not.
   def allocate_virtual_ips_for_cluster_in_networks(cluster, networks)
-    if networks.nil? || networks.empty? || !PacemakerServiceObject.is_cluster?(cluster)
+    if networks.nil? || networks.empty? || !PacemakerServiceObject.is_cluster?(cluster) || !PacemakerServiceObject.is_remotes?(cluster)
```
vuntz: Same here (in short: virtual IPs should not be for remotes).

vuntz:
Marking as WIP because of the TODO (hard-coded clusters).

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -50,39 +43,92 @@ def available_clusters
...
-      "cluster-#{name.gsub("_", "-")}.#{ChefObject.cloud_domain}"
-    end
-
     def cluster_vhostname_from_element(element)
-      if is_cluster? element
+      case
+      when is_remotes?(element)
+        remotes_vhostname_from_name(cluster_name(element))
+      when is_cluster?(element)
         cluster_vhostname_from_name(cluster_name(element))
```
vuntz: Similarly to the VIP comments, I'm not sure we care about a vhostname for remotes.

vuntz:
Did some changes in https://github.com/tboerger/crowbar-ha/pull/1.

tboerger:
@vuntz I merged your pr, so the source is updated now.

vuntz:
I forgot, but +1.

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -81,8 +77,53 @@ def cluster_vhostname_from_name(name)
...
+      end
+    end
+
+    # Returns: name of the barclamp and of the proposal for this cluster
+    def cluster_get_barclamp_and_proposal(element)
+      case
+      when is_remotes?(element)
+        ["pacemaker", cluster_name(element)]
+      when is_cluster?(element)
+        ["pacemaker", cluster_name(element)]
```
jsuchome: Is the result really the same for `is_cluster?` and `is_remotes?`

jsuchome: Ah, `cluster_name` is different in those cases. But why do we need 2 `when` branches here?

crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -50,26 +43,29 @@ def available_clusters
...
     end
 
-    # Returns: name of the barclamp and of the proposal for this cluster
-    def cluster_get_barclamp_and_proposal(element)
-      if is_cluster?(element)
-        ["pacemaker", cluster_name(element)]
-      else
-        [nil, nil]
+    # Returns: List of available clusters including remotes
+    def available_clusters_including_remotes
```
jsuchome: The name is somewhat confusing. I thought it's gonna return all available clusters, so those which has remotes as well as those who have them.

Isn't it rather `available_remotes`, `all_remotes_from_clusters`  or something like that?

vuntz: Agreed, the naming is a bit confusing. I would be fine with `available_remotes` (needs a change in the crowbar-core PR too).

crowbar_framework/app/models/pacemaker_service_object.rb
vuntz: I guess for consistency with some other method. I'm fine if this gets changed.

jsuchome:
+1

tboerger:
So merge this?

vuntz:
> So merge this?

Waiting for the gating job, I guess.

vuntz:
Gating fails due to https://trello.com/c/30w2SCA0/128-c6-ha-waiting-for-testvm-booted-and-ping-returned-timed-out-in-testsetup-step (tempest passed, though)

Given that the code change is on the crowbar side of things, and actually unused yet (is used by other PR), I think we're good enough for merging.

PR #56, opened 2016-01-05
allow CIB objects with arbitrary unparsed definition strings (updated)

vuntz:
This is https://github.com/crowbar/crowbar-ha/pull/54 with some fixes.

PR #51, opened 2015-12-17
apache, haproxy, notifications, stonith: Use new pacemaker_transaction LWRP to create resources

vuntz:

PR #71, opened 2016-01-21
crowbar-pacemaker: Add location constraint for haproxy & apache

vuntz:
~~**Note:** only last commit is of interest here for the review, the rest is from https://github.com/crowbar/crowbar-ha/pull/51 (but is required for this last commit).~~

**Note:** this goes hand in hand with https://github.com/crowbar/crowbar-openstack/pull/215

**Note:** this depends on ~~https://github.com/crowbar/crowbar-ha/pull/56~~

Because the pacemaker resources for haproxy and apache are created here
in crowbar-pacemaker, we, in theory, shouldn't assume there that these
two services will be used for OpenStack and so we can't always just add
the location constraint to tell pacemaker to only run these on
controller nodes.

However, we do need the constraints when it's OpenStack. So crowbar will
add a flag to tell us to add these constraints, and when it's there, we
do what we're asked to do.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
@@ -66,6 +66,11 @@ pacemaker_clone clone_name do
 end
 transaction_objects << "pacemaker_clone[#{clone_name}]"
 
+if node[:pacemaker][:apache2][:for_openstack]
+  location_name = openstack_pacemaker_controller_only_location_for clone_name
+  transaction_objects << "pacemaker_location[#{location_name}]"
+end
+
 pacemaker_transaction "apache service" do
   cib_objects transaction_objects
   # note that this will also automatically start the resources
--- a/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
@@ -91,6 +91,11 @@ if node[:pacemaker][:haproxy][:clusters].key?(cluster_name) && node[:pacemaker][
   end
   transaction_objects << "pacemaker_group[#{group_name}]"
 
+  if node[:pacemaker][:haproxy][:for_openstack]
+    location_name = openstack_pacemaker_controller_only_location_for group_name
+    transaction_objects << "pacemaker_location[#{location_name}]"
+  end
+
   pacemaker_transaction "haproxy service" do
     cib_objects transaction_objects
     # note that this will also automatically start the resources
--- a/chef/cookbooks/crowbar-pacemaker/metadata.rb
+++ b/chef/cookbooks/crowbar-pacemaker/metadata.rb
@@ -10,3 +10,5 @@ depends "drbd"
 depends "haproxy"
 depends "lvm"
 depends "pacemaker"
+
+recommends "crowbar-openstack"


vuntz:
@aspiers wondering if you feel okay with an approach like this. I can't find a way to implement this bit without a hack :/

aspiers:
Filed gating failure in Trello: [c6-ha: Waiting for 'testvm booted and ping returned' timed out in testsetup step](https://trello.com/c/30w2SCA0/128-c6-ha-waiting-for-testvm-booted-and-ping-returned-timed-out)

jsuchome:
+1 (for last commit)

vuntz:
Rebased to only have that last commit.

vuntz:
Gating fails because we depend on https://github.com/crowbar/crowbar-openstack/pull/179.

chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
```diff
@@ -66,6 +66,11 @@
 end
 transaction_objects << "pacemaker_clone[#{clone_name}]"
 
+if node[:pacemaker][:apache2][:for_openstack]
+  location_name = openstack_pacemaker_controller_only_location_for clone_name
```
aspiers: Why not move this line to `crowbar-openstack`? Then `node[:pacemaker][:apache2][:objects][:for_openstack]` could be an `Array` of objects to append to `transaction_objects`.

chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
```diff
@@ -66,6 +66,11 @@
...
+  transaction_objects << "pacemaker_location[#{location_name}]"
```
aspiers: With my suggestion above, this would become:

    node[:pacemaker][:apache2][:objects].values.each do |objects|
      transaction_objects += objects
    end

chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
```diff
@@ -91,6 +91,11 @@
   end
   transaction_objects << "pacemaker_group[#{group_name}]"
 
+  if node[:pacemaker][:haproxy][:for_openstack]
+    location_name = openstack_pacemaker_controller_only_location_for group_name
```
aspiers: Same here.

chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
aspiers: In case it's not obvious, the idea behind this is that this file in the `crowbar-ha` repo wouldn't have to even mention OpenStack (since the keys of this `Hash` are ignored).

vuntz: @aspiers and where would `node[:pacemaker][:apache2][:objects]` be populated? Keep in mind that this recipe is one of the first in the run list, and it's before the openstack recipes. So we can't do that from any openstack recipe.

(+ it has to be done in chef, not crowbar, since you're talking about referencing a chef resource)

That was my initial intent, but I wasn't able to resolve this.

aspiers: Couldn't you have an openstack recipe earlier on the run list than this?

vuntz: Unfortunately, no: the runlist is built from roles, so that means we would need to put an openstack role before the pacemaker one, which wouldn't work (because we need apache / haproxy to be setup before we do the setup for the openstack services).

aspiers: IIUC, it would work if there were two roles: a new role in `crowbar-openstack` coming before the `crowbar-ha` code which only populated `node[:pacemaker][:apache2][:objects]`, and then the existing role which does the usual stuff.  But maybe that's overkill.

aspiers:
Have forced a rebuild since merging https://github.com/crowbar/crowbar-openstack/pull/179 although the rebuild will probably happen before a new package can be built from https://github.com/crowbar/crowbar-openstack/pull/179.

chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
vuntz: The main question is how you would set this new role as part of crowbar with the current architecture. It is possible, but then you move the hack elsewhere, imho.

vuntz:
Gating is green now.

jsuchome:
Still +1 from me

abel-navarro:
LGTM, +1

chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
aspiers: Alright then; I've added this to the growing [list of things which suck due to not having decent orchestration](https://etherpad.nue.suse.com/p/cloud-orchestration).

PR #72, opened 2016-01-22
pacemaker: Be stricter when parsing "crm configure show" output

vuntz:
In some very rare cases, we can get "crm configure show $foo" return
more than one CIB objects. This seems to happen for remote nodes that
have attributes.

$ crm configure show remote-d52-54-77-77-77-04
node remote-d52-54-77-77-77-04:remote \
        attributes OpenStack-role=compute
primitive remote-d52-54-77-77-77-04 ocf:pacemaker:remote \
        params reconnect_interval=60s server=d52-54-77-77-77-04.h9.cloud.suse.de \
        op monitor interval=20s \
        op start timeout=60s interval=0 \
        op stop timeout=60s interval=0

To avoid this situation, we are stricter and only keep for parsing the
object definition that has the exact name
("remote-d52-54-77-77-77-04:remote" is different from
"remote-d52-54-77-77-77-04").

Unfortunately, a side effect is that we cannot use anymore the CIBObject
class to parse remote node.

### DIFF
--- a/chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
+++ b/chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
@@ -20,7 +20,27 @@ module Pacemaker
         cmd.run_command
         begin
           cmd.error!
-          cmd.stdout
+
+          lines_for_def = []
+          cmd.stdout.lines.each do |line|
+            if lines_for_def.empty?
+              # look for the beginning of the resource definition
+              next unless line =~ /\A(\w+)\s#{name}\s/
+            else
+              # stop parsing if we reach the end of our definition
+              break if line !~ /\A\s/
+            end
+            lines_for_def.push(line)
+          end
+
+          if lines_for_def.empty?
+            # in case our attempt to only extract the definition we're
+            # interested in failed, play it safe and return the output
+            ::Chef::Log.warn "Failed to extract definition for #{name} from:\n#{cmd.stdout}"
+            cmd.stdout
+          else
+            lines_for_def.join("")
+          end
         rescue
           nil
         end


vuntz:
An alternative hack would be https://github.com/crowbar/crowbar-ha/compare/master...vuntz:ignore-node-cibs?expand=1 -- but this one feels slightly less bad.

chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
```diff
@@ -20,7 +20,26 @@ def get_definition(name)
         cmd.run_command
         begin
           cmd.error!
-          cmd.stdout
+
+          lines_for_def = []
+          cmd.stdout.lines.each do |line|
+            if lines_for_def.empty?
+              # look for the beginning of the resource definition
+              next unless line =~ /\A(\w+)\s#{name}\s/
```
aspiers: Curious why you chose `\A` here rather than `^`?  Normally that implies matching against a multi-line string, but that's not the case here.

chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
```diff
@@ -20,7 +20,20 @@ def get_definition(name)
...
+            else
+              # stop parsing if we reach the end of our definition
+              break if line !~ /\A\s/
```
aspiers: Ditto here.

chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
```diff
@@ -20,7 +20,26 @@ def get_definition(name)
...
+            end
+            lines_for_def.push(line)
+          end
+
+          if lines_for_def.empty?
+            # in case our attempt to only extract the definition we're
+            # interested in failed, play it safe and return the output
+            cmd.stdout
```
aspiers: Log a warning here?

aspiers:
> Unfortunately, a side effect is that we cannot use anymore the CIBObject class to parse remote node.

Why not?

> An alternative hack would be https://github.com/crowbar/crowbar-ha/compare/master...vuntz:ignore-node-cibs?expand=1 -- but this one feels slightly less bad.

Ugh yeah, this one is much better.

+1 for the approach; just might be a bit more polished taking into account my comments.

chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
vuntz: I don't mind strongly either way. But I thought it was better to use `\A` unless we explicitly only care about one line anywhere in a multi-line string, so even if we know here's it's one line, it's a better habit than `^`.

Want me to change to `^`?

chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
vuntz: fixed

vuntz:
>> Unfortunately, a side effect is that we cannot use anymore the CIBObject class to parse remote node.
> 
> Why not?

```
$ crm configure show d52-54-77-77-77-02
node 1084807505: d52-54-77-77-77-02
```

(and for remote nodes, see the commit message)

Although, it's true that the fallback code should still make it work, I guess (but not for remote nodes, since that's the whole point here).

jsuchome:
+1

chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
aspiers: I dunno... the counter argument is that `\A` is far less common / well-known and therefore not as readable (I knew what it meant but even so I had to look it up in a moment of forgetfulness).  But I see your point.  Let's get another opinion ;-)

vuntz: Just for the record: I really have no strong opinion to this, and I'll change if anyone prefers to use `^`.

PR #73, opened 2016-01-27
Fix intendation

nkrinner:
Maybe not  worth merging yet, other changes might come there too.

So far this is only a simple fix for an intendation error in the pacemaker barclamp.


vuntz:
+1

nkrinner:
Actually, merge at will

jsuchome:
+1

PR #74, opened 2016-01-27
crowbar: Add cluster_from_remotes helper

vuntz:
This is used when orchestrating the application of a role that include
remotes: we then need to find the matching cluster to run some followup
action.

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service_object.rb
+++ b/crowbar_framework/app/models/pacemaker_service_object.rb
@@ -57,6 +57,14 @@ class PacemakerServiceObject < ServiceObject
       remotes
     end
 
+    # Returns: the element name for the cluster that is the match of the
+    # remotes element
+    def cluster_from_remotes(element)
+      return nil unless is_remotes?(element)
+
+      "#{cluster_key}:#{cluster_name(element)}"
+    end
+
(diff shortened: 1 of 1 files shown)


jsuchome:
+1 in general, but it leads me to question about optimalization: in this small method, we're invoking `RoleObject.find_role_by_name` two times. Wouldn't it be possible to reduce such calls? (Probably by making the code less DRY, and maybe it doesn't really matter)

aplanas:
+1

vuntz:
> but it leads me to question about optimalization: in this small method, we're invoking RoleObject.find_role_by_name two times.

Actually, we're not: the methods we're calling are "just" looking at the names of the element.

jsuchome:
Ah, true, I was looking at the wrong place, sorry.. :-(

vuntz:
There's one tempest failure (tempest.scenario.test_volume_boot_pattern.TestVolumeBootPattern.test_volume_boot_pattern); that's https://trello.com/c/98yQpVjE/120-c6-tempest-test-volume-boot-pattern-block-device-mapping-is-invalid

Unrelated to this PR as only a helper method on the crowbar side is added (and it's not used yet) => merging.

PR #69, opened 2016-01-15
STONITH for remote nodes

vuntz:
Please note that this depends (and contains) https://github.com/crowbar/crowbar-ha/pull/55

PR #75, opened 2016-01-28
Code cleanups

vuntz:
Simplify some helper methods by avoiding unneeded blocks.

This should address https://github.com/crowbar/crowbar-ha/pull/69#discussion_r50835031

cc @jsuchome

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -26,29 +26,23 @@ module CrowbarPacemakerHelper
   end
 
   def self.is_cluster_founder?(node)
-    if cluster_enabled?(node)
-      node[:pacemaker][:founder]
-    else
-      false
-    end
+    return false unless cluster_enabled?(node)
+
+    node[:pacemaker][:founder]
   end
 
   # Returns the number of corosync (or non-remote) nodes in the cluster.
   def self.num_corosync_nodes(node)
-    if cluster_enabled?(node)
-      node[:pacemaker][:elements]["pacemaker-cluster-member"].length
-    else
-      0
-    end
+    return 0 unless cluster_enabled?(node)
+
+    node[:pacemaker][:elements]["pacemaker-cluster-member"].length
   end
 
   # Returns the number of remote nodes in the cluster.
   def self.num_remote_nodes(node)
-    if cluster_enabled?(node)
-      node[:pacemaker][:elements]["pacemaker-remote"].length
-    else
-      0
-    end
+    return 0 unless cluster_enabled?(node)
+
+    node[:pacemaker][:elements]["pacemaker-remote"].length
   end
 
   # Returns the name of the cluster containing the given node, or nil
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/pacemaker_service_object.rb
```diff
@@ -97,22 +97,18 @@ def is_remotes?(element)
...
     end
 
     # Returns: name of the barclamp and of the proposal for this cluster
     def cluster_get_barclamp_and_proposal(element)
-      if is_cluster?(element) || is_remotes?(element)
-        ["pacemaker", cluster_name(element)]
-      else
-        [nil, nil]
-      end
+      return [nil, nil] unless is_cluster?(element) || is_remotes?(element)
```
jsuchome: `&&` ??

vuntz: Good catch!

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -108,15 +96,13 @@ def self.cluster_haproxy_vadmin_name(node)
...
-        vpublic_name
-      end
-    else
+    return nil unless cluster_enabled?(node)
+
+    vpublic_name = node[:pacemaker][:haproxy][:public_name]
+    if vpublic_name.nil? || vpublic_name.empty?
       nil
+    else
+      vpublic_name
```
sayalilunkad: I guess you can also change this part to something like this:
return nil if vpublic_name.nil? || vpublic_name.empty?
vpublic_name

crowbar_framework/app/models/pacemaker_service_object.rb
vuntz: Actually no, current code is correct. I think :-)

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
vuntz: Yes, that'd be an option. I only wanted to move away the "is not a cluster" cases, though. Want me to do the change?

crowbar_framework/app/models/pacemaker_service_object.rb
jsuchome: Ouch, yes, you are right and I can't read logical expressions... :-(

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
sayalilunkad: I have no strong opinion. Just thought it would look cleaner this way :)

jsuchome:
+1 (no opinion about that remainng issue)

aplanas:
Gating fail not related (Ports in binding_failed: 1)

vuntz:
We're hitting https://trello.com/c/30w2SCA0/128-c6-ha-waiting-for-testvm-booted-and-ping-returned-timed-out-in-testsetup-step :/ Tempest passed though, so I'll consider this good enough for merging.

PR #77, opened 2016-02-01
crowbar-pacemaker: Set clone-max metadata for clone resources

vuntz:
This is useful to have hawk / crm not be confused by remote nodes which
should not use the clones.


vuntz:
This is hitting https://trello.com/c/NpImhTvk/150-c6-binding-failed-ret-102 but deployment and tempest were fine; so good enough for merging.

Any +1?

aplanas:
+1

jsuchome:
+1

PR #79, opened 2016-02-01
pacemaker: Only call "crm configure show" once per transaction

vuntz:
Previous code was using ::Pacemaker::CIBObject.exists? for each object
in the transaction, which calls "crm configure show" that many times.

crm being a python tool, it can take some time to start, and doing that
so many times is not very optimal. So just run it once and filter the
output to get a list of existing objects.

### DIFF
--- a/chef/cookbooks/pacemaker/providers/transaction.rb
+++ b/chef/cookbooks/pacemaker/providers/transaction.rb
@@ -27,9 +27,27 @@ action :commit_new do
     run_context.resource_collection.lookup(chef_resource_name)
   end
 
+  cmd = Mixlib::ShellOut.new("crm --display=plain configure show")
+  cmd.environment["HOME"] = ENV.fetch("HOME", "/root")
+  cmd.run_command
+
+  existing_objects = begin
+    cmd.error!
+    # Only keep object names from the output; so remove continuation lines and
+    # do some regexp magic.
+    # Also deal with definitions like fencing_topology which have no name (the
+    # regexp won't match, so we'll be good).
+    cmd.stdout.lines.select { |l| l[0] !~ /\s/ }.map do |l|
+      l.chomp.chomp("\\").gsub(/\A\S+ (\S+).*/, "\\1")
+    end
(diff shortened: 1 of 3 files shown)


vuntz:
In a mkcloud setup with all services using HA, this made chef-client go from 6 minutes 30s to 5 minutes 30s.

toabctl:
+1

bmwiedemann:
https://ci.suse.de/view/Cloud/view/Builds/job/openstack-mkcloud/16065
an older version failed in mkcloud with
```
Error: floating ip addr from manila-service instance not available
```

vuntz:
This is hitting https://trello.com/c/30w2SCA0/128-c6-ha-waiting-for-testvm-booted-and-ping-returned-timed-out-in-testsetup-step but deployment and tempest were fine; so good enough for merging.

Any other +1?

aplanas:
+1

chef/cookbooks/pacemaker/providers/transaction.rb
```diff
@@ -27,9 +27,27 @@
...
+  cmd.run_command
+
+  existing_objects = begin
+    cmd.error!
+    # Only keep object names from the output; so remove continuation lines and
+    # do some regexp magic.
+    # Also deal with definitions like fencing_topology which have no name (the
+    # regexp won't match, so we'll be good).
+    cmd.stdout.lines.select { |l| l[0] !~ /\s/ }.map do |l|
+      l.chomp.chomp("\\").gsub(/\A\S+ (\S+).*/, "\\1")
```
aspiers: It might be more reliable and/or cleaner to parse the output of `cibadmin -Q --xpath /cib/configuration/*/* --node-path`

aspiers: Also this code is better placed not in the LWRP but as a class method in `Pacemaker::CIBObject` [as I previously suggested](https://github.com/crowbar/crowbar-ha/pull/67#issuecomment-173874714).  We can fix after the SOC6 release.

PR #80, opened 2016-02-02
Make new remote nodes get the already existing roles applied for remote cluster

vuntz:
If remote nodes of a cluster are assigned to a compute role (for instance), then when we add a new remote node, this remote node should also get the compute role automatically.

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -176,16 +176,16 @@ class PacemakerService < ServiceObject
 
       required_barclamp_roles.each do |required_barclamp_role|
         name = required_barclamp_role[:name]
-        unless node.role? name
-          priority = required_barclamp_role[:priority]
-          element_states = required_barclamp_role[:element_states]
+        next if node.role? name
 
-          @logger.debug("[pacemaker] AR: Adding role #{name} to #{node.name} with priority #{priority}")
-          node.add_to_run_list(name, priority, element_states)
-          save_it = true
+        priority = required_barclamp_role[:priority]
+        element_states = required_barclamp_role[:element_states]
 
-          required_pre_chef_calls << { service: required_barclamp_role[:service], barclamp_role: required_barclamp_role[:barclamp_role] }
-        end
+        @logger.debug("[pacemaker] AR: Adding role #{name} to #{node.name} with priority #{priority}")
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -178,26 +171,55 @@ def apply_cluster_roles_to_new_nodes(role, member_nodes)
...
+    # - we have roles inside each barclamp roles (as in "the role I assign to
+    #   nodes, like provisioner-server")
+
+    # Make sure that all nodes in the cluster have all the roles assigned to
+    # this cluster.
+
+    required_pre_chef_calls = []
+    all_roles = RoleObject.all
+
+    required_pre_chef_calls.concat(
```
jsuchome: Why is `concat` here? Can't you assign result of `apply_cluster_roles_to_new_nodes_for` directly to `required_pre_chef_calls`?

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -114,22 +114,14 @@ def expand_nodes_in_barclamp_role(cluster_role, node_object_all)
     all_nodes_for_cluster_role_expanded
   end
 
-  def apply_cluster_roles_to_new_nodes(role, member_nodes)
+  def apply_cluster_roles_to_new_nodes_for(cluster_element, relevant_nodes, all_roles)
     ### Beware of possible confusion between different level of "roles"!
```
jsuchome: Shouldn't we exit early if `relevant_nodes` is empty?

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Just because I wanted to have the same code for clusters & remotes (see a few lines below). I favored readability over simplicity here.

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Good point, you're right.

vuntz: Done

jsuchome:
+1

vuntz:
This is hitting https://trello.com/c/NpImhTvk/150-c6-binding-failed-ret-102 but deployment and tempest were fine; so good enough for merging.

Any other +1?

aplanas:
+1

PR #82, opened 2016-02-04
dont remove node from unallocated list (bsc#957315)

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=957315

when removing a node from pacemaker-cluster-member assignment,
we must not remove it from the list of available (unallocated) nodes

### DIFF
--- a/crowbar_framework/app/assets/javascripts/barclamps/pacemaker/application.js
+++ b/crowbar_framework/app/assets/javascripts/barclamps/pacemaker/application.js
@@ -95,7 +95,6 @@
     $(document).on('nodeListNodeUnallocated', this.root, function(evt, data) {
       if (self._ignore_event(evt, data)) { return; }
 
-      $(this).find('[data-id="{0}"]'.format(data.id)).remove();
       self.removeJson(data.id, null, "map");
     });
   };


bmwiedemann:
note that tests with this change looked fine, but the code is not fully understood (written by Balazs in 2014)

vuntz:
I'm nearly sure this is wrong: is the list of nodes in the stonith part of the barclamp updated?

bmwiedemann:
ah, it is added, but not removed now (unless you click save).
So we would need to find a way to remove the one entry, but not the other.

vuntz:
@bmwiedemann have you seen my comment in the bug? I think I found back in December what the issue was, but had no idea how to fix it.

crowbar_framework/app/assets/javascripts/barclamps/pacemaker/application.js
```diff
@@ -95,7 +95,6 @@
     $(document).on('nodeListNodeUnallocated', this.root, function(evt, data) {
       if (self._ignore_event(evt, data)) { return; }
 
-      $(this).find('[data-id="{0}"]'.format(data.id)).remove();
```
tboerger: This have to stay, but with a small fix:

```js
self.root.find('[data-id="{0}"]'.format(data.id)).remove();
```

tboerger:
@bmwiedemann  please fix the SUSE-Cloud master branch!

PR #83, opened 2016-02-05
Upgrade CIB syntax version

jsuchome:
This is based on http://docserv.suse.de/documents/SLE-HA12-SP1/SLE-HA-guide/html/cha.ha.migration.html#sec.ha.migration.upgrade.offline

### DIFF
--- a/chef/cookbooks/pacemaker/recipes/setup.rb
+++ b/chef/cookbooks/pacemaker/recipes/setup.rb
@@ -38,3 +38,17 @@ execute "crm initial configuration" do
   subscribes :run, resources(template: crm_conf), :immediately
   action :nothing
 end
+
+# The output of "cibadmin -Q -n" looks like:
+# <cib epoch="41" num_updates="9" admin_epoch="0" validate-with="pacemaker-1.2" ...>
+cib_cmd = "cibadmin -Q -n"
+
+execute "upgrade CIB syntax" do
+  user "root"
+  command "crm configure upgrade force"
+  not_if do
+    cib_meta = Mixlib::ShellOut.new(cib_cmd).run_command.stdout.chomp
+    cib_meta.gsub(/.*validate-with="[^"]*-([^-"]*)".*/, "\\1") >=
+      node[:pacemaker][:cib_syntax_version]
+  end
+end
--- a/chef/cookbooks/pacemaker/attributes/default.rb
+++ b/chef/cookbooks/pacemaker/attributes/default.rb
@@ -54,6 +54,9 @@ default[:pacemaker][:crm][:initial_config_file] = "/etc/corosync/crm-initial.con
 default[:pacemaker][:crm][:no_quorum_policy] = "ignore"
 default[:pacemaker][:crm][:op_default_timeout] = 60
 
+# acceptable CIB syntax version; if lower is detected, we must force its upgrade
+default[:pacemaker][:cib_syntax_version] = "2.4"
+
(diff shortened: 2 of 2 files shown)


dirkmueller:
+1

chef/cookbooks/pacemaker/recipes/setup.rb
```diff
@@ -38,3 +38,13 @@
...
 end
+
+# The output of "cibadmin -Q | grep validate" looks like:
+# <cib epoch="41" num_updates="9" admin_epoch="0" validate-with="pacemaker-1.2" ...>
+cib_cmd = "cibadmin -Q | grep validate | sed -e 's/.*validate-with=\"\\([^\"]*\\)\".*$/\\1/'"
+
+execute "upgrade CIB syntax version" do
+  user "root"
+  command "crm configure upgrade force"
+  not_if { Mixlib::ShellOut.new(cib_cmd).run_command.stdout.chomp.split("-").last > "2.0" }
```
vuntz: Can we make that 2.0 value an attribute in `pacemaker/attributes/default.rb`? This would make it easier to have it different depending on the distro.

Something like `default[:pacemaker][:cib_validate_version] = "2.0"`

chef/cookbooks/pacemaker/recipes/setup.rb
```diff
@@ -38,3 +38,13 @@
   subscribes :run, resources(template: crm_conf), :immediately
   action :nothing
 end
+
+# The output of "cibadmin -Q | grep validate" looks like:
+# <cib epoch="41" num_updates="9" admin_epoch="0" validate-with="pacemaker-1.2" ...>
+cib_cmd = "cibadmin -Q | grep validate | sed -e 's/.*validate-with=\"\\([^\"]*\\)\".*$/\\1/'"
```
vuntz: That grep is scary. Can you do `cibadmin -Q -n` instead?

chef/cookbooks/pacemaker/recipes/setup.rb
vuntz: Btw, can we run this on the founder only?

jsuchome: AFAIK this recipe is executed on founder only

vuntz: You're right.

Re the "2.0": on top of moving it to an attribute, should it be something like `>= "2.4"`?

jsuchome: Yes, I could bump it 2.4. I put 2.0 there based on the value found in Ha guide

vuntz:
If you're fine with it, can you merge https://github.com/jsuchome/crowbar-ha/pull/2 and then squash the commits? Rationale for the changes:

 - no migration is needed the attribute should stay a default attribute outside of the proposal
 - I find the gsub way more readable than the sed
 - the > to >= is actually important :-)

If you think the sed way is better, fine with me -- then please just update the PR for the two other bits :-)

vuntz:
Since there'll be a change, can you prefix the commit message with "pacemaker: "? We try to use the right prefix for each commit in this module :-)

vuntz:
+1

aplanas:
+1

vuntz:
This is hitting https://trello.com/c/30w2SCA0/128-c6-ha-waiting-for-testvm-booted-and-ping-returned-timed-out-in-testsetup-step and manila failures in tempest, but deployment and rest tempest were fine; so good enough for merging.

PR #84, opened 2016-02-07
crowbar-pacemaker: Sort nodes returned by cluster_nodes/remote_nodes

vuntz:
This is not needed, but is slightly nicer since when we run commands on
each node, then we know the order to expect.


vuntz:
This is hitting https://trello.com/c/30w2SCA0/128-c6-ha-waiting-for-testvm-booted-and-ping-returned-timed-out-in-testsetup-step and manila failures in tempest, but deployment and rest tempest were fine; so good enough for merging.

Any +1?

aplanas:
+1

jsuchome:
+1

PR #87, opened 2016-02-09
crowbar-pacemaker: Fix fencing of remote nodes on creation of primitive

vuntz:
We were creating the primitive with [:create, :start]. But the devil is
in the details: with :create, we force the primitive to be created with
target-role="stopped", which results in pacemaker ensuring that the
resource is stopped. That means fencing.

So instead, we put the primitive in a transaction, which leads to the
primitive to be started automatically.


toabctl:
+1

vuntz:
This is hitting https://trello.com/c/NpImhTvk/150-c6-binding-failed-ret-102 and manila failures in tempest, but deployment and rest tempest were fine; so good enough for merging.

But still missing one +1 :-)

jsuchome:
+1

PR #88, opened 2016-02-11
crowbar: Fix migrating adding drbd secret

vuntz:
We were not generating a new secret on migration, resulting in an empty
secret.


toabctl:
+1

aplanas:
+1

vuntz:
Deployment worked, but we hit the binding failed issue in neutron. Since it's a migration, deployment is good enough for merging.

PR #90, opened 2016-02-19
avoid spurious connections to pacemaker_remoted (bsc#967388)

aspiers:
Pacemaker currently makes it too easy to take over a remote, simply by connecting to its TCP port.  So until that bug is fixed, we should avoid making an actual connection, and instead simply check via `lsof` that the port is being listened on.

https://bugzilla.suse.com/show_bug.cgi?id=967388


aplanas:
+1

aspiers:
We have tested this and proven that it no longer kills the connection between the remote and the corosync node.  However given we're already at GMC1 we should also test that it doesn't break initial setup of the remotes.

vuntz:
See my comment on the bug; I'm not against it, but surprised that current behavior is causing fencing.

vuntz:
+1, works (but missing -s tcp:listen, for a followup PR).

PR #91, opened 2016-02-19
haproxy_loadbalancer: add check parameter

aplanas:


chef/cookbooks/haproxy/providers/loadbalancer.rb
```diff
@@ -57,6 +57,7 @@
     end
   end
   section["servers"] = new_resource.servers
+  section["check"] = new_resource.check.map{ |k, v| "#{k} #{v}" }.join(" ")
```
vuntz: Please use a `sort` somewhere (on the keys) to make sure that this always generates the string in the same order.

aplanas: I will add the `sort`, but are not the keys in Ruby > 1.8 ordered by the insertion order? I was reading an article that argument that this is something that the developer can relay on.

aplanas:
* Sort hash map before iterating

chef/cookbooks/haproxy/resources/loadbalancer.rb
```diff
@@ -27,4 +27,5 @@
 attribute :mode,    kind_of: String,  default: "http", equal_to: ["http", "tcp", "health"]
 attribute :use_ssl, kind_of: [TrueClass, FalseClass], default: false
 attribute :options, kind_of: Array,   default: []
+attribute :check,   kind_of: Hash,    default: { inter: 2000, rise: 2, fall: 5 }
```
abel-navarro: I think we should be more aggressive when doing health checks: faster checks and failing faster. At least I would use the same as the http case:
`inter 1000 downinter 3000 rise 3 fall 1`. I would even set `inter` to `250`, to do four checks per second.

aplanas: Ahmm, but this is the default for the rest of the services. In https://github.com/crowbar/crowbar-openstack/pull/298/files#diff-dbdf516dd90cb53427a4489f82729b58R22 you can see the specific value for horizon.

abel-navarro: I agree this is the default, what I'm saying is that it is worth decreasing these values for all services. At least in my trials they worked correctly.

I understand this may be a big change considering we also have to back port it. It may pay instead to be able to configure this in the raw view. This way we can keep the old values but also offer the possibility to decrease them in order to have better response times.

aplanas:
* Use same checks for SSL

chef/cookbooks/haproxy/resources/loadbalancer.rb
abel-navarro: as discussed with @aplanas , we'll move the change to the other services into another task.

abel-navarro:
+1

vuntz:
Can you update the PR to make sure that the hash contains inter/rise/fall keys? Or are they all optional?

aplanas:
@vuntz according to [1], `rise`, `fall`, and more are optionals. So is valid a single `check` without options. For example, in the case of `rise` [2], the default value is `2` if no other value is specified

[1] https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#5.2-check
[2] https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#rise

vuntz:
Cool. Followup question: do you think it's worth checking that the hash only contains valid options (and nothing called foobar), or we ignore that?

(I'd be fine ignoring, just asking)

aplanas:
@vuntz not sure ... , I will prefer to ignore it (as we do for the rest of the options).  Also the available `checks` can change with haproxy versions too

mjura:
LGTM +1

Itxaka:
LGTM +1

aplanas:
does this still makes sense? the change looks idempotent, so can be safe to merge (still missing some +1)

abel-navarro:
This change allows an easier configuration of the check parameter in haproxy config. This alone is ok, but the main point is it was a helper for https://github.com/crowbar/crowbar-openstack/pull/298. That one was the original motivation for changing the check parameter in order to have more aggresive check times. In my opinion we would benefit from faster checks on the services (i think we had some tempest tests failing because of this).

Anyway, merge this one if we merge also the other.

aplanas:
@abel-navarro can you +1 this? I need one more, github doesn't allows me to merge without it : ))

rsalevsky:
But pls merge after the Milestone is out.

chef/cookbooks/haproxy/providers/loadbalancer.rb
aspiers: > I will add the sort, but are not the keys in Ruby > 1.8 ordered by the insertion order?

Correct: https://ruby-doc.org/core-2.1.2/Hash.html

> I was reading an article that argument that this is something that the developer can relay on.

Why not sort by the key?  Wouldn't that make a lot more sense?

chef/cookbooks/haproxy/providers/loadbalancer.rb
```diff
@@ -57,6 +57,7 @@
     end
   end
   section["servers"] = new_resource.servers
+  section["check"] = new_resource.check.sort.map{ |k, v| "#{k} #{v}" }.join(" ")
```
aspiers: As mentioned, I suspect sorting by the key would be more stable than by insertion order.

PR #92, opened 2016-02-23
only check for Pacemaker remote TCP port in LISTEN state

aspiers:
This eliminates the very small corner case that an unexpected connection
was established from one remote node to another, e.g. manually for
testing.

### DIFF
--- a/chef/cookbooks/pacemaker/recipes/remote.rb
+++ b/chef/cookbooks/pacemaker/recipes/remote.rb
@@ -56,7 +56,7 @@ ruby_block "wait for pacemaker_remote service to be reachable" do
       # but this would require building a template for
       # /etc/sysconfig/pacemaker on remote nodes, so it's not worth
       # the effort until we really need it.
-      cmd = "lsof -i tcp:3121 >/dev/null"
+      cmd = "lsof -i tcp:3121 -sTCP:LISTEN >/dev/null"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/pacemaker/recipes/remote.rb
```diff
@@ -56,7 +56,7 @@
       # but this would require building a template for
       # /etc/sysconfig/pacemaker on remote nodes, so it's not worth
       # the effort until we really need it.
-      cmd = "lsof -i tcp:3121 >/dev/null"
+      cmd = "lsof -i tcp:3121 -sTCP:LISTEN >/dev/null"
```
vuntz: Maybe a space after `-s`?

vuntz:
+1

chef/cookbooks/pacemaker/recipes/remote.rb
vuntz: btw, I was doing that with tcp:listen (lowercase) in my tests.

aspiers: I thought it was safer to stick to the exact syntax given in the man page.

aplanas:
+1

vuntz:
Can you backport to 3.0?

PR #94, opened 2016-02-23
Handle DRBD in degraded cluster better

aspiers:
I'm not 100% sure this is the right approach yet, but it's probably better than nothing.  I don't get why the fenced founder loses its `pacemaker-cluster-member` role, since it's showing in Ready state (despite `crowbar_join` bailing at the `/var/spool/corosync/block_automatic_start` test).  (I can't find anything in any logs explaining why the role was removed from node1.)  But that means that the Chef search on the non-founder to find a DRBD peer fails to find the founder.  At this point I'm not sure whether it's better to keep the previous value of `remote_host` (as this PR does), or fail the chef-client run.  But either way we need to improve the error handling.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/providers/drbd.rb
+++ b/chef/cookbooks/crowbar-pacemaker/providers/drbd.rb
@@ -44,7 +44,13 @@ action :create do
     dirty ||= true if resource["master"] != is_master
 
     if dirty && resource["configured"]
-      raise "Configuration for DRBD resource #{name} has changed. If this is really wanted, please manually set node['drbd']['rsc']['#{name}']['configured'] to false with knife; the content of the DRBD resource will be lost!"
+      fmt = "%s / %s / %s"
+      old = fmt % [resource["fstype"], resource["remote_host"], resource["master"]]
+      new = fmt % [fstype, remote_host, is_master]
+      raise "Configuration for DRBD resource #{name} has changed from " \
+            "#{old} to #{new}. If this is really wanted, please manually " \
+            "set node['drbd']['rsc']['#{name}']['configured'] to false with " \
+            "knife; the content of the DRBD resource will be lost!"
     end
 
     node["drbd"]["rsc"][name]["lvm_size"] = lvm_size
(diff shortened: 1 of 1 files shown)


vuntz:
Generally LGTM. Although I'm pretty sure the issue would not happen if we had https://github.com/crowbar/crowbar-ha/pull/34 and https://github.com/crowbar/crowbar-core/pull/100.

aspiers:
Yeah, that's what I was thinking too.  I'd still like to understand how the role removal happened though ...

vuntz:
+1; anyone else with a +1?

aspiers:
Do we want to backport this?

vuntz:
> Do we want to backport this?

Yes. Was waiting for the merge to ask for it, though :-)

aplanas:
+1

aplanas:
Makes sense to revert this once both PR referenced in @vuntz are merged? A hard fail if this issue reappears during Cloud 7 will give the chance to revisit a debug it.

vuntz:
> Makes sense to revert this once both PR referenced in @vuntz are merged?

Agreed. I added a comment to the other PR for this.

aspiers:
The whole commit should definitely *not* be reverted.  The improved logging is useful.  The only bit which is arguably worth reverting is how it keeps the previous value of `remote_host` instead of raising an exception.

But I fundamentally disagree with the approach of "it's better to raise an exception so we can debug it" when there is the possibility to make the code robust.  I think a better approach would be to automate the process of scanning the logs for serious warnings, and then building some health reporting capabilities based on that.  It could even be an extra "Health Check" screen in Crowbar which shows things which were noticed as cause for concern, but have not crashed Crowbar with chef failures etc. because we were smarter than that and used defensive programming :-)

BTW this approach would tie in perfectly with the Cloud 7 themes currently being discussed :)

aplanas:
Well, the missing part of the quote string is "debug it ... during Cloud 7 development". Also there is a frontier between defensive programming and hiding bugs doing extra-assumptions based on external heuristics, and sometimes this border-line is no clear for me (but I am confident that this is not the case)

I completely agree with the point improve the logging analysis: +1

aspiers:
Agreed, the line is not always clear!  Thanks for helping give me this idea though!  Filed as https://trello.com/c/r09uuhvj

aspiers:
I think it's now safe to change this `raise` to a warning, right?  As mentioned in https://github.com/crowbar/crowbar-ha/pull/34#issuecomment-252604950

PR #95, opened 2016-03-02
crowbar-pacemaker: Add sync marks to apache HA setup

vuntz:
The mod_status configuration happening at the beginning of this recipe
needs to be run on all nodes before setting up the apache HA resources.

This is quite important when upgrading from tex, as we might have an old
version of apache before the upgrade with a different syntax in the
mod_status config file. If we don't have sync marks there, then the
resource might start on nodes where the mod_status config file was not
updated, resulting in apache just failing to start due to invalid
syntax.

(And in general, it's the right thing to do anyway)

It's worth mentioning that the revision we're picking here is "good
enough", but not fully correct: technically, we don't know which
barclamp is using apache so we don't know which revision to use; but the
first time we set apache up, this sync mark will be empty so the sync
mark magic will happen anyway.

**Goes with https://github.com/crowbar/crowbar-openstack/pull/316**; we probably need both patches in the gating.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
@@ -47,6 +47,18 @@ if agent_name == "ocf:heartbeat:apache"
   end
 end
 
+# Wait for all nodes to reach this point so we know that all nodes will have
+# all the required packages installed before we create the pacemaker
+# resources
+crowbar_pacemaker_sync_mark "sync-apache_before_ha" do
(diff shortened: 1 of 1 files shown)


aspiers:
+1 (I assume it works when you also have the other PR)

jsuchome:
Failed for missing sync mark, which likely indicates the need of the other patch.

+1

vuntz:
Re-triggered gating to include the other patch.

vuntz:
Gating is green with the two patches.

PR #98, opened 2016-03-09
interleave Apache clones to minimise disruption (bsc#965886)

aspiers:
**NEEDS TESTING**

(There is [a corresponding commit to crowbar-openstack](https://github.com/crowbar/crowbar-openstack/pull/330), in which there are many more clones than in this repository.)

By default, Pacemaker clones aren't interleaved.  This means that if Pacemaker wants to restart a dead clone instance, and there is an order constraint on that clone, it will do the same restart on all other nodes, even if all the others are healthy.

More details on interleaving are here:

 - https://www.hastexo.com/resources/hints-and-kinks/interleaving-pacemaker-clones/

This behaviour is far more disruptive than we want.  For example, in [bsc#965886 (L3: neutron-ha-tool not migrating dhcp networks during failover)](https://bugzilla.suse.com/show_bug.cgi?id=965886) we saw that when a network node dies and Pacemaker wants to stop the instance of cl-g-neutron-agents on that node, it also stops and restarts the same clone instances on the healthy nodes.  This means there is a small window in which there are no neutron agents running anywhere.  If neutron-ha-tool attempts a router migration during this window, it will
fail, at which point things start to go badly wrong.

In general, the cloned (i.e. active/active) services on our controller and compute nodes should all behave like independent vertical stacks, so that a failure on one node should not cause ripple effects on other nodes.  Apache is one example of that, and even though there aren't currently any ordering constraints which would require interleaving, we may add them in the future.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
@@ -73,7 +73,10 @@ transaction_objects << "pacemaker_primitive[#{service_name}]"
 clone_name = "cl-#{service_name}"
 pacemaker_clone clone_name do
   rsc service_name
-  meta ({ "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node) })
+  meta ({
+    "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node),
+    "interleave" => "true",
+  })
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end


vuntz:
+1

jsuchome:
+1

vuntz:
@aspiers is it still WIP? The trello card made me think it has been tested and validated.

aspiers:
Travis is failing due to use of a new Rake (11.0.1) which dropped support for `last_comment` which `rspec-core` uses: https://github.com/rspec/rspec-core/issues/2205

aspiers:
Travis should be fixed by rebasing on top of #106 when it's merged.

aspiers:
Ignore last comment - a force rebuild of the Travis CI was enough to get a newer rake which fixed the issue.

vuntz:
@aspiers can we drop the wip/donotmerge tags and merge this?

aspiers:
I'm not sure we actually tested Apache in failover scenarios with this yet, however. I didn't, at least.

abel-navarro:
Tested this on a cloud5 deployment. Could not say what was the _statu quo_ but seems to be in working condition and ready to merge.

In a two controller setup, when rebooting one of them the horizon connections will be immediately reset.
image
A few tens of seconds. Horizon will complain about an unexpected error.
image
And just a few seconds later Horizon will come back again at the last dialog the user was. I found in one of the tests that right after Horizon came back _usage_ and _limit_ information was unavailable. This was working on the next dialog I clicked so I consider this a temporary issue.
image
I did not measure the total time Horizon was off, but I'd say about 30 seconds.

PR #102, opened 2016-03-15
Make Pacemaker retry start actions if they fail (bsc#965886)

aspiers:
**DO NOT MERGE YET - this is a significant cluster-wide change which needs thorough failover testing, and may even be too invasive to backport to SOC6 or 5.**

If `neutron-ha-tool` fails to start (e.g. because `neutron-server` is not yet listening on a socket due the daemon still being in the startup phase) then we should retry a few times on the same node, rather than immediately migrating to the other node. We already have `migration-threshold` defaulting to 3, which makes sense in this case.

Also in general, retrying startup of services seems like a good idea. Retries happen immediately so it should not significantly slow down fail-overs and increase service downtime.  If in the future we only want to allow retries for certain resources, we can change the cluster-wide default `migration-threshold` to `1` (or `0`?) and then increase it for individual resources.

However this is not a fully reliable fix, since it's conceivable that the daemon could stay in the startup phase for longer than `(migration-threshold * number-of-nodes)` retries.  So we will also need to make start-up of `neutron-server` (and maybe other daemons) asynchronous, i.e. blocking until the API endpoint is actually usable.

https://bugzilla.suse.com/show_bug.cgi?id=965886

### DIFF
--- a/chef/cookbooks/pacemaker/templates/default/crm-initial.conf.erb
+++ b/chef/cookbooks/pacemaker/templates/default/crm-initial.conf.erb
@@ -1,6 +1,7 @@
 property $id="cib-bootstrap-options" \
         stonith-enabled="<%= @stonith_enabled %>" \
         no-quorum-policy="<%= @no_quorum_policy %>" \
+        start-failure-is-fatal=false \
         placement-strategy="balanced"
 op_defaults $id="op-options" \
         timeout="<%= @op_default_timeout %>" \


vuntz:
+1. But do you have a link to some documentation / something else that explains why it's the default and when it's fine to change? That'd be good to include in the commit message.

jsuchome:
Not even in commit message, maybe also as the comment right in the code...

aspiers:
> But do you have a link to some documentation / something else that explains why it's the default and when it's fine to change?

No, I don't.  Information on this option is hard to come by.  I only found stuff like this:

- http://blog.kennyrasschaert.be/blog/2013/12/18/pacemaker-high-failability/
- http://serverfault.com/questions/387425/make-pacemaker-retry-failed-resources
- http://thread.gmane.org/gmane.linux.highavailability.pacemaker/13531/focus=14897

vuntz:
@aspiers I'd add exactly what you wrote in the commit message. Sure none of these are really authorative, but it's useful background info (even the fact that it's hard to come by is useful background info).

abel-navarro:
+1 The change makes sense to me, although I agree with @aspiers that this may have some implications. Further testing is encouraged.

aspiers:
Gate failed with some `cct` crap, trying a rebuild ...

aspiers:
@vuntz Commit message expanded.

aspiers:
I've also just asked for feedback on `#clusterlabs` IRC channel.

aspiers:
OK, just got an answer from Ken Gaillot - commit message updated again accordingly.

abel-navarro:
I tried this option in a cloud5-ha deployment. Before this change neutron-ha-tool was being reported with an error. After the change, you could reboot any of the two controller nodes and no error message from neutron-ha-tool was observed.

vuntz:
Thanks for the changes, Adam. +1

jsuchome:
+1

aspiers:
@vuntz Do you think we should test more before merging? I'm not sure. I guess we can easily revert if we get a nasty surprise.

vuntz:
> @vuntz Do you think we should test more before merging? I'm not sure. I guess we can easily revert if we get a nasty surprise.

We should just merge for Cloud 7 and see how it goes. I'd require more testing for backporting, though.

PR #106, opened 2016-03-24
pin rake to 10.x

aspiers:
Older `rspec-core` versions use `#last_comment`, which `rake` 11.x removed support for.  We're still waiting for the `rspec-core` fix to land, since 3.5.x is still in pre-release:

  rspec/rspec-core#2205

so in the mean time, let's stick with rake 10.x.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ end
 
 unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"
   group :test do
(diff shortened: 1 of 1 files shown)


aspiers:
Oh, looks like the rake folks realised that it would be a good idea to deprecate `#last_comment` by removing it:

https://github.com/ruby/rake/commit/3948349a82a2522f1d8293093164ef395bee542d

so we only need to make sure we have rake >= 11.1.0.

PR #109, opened 2016-04-19
Added http-keep-alive option to haproxy defaults

abel-navarro:
We were using the http-request for haproxy with a 5s timeout. This
option only addresses the initial part of the HTTP negotiation. Once
the HTTP session is established we should use a shorter timeout
for idling connections with the http-keep-alive option. This option
if not set gets the same value as http-request timeout.

This patch will help fix problems with 'nova boot --poll', because
the connections usually ended at the same time a new request was
coming, leading to connection refused errors.

References:
- https://trello.com/c/m9a3MQYS
- https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4.2-timeout%20http-keep-alive

### DIFF
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -17,7 +17,8 @@ defaults
 	option  dontlognull
 	option  redispatch
 	retries 3
-	timeout http-request 5s
+	timeout http-request 10s
+	timeout http-keep-alive 10s
 	timeout queue 1m
 	timeout connect 5s
 	timeout client 3h


abel-navarro:
@aplanas @nicolasbock

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -18,6 +18,7 @@ defaults
 	option  redispatch
 	retries 3
 	timeout http-request 5s
+	timeout http-keep-alive 1s
```
aplanas: > In general it is optimal to set this value to a few tens to hundreds of
milliseconds, to allow users to fetch all objects of a page at once but
without waiting for further clicks.

Maybe we are more aligned to the documentation if we use `250ms`? Also we need this because in other case the `http-keep-alive` valie will be `5s`, isn't?

abel-navarro: I wanted to use a conservative value. You are right that at present it's 5s, so 1s it's a great improvement. I could try with smaller values though and see what happens.

nicolasbock: I second @aplanas comment: Let's try out what happens if we use a smaller value.

abel-navarro: I run more than 60 tests with 250ms timeout: 1 failure was observed
I run more than 60 tests with 500ms timeout: 1 failure was observed
I run more than 160 tests with 1s timeout: no failure was observed

All of this in virtual environments (mkcloud2). I reckon with physical hardware the response times may be better, but since this is affecting our gates, I'd leave it to 1s. Furthermore, this value was before 5s, so it is an improvement over the _statu quo_.

abel-navarro:
Related PRs:

- cloud5: https://github.com/crowbar/barclamp-pacemaker/pull/222
- cloud6: https://github.com/crowbar/crowbar-ha/pull/110

aplanas:
+1

nicolasbock:
:+1:

abel-navarro:
I left the test run with the 1s timeout. It run 391 times without failures. Then the VPN broke :)

nicolasbock:
@abel-navarro, nice testing :) But it sounds like this change addresses the issues we are seeing.

:+1: :+1:

vuntz:
I'm fine with the change, but I'm not 100% sure I fully understand why this fixes the problem. Would it be possible to somehow clarify the commit message for this?

(also, can we run the gating for HA in case it wasn't the case -- didn't check)

abel-navarro:
I hope this drawing helps clarifying the issue. It's a simplistic view but I think highlights the problem.

http-keep-alive

aplanas:
+1 for the drawing

kbaikov:
+1 for the word "kaput" in the drawing. And drawing saved for the history.

nicolasbock:
I don't understand why it's not working if you set `http-keep-alive < 1s`. If I understand your drawing (which is awesome, by the way) correctly, then a smaller value should work as well.

abel-navarro:
@nicolasbock you are right. We were indeed discussing this with @aplanas a few moments ago. The reason it's not completely clear to me, but my guess is that if you shorten too much the `http-keep-alive` option you may be sending a TCP reset before novaclient has finished communication and then you are in the same situation as before.

Note that novaclient requests are not atomic. In this drawing I just put a single line for `servers.get`, but there may be several IP packets involved. Note also that this is a virtual environment and from time to time VMs _freeze_ due to shared resources. This may lead to create a gap enough to trigger the early FIN. These are but guesses since I didn't look that deeply at this other case.

aplanas:
@abel-navarro To have a better idea of this issue, can we test it with `1ms`, or is too late?

abel-navarro:
@aplanas I rebuild my environment to move to other tasks. I guess when it's completed I can run those tests again with 1ms.

nicolasbock:
@abel-navarro Do you still have the tcpdumps you mentioned? Would they reveal more about the traffic between novaclient and haproxy?

abel-navarro:
@nicolasbock I recovered that capture file and removed packets from other flows. You will see at the end of the capture the FIN/RST problem.
[http-keep-alive-RST.zip](https://github.com/crowbar/crowbar-ha/files/227689/http-keep-alive-RST.zip)

nicolasbock:
@abel-navarro: @aplanas and I looked over the trace, and it really looks like as if we are seeing an unfortunate accident: At the end of your trace the server is closing the socket (frame 37), the client sends another request (frame 38), and then (!) acknowledges the closing of the connection (frame fram 40), and the http request fails. I don't see how any value of `http-keep-alive` can fix this. It seems that the only proper way of handling this on the client side is to retry the attempt.

vuntz:
Shouldn't we just disable keep-alive? (we're doing that for the openstack services, actually)

aplanas:
Uhm... we can try `option http-tunnel` or `option forceclose`

abel-navarro:
@nicolasbock it is indeed an _accident_ and maybe you are right that it needs fixing in the novaclient. The fact is novaclient handles properly the FIN but not the RST. Should novaclient actually handle the RST? or better haproxy not send a RST at all? Novaclient doesn't expect an haproxy in the middle and receive RST.

@vuntz all services are using default values in haproxy, so if we don't set the the `http-keep-alive` it's going to use the `timeout http-request`. We could consider removing that option too. But while haproxy is tracking connections a timeout will be needed. I could do more investigation on this matter but we'd need to unblock the gates too.

abel-navarro:
@aplanas i think those options you mention could also help fixing this problem. But again, this would need further investigation and testing.

abel-navarro:
rebased

abel-navarro:
After the rebase, the CCT tests now fail. The error is the same: *ConnectionRefused*. I'm investigating this right now.

aplanas:
@abel-navarro do we know if `option forceclose` worked?

abel-navarro:
@aplanas didn't have the chance to try. With this new environment I cannot reproduce the issue yet.

abel-navarro:
I managed to reproduce the problem in one of the jenkins slaves. It is actually the same problem we were seeing before, but this time it moved to the backend. The client took exactly 1s to follow up on the conversation and haproxy closed the connection.

[backend-refused-trimmed.zip](https://github.com/crowbar/crowbar-ha/files/230271/backend-refused-trimmed.zip)

Next thing is to try those options @aplanas mentions, although before I want to do a bit more reading about the proper way to fix this. I'm trying enabling the stats module so we can have more insight when this problems occur.

abel-navarro:
Reducing `timeout http-keep-alive` to 1s has moved the error from the `nova boot --poll` (5 second timeout) to every single nova call.  In test virtual environments subsequent HTTP requests can take 1s or more and thus raising this issie.

After reading a bit more `haproxy` documentation I think the options `http-tunnel` or `forceclose` are not appropriate for our needs. In the first case that option is not the default anymore and is [not recommended](https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#option%20http-tunnel)  when using HTTP mode. The second one would send a TCP FIN after each HTTP request. This would cause some overhead in the client because it would need to create a new socket for each HTTP request.

The proposed solution is to keep a timeout but increased it's time to 10 seconds. [I have compared](http://etherpad.cloud.suse.de:8000/0kv5p2npry) how other's set this timeout and seems 10 seconds is the most common approach.

The value at `http-request` is the same as `http-keep-alive` if this is not set. I preferred to explicitly declare both arguments to make clear it was a conscious decision to set it.

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -17,7 +17,8 @@ defaults
 	option  dontlognull
 	option  redispatch
 	retries 3
-	timeout http-request 5s
+	timeout http-request 10s
+    timeout http-keep-alive 10s
```
aplanas: tab?

abel-navarro: oops! fixing that.

aplanas:
Actually I do not see in the documentation that other options are not recommended. The default was changed, but still makes sense in some scenario.

Are you able to reproduce the issue with this 10s? Do you have different results with `forceclose`?

abel-navarro:
I did not try the `option http-tunnel` because we need HTTP for the checks and for this comment in the documentation:

> - TUN: tunnel ("option http-tunnel") : this was the default mode for versions
    1.0 to 1.5-dev21 : only the first request and response are processed, and
    everything else is forwarded with no analysis at all. This mode should not
    be used as it creates lots of trouble with logging and HTTP processing.

The `forceclose` I did not try because creating sockets for every connection looks quite expensive to me.

About the 10 seconds timeout, I did some tries and worked properly.

abel-navarro:
@jdsn @MaximilianMeister mind reviewing this PR? it would help making the gates greener.

jdsn:
Just for the record:
I understand that this PR is just a workaround that makes it just less likely that the bug will be hit. The real bug is still not addressed.

My recommendation:
Create upstream bug for this and then add a comment around the two config lines with a short explaination that this is a workaroud for upstream bug XYZ. It should be visible that these lines should be adapted once upstream merged a proper fix.

abel-navarro:
Although I initially agreed with @jdsn for this being a bug, I think this is in fact an expected behavior for any system dealing with TCP-RST. If we didn't had **haproxy** between *nova-client* and *nova-server*, we would expect *nova-client* to return an error when *nova-server* sends a TCP-RST. That would mean the server process has ended or that the server socket has been closed. In either case *nova-client* has to treat that as an exception.

Now with **haproxy** in the middle we are considering *nova-client* should deal with TCP-RST, may be by retrying the communication. In don't think that's the right behavior because we are dealing here with **haproxy** own behavior. From my point of view **haproxy** should be as transparent as possible.

**haproxy** configuration docs highly recommend using timeouts when services are configured in **http** mode. Otherwise connections may be open for a very long time and become a resource hog. Having timeouts means at some point **haproxy** will close the communication (TCP-FIN) and if at that same moment comes a new request, a TCP-RST will be sent, because the socket was already closed.

There is always going to be a risk of that happening regardless of the chosen timeout value. If two HTTP requests are spaced out exactly the same amount of time as the timeout, this may happen again. It is very unlikely though that nova-client leaves the socket open for single commands, as those can be resolved with few HTTP requests and the socket is closed almost immediately. The offending case for this issue was with `nova boot --poll`, because this command leaves the socket open and HTTP requests were spaced out exactly 5 seconds, the same as `haproxy` timeout. So moving to a 10s timeout should avoid those situations at all.

aplanas:
+1

jsuchome:
@rhafer don't you have any comment?

rhafer:
@abel-navarro, @jsuchome: In general I agree with the change, but can't a single ```nova boot --poll``` call take easily much longer than 10 seconds? E.g. when it needs to fetch a some large image from glance?

Isn't setting ```timeout http-request 10s``` and ```timeout http-keep-alive 10s``` redundant if they have the same value?

BTW, what's the reason for increasing ```timeout http-request``` I guess just using a larger value for ```http-keep-alive``` would be enough, right?

aplanas:
@rhafer let me answer some points:

> can't a single ```nova boot --poll``` call take easily much longer than 10 seconds?
This is not the problem, but that every poll is spaced by 5s, so the keep-alive will be reset to be 10s more.

> Isn't setting ```timeout http-request 10s``` and ```timeout http-keep-alive 10s``` redundant if they have the same value?
Yep, but I agree with @abel-navarro to be explicit here. The original error was because http-keep-alive was implicitly set to 5s because of http-request. But is indeed redundant.

> BTW, what's the reason for increasing ```timeout http-request``` I guess just using a larger value for ```http-keep-alive``` would be enough, right?

Yes. The decision to change http-request from 5s to 10s is because other projects are using those values. We align more with other solutions like Fuel.

abel-navarro:
@rhafer `nova boot --poll` can effectively take more than 10 seconds to complete. `nova boot --poll` does the polling every 5 seconds (hardcoded). It's the time between polls (http requests) that `timeout http-keep-alive` takes care about.

rhafer:
@abel-navarro @aplanas thanks for clarifying

+1

PR #110, opened 2016-04-19
[3.0] Added http-keep-alive option to haproxy defaults

abel-navarro:
We were using the http-request for haproxy with a 5s timeout. This
option only addresses the initial part of the HTTP negotiation. Once
the HTTP session is established we should use a shorter timeout
for idling connections with the http-keep-alive option. This option
if not set gets the same value as http-request timeout.

This patch will help fix problems with 'nova boot --poll', because
the connections usually ended at the same time a new request was
coming, leading to connection refused errors.

References:
- https://trello.com/c/m9a3MQYS
- https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4.2-timeout%20http-keep-alive

### DIFF
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -17,7 +17,8 @@ defaults
 	option  dontlognull
 	option  redispatch
 	retries 3
-	timeout http-request 5s
+	timeout http-request 10s
+	timeout http-keep-alive 10s
 	timeout queue 1m
 	timeout connect 5s
 	timeout client 3h


aplanas:
+1

nicolasbock:
:+1:

tboerger:
:+1:

aplanas:
I am not sure that this is the proper solution, but because is tested +1 for me.

Also if this is not the correct solution I agree to align more with the configuration in other projects.

abel-navarro:
@jdsn @MaximilianMeister mind reviewing this PR? it would help making the gates greener.

jdsn:
Just for the record:
I understand that this PR is just a workaround that makes it just less likely that the bug will be hit. The real bug is still not addressed.

My recommendation:
Create upstream bug for this and then add a comment around the two config lines with a short explaination that this is a workaroud for upstream bug XYZ. It should be visible that these lines should be adapted once upstream merged a proper fix.

PR #111, opened 2016-04-20
Fix for updated rubocop version

abel-navarro:


rsalevsky:
@abel-navarro Can you check if we need this also for the other repositories?

abel-navarro:
@rsalevsky we will surely need this for other repos. I'll do the porting once this once is accepted.

tboerger:
:+1:
I will merge it directly as this is unblocking and not a functional change. This is a documented behavior of current rubocop versions.

PR #113, opened 2016-04-21
test - do not merge

abel-navarro:
testing rubocop and codeclimate.


abel-navarro:
test completed. Codeclimate works ok without the lastest change added in #111. I worked on my code with an old version of master and that caused codeclimate to fail. The change seems to work, but probably it does not make sense to port it to other repositories.

tboerger:
The rubocop config change needs to be ported to the other repos as well.

PR #116, opened 2016-06-08
don't tell systemd to enable drbd (bsc#971771)

aspiers:
We only want DRBD started by Pacemaker, with the possible exception of during initial DRBD setup.  If it gets started by `systemd`, `systemd` will believe it owns the service, in which case during system shutdown it will prematurely shut down DRBD without regard for any of the other services and resources depending on it.  Instead we want Pacemaker to stop things in the correct order, even taking care of inter-node dependencies.
- https://bugzilla.suse.com/show_bug.cgi?id=971771

This fix will not be sufficient by itself, since we still need to address:
- https://bugzilla.suse.com/show_bug.cgi?id=980341

### DIFF
--- a/chef/cookbooks/drbd/recipes/default.rb
+++ b/chef/cookbooks/drbd/recipes/default.rb
@@ -28,6 +28,6 @@ service "drbd" do
   if node["drbd"]["rsc"].empty?
     action :nothing
   else
-    action [:enable, :start]
+    action [:disable, :start]
   end
 end


chef/cookbooks/drbd/recipes/default.rb
```diff
@@ -28,6 +28,6 @@
   if node["drbd"]["rsc"].empty?
     action :nothing
   else
-    action [:enable, :start]
+    action [:disable, :start]
```
abel-navarro: wouldn't it make sense to set it to just `[:disable]`? we don't want drbd started by anyone else than pacemaker.

aspiers: Like I said in the commit message, I think initial startup is the exception where we need to start it by ourselves first.  I'm basing that on my memory of a conversation with @vuntz after I discovered this bug.

vuntz: So I believe we indeed need to start drbd the first time, but we shouldn't do that here like that, as it's a bit too unconditional... We probably find a way to say "it's the first time and only the first time so start it".

aspiers: @vuntz So you are suggesting to change this to `[:disable]` and add the start elsewhere, e.g. via `execute 'rcdrbd start'` or similar?

vuntz: I'd use an immediate notification, but yes.

aspiers: @vuntz OK.  Is this the place? https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/drbd/providers/resource.rb#L71

aplanas:
@aspiers @vuntz  any follow up to this?

aspiers:
@aplanas No, nothing new.

aplanas:
@aspiers so still WIP and do not merge yet are valid here. Ok, thanks! : )

vuntz:
https://github.com/crowbar/crowbar-ha/pull/164 has https://github.com/crowbar/crowbar-ha/pull/164/commits/52d11b7d94cc0f0eb14be3462b7cd2d6814e41ac which obsoletes this. But I think we need more (I started working on this in https://github.com/crowbar/crowbar-ha/pull/165; should resume this)

PR #117, opened 2016-06-17
Switch from closure-compiler to uglifier

rsalevsky:
Since we switched to SLE12 we can use the faster uglifier which uses
NodeJS.


rsalevsky:
The gating will fail due to dependencies issues which needs to be adapted when this PR is merged.

MaximilianMeister:
+1

nicolasbock:
+1

PR #119, opened 2016-06-24
crowbar-pacemaker: Use Barclamp::Inventory.Network objects

vuntz:
We do this instead of accessing attributes, in order to centralize the
location of the network definitions, so it can be changed.

Should go with https://github.com/crowbar/crowbar-core/pull/510

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
@@ -108,10 +108,11 @@ when "ipmi_barclamp"
     end
 
     stonith_node_name = pacemaker_node_name(cluster_node)
+    bmc_net = Barclamp::Inventory.get_network_by_type(cluster_node, "bmc")
 
     params = {}
     params["hostname"] = stonith_node_name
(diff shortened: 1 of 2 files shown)


rhafer:
Shouldn't this also go away: https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb#L54 ?

vuntz:
> Shouldn't this also go away: https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb#L54 ?

Not in this PR, but in https://github.com/crowbar/crowbar-ha/pull/120. Not sure why I didn't catch it there, though. Will copy your comment there.

rhafer:
> Not in this PR, but in #12

Ah, ok. Didn't see that one.

+1 once https://github.com/crowbar/crowbar-core/issues/510 is merged.

vuntz:
> +1 once crowbar/crowbar-core#510 is merged.

Actually, it doesn't depend on the crowbar-core one, so we could merge it as soon as we have two +1.

toabctl:
+1

PR #120, opened 2016-06-27
crowbar, crowbar-pacemaker: Add helpers to fetch the VIP of a hostname

vuntz:
We have code in several locations that does this, and it feels better to
have a helper to not have too many copy & paste. It will also make it
easier to change the implementation of how IP assignments are stored.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -106,6 +106,22 @@ module CrowbarPacemakerHelper
     end
   end
 
+  # Returns the VIP matching a specific virtual hostname used by a cluster.
+  # Note that vhostname is not a FQDN, but a short hostname.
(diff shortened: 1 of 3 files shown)


nicolasbock:
+1

vuntz:
@rhafer mentioned https://github.com/crowbar/crowbar-ha/pull/119#issuecomment-230771215 which we should fix here.

vuntz:
> @rhafer mentioned #119 (comment) which we should fix here.

done

abel-navarro:
+1

nicolasbock:
+1

aspiers:
Hmm.  I think I'd probably give a retroactive -1 for this: doesn't it introduce a regression of 164397d36?  If the data bag is missing or corrupt for some reason, haproxy setup will now fail with a cryptic error because you dropped the `raise` clauses and ignore the case where `cluster_vip` unexpectedly returns `nil`.

PR #122, opened 2016-06-27
crowbar: Update for NodeObject.add_to_run_list changes

vuntz:
The states parameter is now useless, so drop it.

See crowbar/crowbar-core@84827646

Depends on https://github.com/crowbar/crowbar-core/pull/515

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -134,7 +134,6 @@ class PacemakerService < ServiceObject
 
       deployment = cluster_role.override_attributes[cluster_role.barclamp]
       runlist_priority_map = deployment["element_run_list_order"] || {}
-      role_map = deployment["element_states"] || {}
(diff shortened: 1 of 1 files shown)


toabctl:
beside the commit message, +1

vuntz:
> @vuntz From the commit msg: The states parameter is now useless, so drop it.. Can you clarify what now means? mayb with a commit sha?

@toabctl I'm happy to do it, but a stupid question: do we have a standard way of referring to commits in other repos? Or should I just do "see sha in crowbar-core"?

toabctl:
> @toabctl I'm happy to do it, but a stupid question: do we have a standard way of referring to commits in other repos? Or should I just do "see sha in crowbar-core"?

@vuntz Maybe ```see crowbar/crowbar-core@9f06b04``` which at least links in github and is imo human readable.

vuntz:
Updated the commit message as suggested, no code change.

PR #124, opened 2016-06-30
corosync: Rework "block autostart" to not rely on chef-client on boot

vuntz:
Right now, the feature to block autostart of corosync was relying on
chef to start corosync (or pacemaker, on newer platforms) on boot.

In order to move away from the requirement of having chef-client running
on boot, this is reworked so that this relies on a new service (for
sysvinit) or a overridden corosync service (on systemd) for the whole
logic. This way, the feature work just fine without any chef run on
boot.

For SLE12, it also means we can always enable pacemaker now, since it
will fail to start due to corosync if needed.


chef/cookbooks/corosync/recipes/service.rb
```diff
@@ -127,55 +148,63 @@
...
-      subscribes :run, resources(template: "/etc/init.d/#{corosync_shutdown}"), :delayed
+      subscribes :run, resources(template: "/etc/init.d/#{sysvinit_corosync_wrapper}"), :delayed
+    end
+
+    service sysvinit_corosync_wrapper do
+      action [:enable, :start]
     end
+
+    # we make sure that corosync is not enabled to start on boot
+    enable_or_disable = :disable
```
vuntz: This is not true.

chef/cookbooks/corosync/recipes/service.rb
```diff
@@ -80,22 +80,45 @@
   password node[:corosync][:password]
 end
 
-# If this file exists, then we will require that corosync is either manually
-# started or that the file gets removed to have chef-client start corosync.
+# If this file exists, then we will require that corosync that the file gets
```
aplanas: Can we rephrase this?

"If this file exists, then we will require that the file gets removed to have corosync be able to start."

or something like that

chef/cookbooks/corosync/recipes/service.rb
```diff
@@ -80,22 +80,45 @@
...
+# install will remove the file, which will allow the wrapper/override to start
+# corosync on next boot.
 #
 # If the node goes down without the proper shutdown process (it has been
 # fenced, or it lost power, or it crashed, or...), then the file will exist
 # and corosync will not start on next boot, requiring manual intervention.
 block_corosync_file = "/var/spool/corosync/block_automatic_start"
-corosync_shutdown = "#{node[:corosync][:platform][:service_name]}-shutdown-cleaner"
+
+sysvinit_corosync_wrapper = "#{node[:corosync][:platform][:service_name]}-wrapper"
```
aplanas: Is this service now in the package? : )

chef/cookbooks/corosync/recipes/service.rb
```diff
@@ -80,22 +80,45 @@
...
+if File.exist?("/etc/init.d/#{old_corosync_shutdown}") ||
+    File.exist?("/etc/systemd/system/#{old_corosync_shutdown}.service")
+  service old_corosync_shutdown do
+    # There's no need to stop anything here.
+    action [:disable]
+  end
+  file "/etc/init.d/#{old_corosync_shutdown}" do
+    action :delete
+  end
+  file "/etc/systemd/system/#{old_corosync_shutdown}.service" do
```
aplanas: There is and 'or' in L100, I guess that both file exist, but this delete can fail if one of them exits (enter the `if` block) but the other not. But I guess that it can't happen?

chef/cookbooks/corosync/recipes/service.rb
```diff
@@ -106,17 +129,16 @@
   # If b) is not true, then corosync is already running, which either means
   # that we went through a) in an earlier chef run, or that the user manually
   # started the service (and acknowledged the issues with improper shutdown).
-  if ::File.exists?(block_corosync_file) && !system("crm status &> /dev/null")
+  if ::File.exist?(block_corosync_file) && !already_running
```
aplanas: IIUC is the wrapper the one that check the file and start the service? Why we are checking it here again?

chef/cookbooks/corosync/recipes/service.rb
vuntz: It's really a crowbar-specific feature, which is why it's not in the HA packages.

chef/cookbooks/corosync/recipes/service.rb
vuntz: Actually, `:delete` doesn't fail if the file doesn't exist. Chef only ensures that the file doesn't exist anymore, so it doesn't do anything if the file is missing.

(And to answer your other question: in general, only one or the other exists)

chef/cookbooks/corosync/recipes/service.rb
vuntz: To avoid chef-client starting things automatically due to the chef-client call on boot (and then later the periodic chef-client runs every 15 minutes). We want the user to explicitly fix this, as the goal is really to avoid a fencing loop, which means the user has to take action.

chef/cookbooks/corosync/recipes/service.rb
vuntz: Oops, good catch, fixed in new commit.

aplanas:
+1

aplanas:
Gating fail when applying Neutron. Rechek.

aplanas:
Gating green: +1

chef/cookbooks/corosync/recipes/service.rb
aspiers: Why in a new commit?  It should be squashed into this one, otherwise you are introducing a nonsensical comment and then fixing it all in the same PR.

vuntz:
@aspiers if you have time for this, that'd be great. Once we have this, we can actually stop running the pacemaker bits in crowbar_join, which will simplify things quite a bit when rebooting.

aplanas:
@aspiers any thoughts about this change?

chef/cookbooks/corosync/recipes/service.rb
```diff
@@ -127,55 +148,63 @@
...
         block_corosync_file: block_corosync_file
       )
     end
 
-    bash "reload systemd after #{corosync_shutdown} update" do
+    bash "reload systemd after #{systemd_corosync_override_dir}/crowbar.conf update" do
       code "systemctl daemon-reload"
       action :nothing
-      subscribes :run, resources(template: "/etc/systemd/system/#{corosync_shutdown}.service"), :immediately
+      subscribes :run, resources(template: "#{systemd_corosync_override_dir}/crowbar.conf"), :immediately
```
aspiers: Please split this too.  Currently Hound is not optional on this repo :)

vuntz:
I fixed hound warnings and clarified a comment as requested by @aspiers. You can see the changes in the last two commits -- I didn't change previous commits.

If gating passes, I'll merge this as there were approvals already.

vuntz:
https://ci.suse.de/job/openstack-mkcloud/43199/ failed in cct... Trying one more time.

PR #125, opened 2016-07-01
haproxy: Make the sections for haproxy config persistent on each node

vuntz:
Until now, we were saving the sections as a default attribute, so that
they would be removed on every new run, resulting in the config being
regenerated from scratch. The downside of this is that all active roles
that use a loadbalancer resource must always be used in the chef run,
otherwise we lose part of the config. And this goes against our goal of
running only the strict minimal recipes when required.

### DIFF
--- a/chef/cookbooks/haproxy/providers/loadbalancer.rb
+++ b/chef/cookbooks/haproxy/providers/loadbalancer.rb
@@ -61,3 +61,9 @@ action :create do
   node["haproxy"]["sections"][new_resource.type] ||= {}
   node["haproxy"]["sections"][new_resource.type][new_resource.name] = section
 end
+
+action :delete do
+  node["haproxy"]["sections"].keys.each do |type|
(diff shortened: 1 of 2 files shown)


nicolasbock:
+1

aspiers:
Gating failed but AFAICS for an unrelated reason, so I have triggered a rebuild. Assuming it passes, I'm +1. Presumably nothing will use `:delete` until we actually start properly implementing role removal, right?

aspiers:
The rebuild failed but it seems the failure has already been fixed by https://github.com/crowbar/crowbar-ha/pull/137.  I'm not sure where this new breakage is coming from, but let's see if another rebuild helps. I'm still +1 as soon as this passes the gate.

vuntz:
> Presumably nothing will use :delete until we actually start properly implementing role removal, right?

Correct.

vuntz:
Gating only failed because of the ceph testsuite not being ready for SES 4.

PR #127, opened 2016-07-06
Fix travis

vuntz:


toabctl:
typo in commit message (rack vs rake)

vuntz:
> typo in commit message (rack vs rake)

Hrm, I don't see it. Could it be that you were looking at the changes before I force-pushed again? (previous change was for rake, but didn't work)

rsalevsky:
+1

toabctl:
+1

PR #128, opened 2016-07-11
Ensure the autostart blocking file is cleaned up

nkrinner:
The systemd corosync service did not delete the
/var/spool/corosync/block_automatic_start when shut down properly,
so restarting corosync was broken. The involved service files now
contain the correct ordering.

See also https://bugzilla.suse.com/show_bug.cgi?id=983617

### DIFF
--- a/chef/cookbooks/corosync/recipes/service.rb
+++ b/chef/cookbooks/corosync/recipes/service.rb
@@ -144,6 +144,16 @@ if node[:corosync][:require_clean_for_autostart]
       )
     end
 
+    template "/etc/systemd/system/#{node[:corosync][:platform][:service_name]}.service" do
+      source "corosync.service.erb"
+      owner "root"
+      group "root"
+      mode "0644"
+      variables(
+        service_name: corosync_shutdown
+      )
+    end
+
     bash "reload systemd after #{corosync_shutdown} update" do
       code "systemctl daemon-reload"
       action :nothing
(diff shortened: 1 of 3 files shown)


aplanas:
@nkrinner did you checked this? https://github.com/crowbar/crowbar-ha/pull/124/files

chef/cookbooks/corosync/recipes/service.rb
```diff
@@ -144,6 +144,16 @@
       )
     end
 
+    template "/etc/systemd/system/#{node[:corosync][:platform][:service_name]}.service" do
+      source "corosync.service.erb"
+      owner "root"
+      group "root"
+      mode "0644"
+      variables(
+        service_name: corosync_shutdown
```
nicolasbock: Do we anticipate that the `service_name` might change in the future, and need interpolation here?

vuntz:
@nkrinner this is very surprising, as I thought it's working. Can't we simply change the dependency to be on the pacemaker.service file instead of corosync? I dislike the fact that we override the corosync service...

(also, as @aplanas pointed out, I have some work which might help -- if that's the case, we might want to simply backport it)

nkrinner:
closed in favor of a better solution that @Itxaka is working on.

aspiers:
@nkrinner A link to the better solution (or at least some info on it) would be helpful here.

aspiers:
@nkrinner @Itxaka I guess the better solution is crowbar/crowbar-core#563?

PR #129, opened 2016-07-13
Enable role deployment for SLES12SP2 only

jsuchome:
(Cannot work until nodes are switched to SP2)


vuntz:
Any reason for this? I think we always allowed any SUSE version there.

jsuchome:
Huh. I guess it was there to ensure we're not having SP1 nodes. But we have other ways to do it already, I assume.

PR #130, opened 2016-07-14
Updated haproxy cfg for version 1.6

abel-navarro:
The 'listen' command in haproxy configuration file no longer supports
setting the listening address right away. Instead it has to be put in a
'bind' command.

This change should be backwards compatible with previous haproxy
versions.


jsuchome:
Do you have link to haproxy changelog?

abel-navarro:
http://www.haproxy.org/download/1.6/src/CHANGELOG

We are using haproxy 1.6.5. I think we were using 1.5.14 before.

jsuchome:
Well, I do not see if exactly this change was mentioned in the changelog, but the change seems correct and I believe you've tested it. 

+1

dguitarbite:
+1 Makes sense to me. Quick look at change log does not provide this information right away.

dguitarbite:
+2

We should merge this patch. I tested it with SP2 HA and it fixes the HAProxy issue. Thanks abel!

PR #131, opened 2016-07-15
Adapt to changes in drbd-overview

jsuchome:
drbd-overview has changed in SP2 and it will not give us the previous "Unconfigured" status.

Now we need another way to find out when we want to execute `drbdadm -- --force create-md` for initialization.

### DIFF
--- a/chef/cookbooks/drbd/libraries/overview.rb
+++ b/chef/cookbooks/drbd/libraries/overview.rb
@@ -18,7 +18,7 @@ require "chef/shell_out"
 
 module DrbdOverview
   def self.get(resource)
-    cmd = "drbd-overview"
+    cmd = "drbd-overview --color=no"
     output = Chef::ShellOut.new(cmd).run_command.stdout
 
     resource_output = ""
@@ -34,26 +34,32 @@ module DrbdOverview
     if resource_output.empty?
       nil
     else
-      # Example:
-      # ["0:postgresql/0", "Connected", "Primary/Secondary", "UpToDate/Diskless", "C", "r----- /var/lib/pgsql xfs 4.0G 67M 4.0G 2%"]
-      parsed = resource_output.split(" ", 6)
+      # Examples:
+      # "0:postgresql/0  Connected(2*) Primar/Second UpToDa/Incons"
+      # "0:postgresql/0  Connected(2*) Primar/Second UpToDa/UpToDa /var/lib/pgsql    xfs 5.0G 94M 4.9G 2%"
+      parsed = resource_output.split " "
       item_1, item_2 = parsed[2].split("/", 2)
       state_1, state_2 = parsed[3].split("/", 2)
+
+      # Example: "0:postgresql/0  Connected(2*) Primar/Second UpToDate(2*)"
+      state_2 = state_1 if state_2.nil? && state_1.include?("2*")
+      # Example: "1:rabbitmq/0    Connected(2*) Secondary(2*) Incons/Incons"
+      item_2 = item_1 if item_2.nil? && item_1.include?("2*")
 
       overview = {}
       overview["state"] = parsed[1]
 
       overview["primary"] = nil
-      if item_1 == "Primary"
+      if item_1.include? "Primar"
         overview["primary"] = state_1
-      elsif item_2 == "Primary"
+      elsif item_2.include? "Primar"
         overview["primary"] = state_2
       end
 
       overview["secondary"] = nil
-      if item_1 == "Secondary"
+      if item_1.include? "Second"
         overview["secondary"] = state_1
-      elsif item_2 == "Secondary"
+      elsif item_2.include? "Second"
         overview["secondary"] = state_2
       end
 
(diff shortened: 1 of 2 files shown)


jsuchome:
What about this:

```ruby
t = template "/etc/drbd.d/#{name}.res" do
 ... # same as above
end
t.run_action(:create)

p = execute "drbdadm -- --force create-md #{name}" do
   only_if { t..updated_by_last_action? }
   action :nothing
end
p.run_action(:run)

```

chef/cookbooks/drbd/providers/resource.rb
```diff
@@ -31,7 +31,7 @@
   raise "Remote node #{remote_host} not found!" if remote_nodes.empty?
   remote = remote_nodes.first
 
-  template "/etc/drbd.d/#{name}.res" do
+  t = template "/etc/drbd.d/#{name}.res" do
```
vuntz: Can we name it `drbd_resource_template` (or something a bit shorter if you prefer, as long as it's meaningful)?

chef/cookbooks/drbd/libraries/overview.rb
```diff
@@ -35,25 +35,30 @@ def self.get(resource)
       nil
     else
       # Example:
-      # ["0:postgresql/0", "Connected", "Primary/Secondary", "UpToDate/Diskless", "C", "r----- /var/lib/pgsql xfs 4.0G 67M 4.0G 2%"]
+      # "0:postgresql/0  Connected(2*) Primar/Second UpToDa/Incons"
       parsed = resource_output.split(" ", 6)
```
vuntz: If your example string is correct, then s/6/4/

chef/cookbooks/drbd/libraries/overview.rb
```diff
@@ -35,25 +35,30 @@ def self.get(resource)
...
       item_1, item_2 = parsed[2].split("/", 2)
       state_1, state_2 = parsed[3].split("/", 2)
 
+      # Example: "0:postgresql/0  Connected(2*) Primar/Second UpToDate(2*)"
+      state_2 = state_1 if state_2.nil? && state_1.include?("2")
```
vuntz: What does this "2*" thing means? Should we check for "(2*)" instead of just "2"?

vuntz: I'm also mixed on keeping the `(2*)` as part of the string, as this is what is causing all the `include?` changes later on, but these changes are a bit confusing, tbh.

chef/cookbooks/drbd/libraries/overview.rb
vuntz: Ah, there's actually more output...

jsuchome: It doesn't have any real effect, we're looking at [2] and [3] items anyway

chef/cookbooks/drbd/libraries/overview.rb
jsuchome: > Should we check for (2*) instead of just 2?

Would it make any difference? It looks like the number cannot be there with other meanings...

vuntz:
Maybe it'd make sense to indeed parse the output of `drbdadm status`. That seems more predictable nowadays.

chef/cookbooks/drbd/libraries/overview.rb
jsuchome: OK,  the example may be more complicated, it may include  that FS/disk info as in previous one, and split(" ", 4) would merge the rest into the last item. So I assume I need to adapt the example once more.

chef/cookbooks/drbd/libraries/overview.rb
vuntz: Well, generally speaking, it's better to be as specific as possible when matching strings to avoid issues later on :-) I'm also unclear if it's always going to be `2` appearing in there.

jsuchome: We only allow 2 nodes for drbd cluster, right?

vuntz: > We only allow 2 nodes for drbd cluster, right?

Now that we have drbd9, we might actually want to revisit this :-)

jsuchome: OK, so let's adapt/rewrite the overview check once again if we agree to include more nodes

abel-navarro:
@vuntz `drbdadm dstate {all|RESOURCE...}` does not give information about `primary|secondary` which we use in [while creating resources](https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/drbd/providers/resource.rb#L75).

vuntz:
> @vuntz drbdadm dstate {all|RESOURCE...} does not give information about primary|secondary which we use in while creating resources.

s/dstate/status/ :-)

abel-navarro:
`drbdadm status` command was not available on drbd for SLE12SP1. For SP2 it does indeed make sense.

jsuchome:
So does it actually mean -1 from you? This seems to work, even gating was green...

vuntz:
For the record, I'm fine with the change, but I think the code will have to be reworked later on.

dguitarbite:
+1

abel-navarro:
+1

PR #132, opened 2016-07-15
Add lgtm.co config

rsalevsky:


MaximilianMeister:
+1

.lgtm
```diff
@@ -0,0 +1,3 @@
+approvals = 2
+pattern = "(?i):\\+1:|\\+1"
```
nicolasbock: Also include `(?i)lgtm`?

rsalevsky: Nop, we decided that LGTM does not count as +1. It's more like looks ok but I have no real idea what happens in detail to decide.

nicolasbock: Ok, sounds good, just wanted to make since I had seen this before in comments.

+1

PR #133, opened 2016-07-20
Create base of Cloud 7 Upgrade API V2 for crowbar-ha

MaximilianMeister:


rsalevsky:
+1

rsalevsky:
+1

crowbar_framework/app/controllers/api/v2/clusters_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Api::V2::ClustersController < ApplicationController
+  api :GET, "/api/v2/clusters", "List all HA clusters"
```
aspiers: Why isn't this `/api/v2.0/clusters`?  And the same for all the other routes?  Surely it should be consistent with the `api_version` value?

rsalevsky: The reason behind this was that our API will most likely not have minor versions. But we are even unsure with this so feedback is welcome.

aspiers: @rsalevsky If it will not have minor versions then why doesn't it say `api_version "2"` instead of `"2.0"`?  Again, surely the version should be consistent between the route and the `api_version` value?

Having said that, whilst I certainly don't claim to be an expert in REST API design, a quick google reveals that the industry seems to have reached a reasonable consensus that thinking only about API version numbers is not the best way to design a future-proof API:

- https://www.narwhl.com/2015/03/the-ultimate-solution-to-versioning-rest-apis-content-negotiation/
- http://www.ben-morris.com/rest-apis-dont-need-a-versioning-strategy-they-need-a-change-strategy/
- https://github.com/restfulapi/api-strategy#use-api-versioning
- https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/
- http://stackoverflow.com/questions/389169/best-practices-for-api-versioning

So I would recommend reading these and considering a solution which supports versioning via the `Accept` header. (It could still support embedding the version in the route too, however.)

MaximilianMeister:
> @rsalevsky If it will not have minor versions then why doesn't it say api_version "2" instead of "2.0"? Again, surely the version should be consistent between the route and the api_version value?

I can adapt the api_version value as well, it is used for the API documentation.

> 
> Having said that, whilst I certainly don't claim to be an expert in REST API design, a quick google reveals that the industry seems to have reached a reasonable consensus that thinking only about API version numbers is not the best way to design a future-proof API:
> 
> https://www.narwhl.com/2015/03/the-ultimate-solution-to-versioning-rest-apis-content-negotiation/

I would just like to quote one specific statement here, it says:

>> "Rather than versioning the entire API when a single resource has a breaking change, you can version directly at the resource level. This allows teams responsible for only a handful of resources to operate independently and enforces the constraint that a resource is solely responsible for its state."

which sounds exactly like what we are trying to do here. see the backups API.
all the other v2 endpoints are basically new in terms of resources because in our current API you can see that almost nothing is resourceful.

> http://www.ben-morris.com/rest-apis-dont-need-a-versioning-strategy-they-need-a-change-strategy/
> https://github.com/restfulapi/api-strategy#use-api-versioning
> https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/
> http://stackoverflow.com/questions/389169/best-practices-for-api-versioning

Thanks for digging all this out. I don't know the people that wrote this, but I trust you there that they represent the industry. At least they seem to know much more about REST API's than we do.

> So I would recommend reading these and considering a solution which supports versioning via the Accept header. (It could still support embedding the version in the route too, however.)

I suppose this mostly relates to the Upgrade "API" that we have right now.
reusing the existing upgrade API endpoints with a version 2 in the header is not an option IMHO as we don't need all of the old endpoints, and the new ones we have may not fit in one of the old.

I think overall it is better to implement something new now, something that fits the resources that we have. we haven't considered a proper (REST) API implementation at any point in the past, so why trying to adapt/reuse the non restful API that we are stuck with now? I personally think the API endpoints we have right now are horrible and unintuitive.

Unfortunately we don't have a REST API specialist here in our dev team, that would really help I guess, and so we try to do everything we can to make it good in the time that we have. But I guess however we will do it...

>> "Given the contentious nature of REST, you’ll always be wrong in somebody’s eyes no matter what approach you take."

We definitely need more input/reviewers here. thanks @aspiers for the review.

MaximilianMeister:
@aspiers In the beginning, technically we had sth. like https://www.bignerdranch.com/blog/adding-versions-rails-api/ but we dropped it again, not sure why. would you be okay with an approach like this?

MaximilianMeister:
stripped the `/v2` from the url. the version will be sent now through the accept header

rsalevsky:
+1

toabctl:
@rsalevsky can you please stop merging things where we have a) only 1x +1 and b) a failing gate?

rsalevsky:
@toabctl Ah sorry, I missed the +1 I mixed this with the crowbar-openstack PR. But the code it not used atm so the gating should be anyway not relevant.

PR #135, opened 2016-07-29
Remove HA roles on node cluster leave

Itxaka:
- [X] Remove cluster roles
- [x] Remove parent roles if all child roles are in a cluster
- [x] Remove founder mark?
- [x] Decide a new founder if we are removing the founder?

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -393,10 +393,69 @@ class PacemakerService < ServiceObject
                                remote_nodes, member_nodes, remotes, members)
 
     role.save
-
+    remove_cluster_roles_from_removed_nodes(old_role, member_nodes)
     apply_cluster_roles_to_new_nodes(role, member_nodes, remote_nodes)
 
     @logger.debug("Pacemaker apply_role_pre_chef_call: leaving")
+  end
+
+  def remove_cluster_roles_from_removed_nodes(old_role, member_nodes)
+    @logger.debug("Pacemaker remove_cluster_roles_from_removed_nodes: entering")
+    if old_role.nil? || old_role.empty?
+      @logger.debug("Pacemaker remove_cluster_roles_from_removed_nodes: leaving")
+      return
+    end
+    new_member_list = member_nodes.map(&:name)
+    old_member_list = old_role.override_attributes[@bc_name]["elements"]["pacemaker-cluster-member"]
+    member_difference = old_member_list - new_member_list
+    # return if no membership change
+    return if member_difference.empty?
+
+    @logger.debug("Found leaving nodes: #{member_difference}")
+    member_difference.each do |n|
+      changed = false
+      node = NodeObject.find_node_by_name(n)
+      if node.nil?
+        @logger.warn("Node #{n} disappeared while trying to remove HA roles.")
+        next
+      end
+      cluster_name = old_role.proposal.name
+      node.roles.each do |role|
+        # only pick barclamp-config-proposal
+        next unless role =~ /(.*)-config-(.*)/
+        # pick the first part of the role which is the name of the barclamp
+        # to check for that ha attribute
+        barclamp = role.split("-")[0]
+        next unless node[barclamp]
+        next unless node[barclamp]["ha"]
+        # using to_s because if not it will appear as Attribute object
+        next unless node[barclamp]["ha"]["enabled"].to_s == "true"
+        # check the rol elements to see if they are applied to a cluster
+        role = RoleObject.find_role_by_name(role)
+        role.override_attributes[barclamp]["elements"].each do |subrole, members|
+          members.each do |member|
+            # only remove the roles that are applied to our cluster
+            next unless cluster_name(member) == cluster_name
+            @logger.debug("Removing role role[#{subrole}] from #{node.role.name}")
+            node.delete_from_run_list("role[#{subrole}]")
+            changed = true
+          end
+        end
+        # remove the barclamp-config-proposal if all its elements are part of the cluster
+        remove_father_role = role.override_attributes[barclamp]["elements"].all? do |r, members|
+          # modify the list inplace to have the proper cluster name
+          members.map! { |m| cluster_name(m) }
+          members.include?(cluster_name) || members.empty?
+        end
+        if remove_father_role
+          @logger.debug("Removing role role[#{role.name}] from #{node.role.name}")
+          node.delete_from_run_list("role[#{role.name}]")
+        end
+      end
+      node[:pacemaker][:founder] = false
+      node.role.save if changed
+    end
+    @logger.debug("Pacemaker remove_cluster_roles_from_removed_nodes: leaving")
   end
 
   def preserve_existing_password(role, old_role)


Itxaka:
Not liking the way we need to iterate over the roles and split them to get to their attributes, thus knowing if those roles are on ha or not. There is probably a simpler approach but I could not find it.

Maybe the roles should also have the ['ha']['enabled'] attributes on the override_attributes so its easier to get the roles, check for that key and remove them if true?

Itxaka:
Dont think the gating failure had to do with this patch.

aspiers:
IIUC, this gating failure should be fixed by https://github.com/crowbar/crowbar-openstack/pull/478

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,44 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+    new_member_list = member_nodes.map(&:name)
+    old_member_list = old_role.override_attributes[@bc_name]["elements"]["pacemaker-cluster-member"]
+    member_difference = old_member_list - new_member_list
+    unless member_difference.empty?
+      @logger.debug("Found leaving nodes: #{member_difference}")
+      member_difference.each do |n|
+        changed = false
+        node = NodeObject.find_node_by_name(n)
+        node.roles.each do |rol|
+          # pick the first part of the role which is the name of the barclamp
```
aspiers: I think it would be a tiny fraction safer to do it more like this:

    next unless rol =~ /(.+)-config-(.+)/
    bc, proposal = $1, $2

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,44 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+          # to check for that ha attribute
+          attribute = rol.split("-")[0]
```
aspiers: As in my above comment, `bc` or `barclamp` would be a slightly nicer name for this variable than `attribute`.

aspiers:
Other than my nitpicks, this is looking good to me!

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: While this would work, why not match the `barclamp-server` role instead? is there any reason to match the `barclamp-config-default` one?

Itxaka: Ill answer myself while I was testing. Seems that for barclamps with several roles (cinder for example) there is no `cinder-server` but there is a `cinder-config-default``role!

Itxaka: Umm, but an issue with this is that matching against `/(.+)-config-(.+)/` will only remove those roles, and we also want to remove the other roles as well. In the case of `cinder` we want to remove:

`cinder-config-default`
`cinder-controller`
`cinder-volume`

If we match only against cinder-config-default, the other roles would still be left there.

aspiers: Crowbar creates three different types of Chef roles; please see slide 65 of [the HA training deck](https://w3.suse.de/~aspiers/cloud/HA-training/HA-training-all-slides.pdf) as this is a crucial point to understand about the internals of crowbar-core.  You are right that we need to remove the proposal roles (e.g. `cinder-config-proposalname`) and also the standard Chef roles (e.g. `cinder-controller`), however neither the current code nor my proposed change will remove the latter, because the latter does not have the `ha.enabled` attribute set.

I think the right way to remove both can be found in the `#apply_cluster_roles_to_new_nodes_for` method.  In particular look how [it iterates over `cluster_role.elements`](https://github.com/crowbar/crowbar-ha/blob/master/crowbar_framework/app/models/pacemaker_service.rb#L140).

You can also see this by running this command on the admin node:

```
root@crowbar:~ # knife role show cinder-config-default -a override_attributes.cinder.elements
override_attributes.cinder.elements: 
  cinder-controller:  ["cluster:services"]
  cinder-volume:     
      d52-54-77-77-77-01.c17.cloud.suse.de
      d52-54-77-77-77-02.c17.cloud.suse.de
```

So the `cinder.elements` attribute in the cinder proposal called `default` contains two elements: `cinder-controller` and `cinder-volume`.  You will need to remove these by first looking them up from the proposal role.

aspiers:
Sorry for my bad review before, but hopefully now I finally realised and explained what needs to be done. Let me know if you have questions.

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: I see, but in my code Im checking against the **node** attributes to see if ha is enabled for that barclamp and removing all roles from there:
`next unless node[barclamp]["ha"]["enabled"].to_s == "true"`

So with this code, all cinder roles will be removed as they will all match cinder-* names and the check against HA for that barclamp is done against the node itself instead of against the attributes in the role.

What this could lead to is to removing roles that are not in HA in case one of the roles was deployed to a cluster and another not (i.e. cinder-volume is deployed in the cluster but cinder-controller isn't) which, tbh I find a very strange deployment :)

Also, in the case of any future barclamps not having their roles named `barclamp-role` it could also lead to failure, but I dont think that is something to be worried about seeing the names for all the actual roles, as they follow that nomenclature.


I though about iteration over roles based on the role node that match the names, but can we guarantee that all the roles will have those attributes_override? Looks to me that only the `barclamp-config-proposal` is the only one that has the attribute `ha.enabled` but the other roles do not, which will cause the same issues (not knowing if one role is on ha and the other isnt)


Sorry for the long post, trying to understand this correctly. Appreciate the comments and explanation though! :smile:

aspiers: > I see, but in my code Im checking against the **node** attributes to see if ha is enabled for that barclamp and removing all roles from there:
> `next unless node[barclamp]["ha"]["enabled"].to_s == "true"`

Ahah, I see! Sorry, I missed that. Yes, I was thinking you were checking the role attributes.

> So with this code, all cinder roles will be removed as they will all match cinder-* names and the check against HA for that barclamp is done against the node itself instead of against the attributes in the role.
> 
> What this could lead to is to removing roles that are not in HA in case one of the roles was deployed to a cluster and another not (i.e. cinder-volume is deployed in the cluster but cinder-controller isn't)

Exactly, and that would be wrong - we must only remove the roles which were assigned by expanding the clustered elements.

> which, tbh I find a very strange deployment :)

It makes sense because we don't support cinder-volume HA yet.  In fact that's [in the current sprint](https://trello.com/c/QkpcjlzL).

> Also, in the case of any future barclamps not having their roles named `barclamp-role` it could also lead to failure, but I dont think that is something to be worried about seeing the names for all the actual roles, as they follow that nomenclature.

There is a rule for [the naming of barclamp proposal roles](https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/proposal.rb#L180), but I'm not sure if there is also one for naming of element roles. I think that's more of a convention, and the correct way to iterate over them is via the deployment element attributes as mentioned in my previous comment and below.

> I though about iteration over roles based on the role node that match the names, but can we guarantee that all the roles will have those attributes_override?

No we can't, for the reason I explained in my last comment.

> Looks to me that only the `barclamp-config-proposal` is the only one that has the attribute `ha.enabled` but the other roles do not, which will cause the same issues (not knowing if one role is on ha and the other isnt)

Correct.  That's what I was saying in my last comment - it's crucial to understand the 3 types of Chef roles in play.  And that's why we need to iterate over `cluster_role.elements` from `*-config-*`.  In the example I gave, the `knife` command clearly shows that `cinder-controller` is assigned to the cluster but `cinder-volume` isn't.

> Sorry for the long post, trying to understand this correctly. Appreciate the comments and explanation though! :smile:

No apology required - it's really impressive how quick you are picking this stuff up :)

Itxaka:
@aspiers Thanks for the commments, I updated to take into account the elements and check against their membership to find out if they are part of the cluster.

But....

In the case of cinder and the 3 roles in there:

cinder-config-default -> attribute cinder.ha.enabled = true
cinder-controller -> applied to cluster
cinder-volume -> applied to a single node

We should remove the clustered role`cinder-controller` but we should not remove `cinder-config-default` as it contains attributes used by the `cinder-volume` one but... having a the volume service without the controller makes no sense.

Seems to me that trying to remove the cluster roles and leave the non-clustered ones can lead to a very unusual screw up for crowbar with unexpected consequences.

I would say that the correct action here would be, if ha.enabled is found in the `barclamp-config-default`, we should wipe all the roles contained there to avoid unexpected issues. Make a note of it in the documentation or even in the UI when removing a node from a cluster and move on.

Seems to me that this feature (and the parent feature) requires deeper discussion and some guidelines on how to tackle it properly and the downfall of the different options as it doesnt seem very clear to me whats the correct action to take.

thoughts?

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+        barclamp = rol.split("-")[0]
+        next unless node[barclamp]
+        next unless node[barclamp]["ha"]
+        # using to_s because if not it will appear as Attribute object
+        next unless node[barclamp]["ha"]["enabled"].to_s == "true"
+        # check the rol elements to see if they are applied to a cluster
+        role.override_attributes[barclamp]['elements'].each do |subrole, members|
+          members.each do |value|
+            # only remove the roles that are applied to our cluster
+            next unless value =~ /cluster:#{cluster_name}/
```
aspiers: Here you should use `#is_cluster?`

aspiers:
> In the case of cinder and the 3 roles in there:
> 
> cinder-config-default -> attribute cinder.ha.enabled = true
> cinder-controller -> applied to cluster
> cinder-volume -> applied to a single node

Correct.

> We should remove the clustered role `cinder-controller`

Yes, but only from the current node of course.

> but we should not remove `cinder-config-default` as it contains attributes used by the `cinder-volume` one

The reason for not removing it is stronger than that: it contains all the data for the cinder barclamp's `default` proposal, and we are not removing the proposal (which applies to multiple nodes).

> but... having a the volume service without the controller makes no sense.

That would not happen.  Removing `cinder-controller` from a node which is no longer in the cluster does not mean it is removed from all the other nodes.

> Seems to me that trying to remove the cluster roles and leave the non-clustered ones can lead to a very unusual screw up for crowbar with unexpected consequences.

That's not what we are doing.  We're only removing clustered roles from nodes which have been removed from the cluster, not from *all* nodes in the cluster.

> I would say that the correct action here would be, if ha.enabled is found in the `barclamp-config-default`, we should wipe all the roles contained there to avoid unexpected issues. Make a note of it in the documentation or even in the UI when removing a node from a cluster and move on.

Absolutely not!  We are only dealing with the case of removing a node from the cluster.  In that case the remaining nodes should still function as before.

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: But I want to match the full thing.
The `is_cluster? element` method only matches the start:
`element.start_with? "#{cluster_key}:"` where `cluster_key` seems to be always `"cluster"`

Itxaka:
> The reason for not removing it is stronger than that: it contains all the data for the cinder barclamp's default proposal, and we are not removing the proposal (which applies to multiple nodes).

Thank you! I knew I was missing something important here.

> That would not happen. Removing cinder-controller from a node which is no longer in the cluster does not mean it is removed from all the other nodes.

> That's not what we are doing. We're only removing clustered roles from nodes which have been removed from the cluster, not from all nodes in the cluster.

> Absolutely not! We are only dealing with the case of removing a node from the cluster. In that case the remaining nodes should still function as before.

Yes, you are right, I messed up the concepts there for a second. :+1:

aspiers:
No problem, at least you have an excuse being new - I am supposed to know all this well and it still took me 3 attempts to get it clear in my head ;-/

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Oh yeah sorry - in that case it should be

    members.each do |member|
      next unless cluster_name(member) == cluster_name

(`member` is a less vague variable name than `value`, so I think it's worth changing.)

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+        next unless rol =~ /(.*)-config-(.*)/
+        # pick the first part of the role which is the name of the barclamp
+        # to check for that ha attribute
+        barclamp = rol.split("-")[0]
+        next unless node[barclamp]
+        next unless node[barclamp]["ha"]
+        # using to_s because if not it will appear as Attribute object
+        next unless node[barclamp]["ha"]["enabled"].to_s == "true"
+        # check the rol elements to see if they are applied to a cluster
+        role.override_attributes[barclamp]['elements'].each do |subrole, members|
```
aspiers: Hound is right here - should be `"elements"` for consistency.

aspiers:
Thanks - if you fix the two remaining nitpicks then this gets my vote.

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: done!

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: changed the variable but not sure where did you get the `cluster_name(member)` from?

I changed it to be a direct comparision, please let me know if that is wrong :)

Itxaka: oooh, I see it comes from PacemakerServiceObject <- ServiceObject ?

Itxaka: Updated!

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+
+    @logger.debug("Found leaving nodes: #{member_difference}")
+    member_difference.each do |n|
+      changed = false
+      node = NodeObject.find_node_by_name(n)
+      # get the cluster name for the node pacemaker-config-NAME
+      cluster_name = node["pacemaker"]["config"]["environment"].split("-")[2]
+      node.roles.each do |rol|
+        # only pick barclamp-config-proposal
+        next unless rol =~ /(.*)-config-(.*)/
```
vuntz: Why do we do this? Don't we need to remove roles like nova-controller?

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+    new_member_list = member_nodes.map(&:name)
+    old_member_list = old_role.override_attributes[@bc_name]["elements"]["pacemaker-cluster-member"]
+    member_difference = old_member_list - new_member_list
+    # return if no membership change
+    return if member_difference.empty?
+
+    @logger.debug("Found leaving nodes: #{member_difference}")
+    member_difference.each do |n|
+      changed = false
+      node = NodeObject.find_node_by_name(n)
```
vuntz: Probably worth a `next if node.nil?`

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+      node.roles.each do |rol|
+        # only pick barclamp-config-proposal
+        next unless rol =~ /(.*)-config-(.*)/
+        # pick the first part of the role which is the name of the barclamp
+        # to check for that ha attribute
+        barclamp = rol.split("-")[0]
+        next unless node[barclamp]
+        next unless node[barclamp]["ha"]
+        # using to_s because if not it will appear as Attribute object
+        next unless node[barclamp]["ha"]["enabled"].to_s == "true"
```
vuntz: This code is a bit too fragile, as the "ha enabled"' attribute is not always like this (take neutron, for instance).

The way I would have done it is to iterate through all barclamps, and see which applied proposals for the barclamps have a role for this cluster, and then I would have removed these roles. Now if we simply iterate over the roles of the removed nodes, the challenge is indeed to identify if the role is applied to a cluster or not, and the only real way to do this, afaik, is to look at the proposal itself.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+        # check the rol elements to see if they are applied to a cluster
+        role.override_attributes[barclamp]["elements"].each do |subrole, members|
```
vuntz: I think `role` is not defined.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+    member_difference = old_member_list - new_member_list
+    # return if no membership change
+    return if member_difference.empty?
+
+    @logger.debug("Found leaving nodes: #{member_difference}")
+    member_difference.each do |n|
+      changed = false
+      node = NodeObject.find_node_by_name(n)
+      # get the cluster name for the node pacemaker-config-NAME
+      cluster_name = node["pacemaker"]["config"]["environment"].split("-")[2]
```
vuntz: Don't we have a method that does this?

vuntz:
Wondering out loud? Should we reset the founder attribute on removed nodes, just to be on the safe side?

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Yes, `#cluster_name` which I already mentioned in [one of the outdated diffs](https://github.com/crowbar/crowbar-ha/pull/135#discussion_r74463715)...

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+      node.roles.each do |rol|
```
aspiers: Why `rol` not `role`?

aplanas: Spanish : ))

crowbar_framework/app/models/pacemaker_service.rb
aspiers: > the only real way to do this, afaik, is to look at the proposal itself.

Isn't that what the code immediately below is doing?  It checks the `elements` array for roles which are applied to this cluster.

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Yeah, I guess that is supposed to be `rol`, which makes me wonder whether this latest version of the PR was actually tested.

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Yes, that happens below inside the loop.

vuntz: I'm confused: how can it happen inside the loop with this `next`? It sounds to me we only remove "foo-config-default" roles.

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Maybe it's the intent, but I don't see where we load a proposal in this code...

crowbar_framework/app/models/pacemaker_service.rb
vuntz: No, I think `rol` is different as it's one role on a node, while here we want the role of a proposal.

aspiers: But `rol` has already been forced to be a proposal role...

crowbar_framework/app/models/pacemaker_service.rb
aspiers: What do you mean by "load a proposal"?  The below code is removing `role[#{subrole}]` from the run list, where `subrole` is a role applied to this cluster.

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Oh, hahah.  Please stick to English ;-)

crowbar_framework/app/models/pacemaker_service.rb
aspiers: If `node` is `nil` here, then hasn't something gone wrong?  Or you are saying that nodes could vanish even before the roles are removed, e.g. via Forget?

vuntz: Yes, was thinking about `forget`. And I'm fine with having a log there too.

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Got it after re-reading the patch.

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Yeah, got it now.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+        # check the rol elements to see if they are applied to a cluster
+        role.override_attributes[barclamp]["elements"].each do |subrole, members|
+          members.each do |member|
+            # only remove the roles that are applied to our cluster
+            next unless cluster_name(member) == cluster_name
+            @logger.debug("Removing role role[#{subrole}] from #{node.role.name}")
+            node.role.run_list.delete("role[#{subrole}]")
+            changed = true
+          end
+        end
```
vuntz: I think we should remove the `foo-config-bar` role from the node too in case the node is not used anymore in the proposal.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,56 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+        barclamp = rol.split("-")[0]
+        next unless node[barclamp]
+        next unless node[barclamp]["ha"]
+        # using to_s because if not it will appear as Attribute object
+        next unless node[barclamp]["ha"]["enabled"].to_s == "true"
+        # check the rol elements to see if they are applied to a cluster
+        role.override_attributes[barclamp]["elements"].each do |subrole, members|
+          members.each do |member|
+            # only remove the roles that are applied to our cluster
+            next unless cluster_name(member) == cluster_name
```
vuntz: Also add a `next unless members.include?(node.name)`, I guess.

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Right, but as I think you're implying here, it's possible that the proposal might still be applying a non-HA role to the node.  So we need to check if that's the case.

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Why?  If the cluster is assigned to the role, the individual node names will not be inside the `elements` `Hash`; they'll instead be in `elements_expanded`.

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Yes, that was implied ("in case...") :-)

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Because it's an easy one line test that can help avoid weird bugs in case people did some bad tricks :-)

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: I think we only pick the `barclamp-default-proposal` in here, so we can iterate over its elements as that role is kind of the father role for any `barclamp-role` inside it, instead of having to iterate over all roles.

```
  barclamp-config-default
   +
   |
   +----->Elements
             +
             |
             +---->barclamp-controller
             |
             +---->barclamp-server
```

Althougth, now Im not sure why I went that way. I think the first iteration _did_ went over all roles, but still, this will allows us to remove the `barclamp-default-proposal` at the end of the run as you commented below:

> I think we should remove the foo-config-bar role from the node too in case the node is not used anymore in the proposal.

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: wooops :smile:

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: done!

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: done!

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: done

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: done

Itxaka:
Not tested and missing some items:
 - I think we should remove the foo-config-bar role from the node too in case the node is not used anymore in the proposal.
 - Wondering out loud? Should we reset the founder attribute on removed nodes, just to be on the safe side?

Deploying an HA cloud to test this further.

crowbar_framework/app/models/pacemaker_service.rb
aspiers: I thought so, although "in case" is not the right preposition to use here, because "we should do A in case B" means "we should *always* do A, because we need to protect against the risk of B, even if B might not be true".  Whereas in this case we should remove `foo-config-bar` from the node *if and only if* the node is not used any more in the proposal.  So "if and only if" (or "iff" for short) would have been the right preposition.  But I freely admit your English is *way* better than my French ;-)

aspiers: FWIW, another correct way of wording it would be: we should remove the `foo-config-bar` role from the node too *only in the case where* the node is not used anymore in the proposal.

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: Actually Vincent is rigth on this one. Im sure this is a leftover from before where I iterated over all the roles and created a RoleObject for each, as rigth now the `rol` variable points to a string, which has no override_attributes.

Working on fixing this strange mix that I ended up with :/

aspiers:
> Should we reset the founder attribute on removed nodes, just to be on the safe side?

This raises the question: what happens if the founder is removed from the cluster? In that case we need to pick a new "founder".

Itxaka:
Will now remove the `barclamp-config-proposal` role in case that all elements dependent of that  role are part of the cluster.

Tested locally with:

database-config-default
 - database-server (On cluster)

cinder-config-default
 - cinder-controller (On cluster)
 - cinder-volume ( NOT on cluster)

In this example, both `database-config-default` and `database-server` got removed. 
`cinder-controller` was removed BUT `cinder-volume` and `cinder-config-default` were left alone.


> This raises the question: what happens if the founder is removed from the cluster? In that case we need to pick a new "founder".

Im not even sure where it gets decided who is the founder after looking on the crowbar-pacemaker cookbook :dizzy_face:

vuntz:
> This raises the question: what happens if the founder is removed from the cluster? In that case we need to pick a new "founder".

This already happens automatically in `apply_role_pre_chef_call`.

vuntz:
I just realized: you also need to do the magic for pacemaker-remote nodes.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,71 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+      @logger.debug("Pacemaker remove_cluster_roles_from_removed_nodes: leaving")
+      return
+    end
+    new_member_list = member_nodes.map(&:name)
+    old_member_list = old_role.override_attributes[@bc_name]["elements"]["pacemaker-cluster-member"]
+    member_difference = old_member_list - new_member_list
+    # return if no membership change
+    return if member_difference.empty?
+
+    @logger.debug("Found leaving nodes: #{member_difference}")
```
vuntz: Maybe use `#{member_difference.join(", ")}`?

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,71 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+      node.roles.each do |role|
+        # only pick barclamp-config-proposal
+        next unless role =~ /(.*)-config-(.*)/
+        # pick the first part of the role which is the name of the barclamp
+        # to check for that ha attribute
+        barclamp = role.split("-")[0]
+        next unless node[barclamp]
+        next unless node[barclamp]["ha"]
+        # using to_s because if not it will appear as Attribute object
+        next unless node[barclamp]["ha"]["enabled"].to_s == "true"
```
vuntz: As I wrote in an earlier comment, the check about `node[barclamp]["ha"]["enabled"]` is wrong. And I don't think it's actually needed: things will just work fine if we remove these 4 lines.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,71 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+    @logger.debug("Found leaving nodes: #{member_difference}")
+    member_difference.each do |n|
+      changed = false
+      node = NodeObject.find_node_by_name(n)
+      if node.nil?
+        @logger.warn("Node #{n} disappeared while trying to remove HA roles.")
+        next
+      end
+      # get the cluster name for the node pacemaker-config-NAME
+      cluster_name = node["pacemaker"]["config"]["environment"].split("-")[2]
```
vuntz: Thinking about it more: we should just get the cluster name from the currently applying proposal, which you can get from `old_role`

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,71 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+        # using to_s because if not it will appear as Attribute object
+        next unless node[barclamp]["ha"]["enabled"].to_s == "true"
+        # check the rol elements to see if they are applied to a cluster
+        role = RoleObject.find_role_by_name(role)
+        role.override_attributes[barclamp]["elements"].each do |subrole, members|
+          members.each do |member|
+            # only remove the roles that are applied to our cluster
+            next unless cluster_name(member) == cluster_name
+            @logger.debug("Removing role role[#{subrole}] from #{node.role.name}")
+            node.role.run_list.delete("role[#{subrole}]")
```
vuntz: You should use `node.delete_from_run_list(subrole)` here; it's actually pretty critical :-)

(that's because we use `node.add_to_run_list` when adding roles, see the code path for the opposite case)

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Don't mess up with my frenchified english! ;-)

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,71 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+            changed = true
+          end
+        end
+        # remove the barclamp-config-proposal if all its elements are part of the cluster
+        remove_father_role = role.override_attributes[barclamp]["elements"].all? do |r, members|
+          # modify the list inplace to have the proper cluster name
+          members.map! { |m| cluster_name(m) }
+          members.include?(cluster_name)
```
vuntz: If I understand correctly, we check that all roles have the cluster, to ensure that no roles has the individual node. That sounds wrong to me, as some roles might have none of them.

Since we already loop over all the roles earlier on, why not simply add a `remove_father_role ||= members.include?(node.name)` there (before the `next unless cluster_name(member) == cluster_name`)? Of course you need to initialize `remove_father_role = false` before the loop.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -386,12 +386,71 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+        end
+        if remove_father_role
+          @logger.debug("Removing role role[#{role.name}] from #{node.role.name}")
+          node.role.run_list.delete("role[#{role.name}]")
```
vuntz: Same comment as earlier: use `node.delete_from_run_list`

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: > If I understand correctly, we check that all roles have the cluster, to ensure that no roles has the individual node. That sounds wrong to me, as some roles might have none of them.

Indeed, stuff like pacemaker can have roles that have empty lists. :/ but that should be easy to fix.

> Since we already loop over all the roles earlier on, why not simply add a remove_father_role ||= members.include?(node.name) there (before the next unless cluster_name(member) == cluster_name)? Of course you need to initialize remove_father_role = false before the loop.

Not sure if I want to mess that part by adding more stuff to it, rather have it separated for clarity.

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: Isnt this just faster? If a barclamp doesnt have the ha.enabled flag to just skip over it?

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: done!

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: done!

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: done! old_role.proposal.name

Itxaka:
wat:
00:35:59.665 +(qa_crowbarsetup.sh:106) complain(): printf 'Error: %s\n' 'No suitable node(s) for ceilometer-agent found.'
00:35:59.665 Error: No suitable node(s) for ceilometer-agent found.

Itxaka:
rebasing to see if the issue goes away..

Itxaka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;lF3lFDBD&#x2F;18-3-p1-important-bsc-979903-l3-removing-node-from-cluster-doesn-t-remove-cluster-roles-from-node"> (3) P1 #important bsc#979903 - L3: Removing node from cluster doesn&#39;t remove cluster roles from node</a>

Itxaka:
@aspiers @vuntz @aplanas need an approved review for this to be merged. Thanks!

Itxaka:
rebased!

PR #137, opened 2016-08-22
Rewrite possible existing signatures when creating a volume

jsuchome:


chef/cookbooks/lvm/libraries/provider_lvm_physical_volume.rb
```diff
@@ -52,7 +52,7 @@ def action_create
           Chef::Log.info "Physical volume '#{new_resource.name}' found. Not creating..."
         else
           Chef::Log.info "Creating physical volume '#{new_resource.name}'"
-          cmd = shell_out("pvcreate #{new_resource.name}")
+          cmd = shell_out("pvcreate --yes #{new_resource.name}")
```
rhafer: @jsuchome ```-f``` or probaby ```-Z``` should also do something similar. Do you have and idea which of the three options should be prefered? (I don't)

jsuchome: Hm, `-f` is already used in cinder cookbook, so it might be good to use it here as well

rhafer: Yeah, let's use ```-f``` then. That's also how it got addressed in ansible it seems

rhafer:
+1

toabctl:
@jsuchome can you update the commit msg to mention why we need this?

jsuchome:
@toabctl commit message updated

toabctl:
+1

rhafer:
Still +1

vmoravec:
Does anyone know why pacemaker failed with `"error":"Failed to validate proposal: All nodes in proposal must have the same platform` despite the fact that all nodes were on SP2? We experienced this already once today on qa hw.

rhafer:
@vmoravec that should be fixed with: https://github.com/SUSE-Cloud/automation/pull/1172 I assume

jsuchome:
Ah; gating failed because it assigned cinder-volume to SP1 nodes...

jsuchome:
OK, it did not, correction in https://github.com/SUSE-Cloud/automation/pull/1180

PR #138, opened 2016-08-25
Simple check if HA clusters report some problems

jsuchome:
Requires https://github.com/crowbar/crowbar-core/pull/619


jsuchome:
That `check_if_nodes_are_available` method should also go into some library...

crowbar_framework/app/models/api/cluster.rb
```diff
@@ -16,5 +16,44 @@
 
 module Api
   class Clusters
```
rsalevsky: `class Clusters < Tableless`

crowbar_framework/app/models/api/cluster.rb
```diff
@@ -16,5 +16,44 @@
...
+    # {
+    #     "crm_failures" => {
+    #             "node1" => "reason for crm status failure"
+    #     },
+    #     "failed_actions" => {
+    #             "node2" => "Failed action on this node"
+    #     }
+    # }
+    # User has to manually clean pacemaker resources before proceeding with the upgrade.
+    def health_check
```
rsalevsky: `healthy?(cluster)`

We should run this check per cluster.

rsalevsky:
In general we should design this model like a real rails model also the action in the controller is missing.

crowbar_framework/app/models/api/cluster.rb
jsuchome: Well, it does not return boolean but a hash...

jsuchome:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;sKTTY1Wp&#x2F;65-3-s46p13-implement-api-for-first-step-of-upgrade-checks"> (3) S46P13: Implement API for first step of upgrade: checks</a>

aspiers:
Hmm, I'm not convinced it makes sense to add a health check API specifically for HA, when we already know that [we need a health check API which also covers non-HA issues](https://trello.com/c/r09uuhvj), *and* that this health check of non-HA issues would be useful in the upgrade case.  For example, checking for stale `/var/chef/cache/pause-file.lock` files is probably just as important to live upgrade as checking for cluster health issues.

I guess there must have been some discussion about the addition of an HA cluster API which I missed.  Maybe that's my fault, but it still would have been nice if someone had pinged me about it :-/

rsalevsky:
I hope I can clarify this... Our ideas was to make the API very granular so we will not have a general health API endpoint. But we want to have health endpoints per "module" so crowbar, ha, ceph and openstack can have there own endpoints. This should prevent that we mix stuff again we don't want to have HA stuff in the crowbar(-core) endpoints otherwise we need to check always if we have HA.

aspiers:
@rsalevsky An API endpoint is usually associated with a single HTTP service on a single TCP port. So when you say 'we want to have health endpoints per "module"', do you really mean multiple HTTP services (which makes no sense to me), or do you actually mean health *routes* per module, within the normal Crowbar API endpoint which provides routes for all modules?

I agree that each barclamp should take care of its own health checks, and I'm also fine with exposing the results of these checks individually through routes, so that an API client could ask for any health problems associated with barclamp `foo`.  However:

* the framework should provide a unified facility for health checks across all barclamps, rather than each one having to reinvent the wheel
* health checks from all barclamps still need to be aggregated into a single view in the UI

Therefore I still think that this needs further design and coding.

jsuchome:
@aspiers could we merge it now? This is only one remaining bit to the API that already exists.

We could possibly look at the general health check  code while we do the API cleanup, based on the card you have proposed.

crowbar_framework/app/models/api/cluster.rb
```diff
@@ -15,6 +15,45 @@
...
+      service_object = CrowbarService.new(Rails.logger)
+      service_object.check_if_nodes_are_available founders
+
+      founders.each do |n|
+        ssh_retval = n.run_ssh_cmd("crm status 2>&1")
+        if ssh_retval[:exit_code] != 0
+          crm_failures[n.name] = ssh_retval[:stdout]
+          next
+        end
+        ssh_retval = n.run_ssh_cmd('crm status | grep -A 2 "^Failed Actions:"')
```
aspiers: Hmm, I'm not sure it's a good idea to parse `crm status` output, since AFAIK it's intended to be human-readable not machine-readable. At least it's worth checking with `#sleha` first ...

crowbar_framework/app/controllers/api/clusters_controller.rb
```diff
@@ -31,7 +31,9 @@ def show
   api :GET, "/api/clusters/health", "Health check HA clusters"
```
aspiers: Why not `/api/pacemaker/health`? Then we can establish a pattern whereby every barclamp has the possibility to provide a health status? In fact, along the same lines of reasoning, why do we have *any* routes under `/api/clusters` when they could be under `/api/pacemaker` instead? What is the justification for making a special exception for clusters?

aspiers:
Reviewed - as already stated I'm not convinced with this approach as it stands, because I don't see a good reason to special-case clusters. And it seems extremely easy to fix - just rename the route and controller.

BTW, in future please assign me to PRs if you want me to review them, because for some reason github is not reliably notifying me when someone mentions me in a comment :-(

crowbar_framework/app/models/api/cluster.rb
jsuchome: This is the same check we've done already in Cloud5-6 upgrade.

I'm not saying it's perfect, it is actually a notification that something was wrong and the output should be shown to user.

crowbar_framework/app/controllers/api/clusters_controller.rb
jsuchome: I'm fine with moving it to api/pacemaker_controller.rb and modes/api/pacemaker.rb.

@rsalevsky @MaximilianMeister , any objections?

MaximilianMeister: > I'm fine with moving it to api/pacemaker_controller.rb and modes/api/pacemaker.rb.

I think it's fine, but then we should maybe also rename the storage to ceph. I think it makes sense, but lets wait for some input from @rsalevsky 

> Then we can establish a pattern whereby every barclamp has the possibility to provide a health status?

@aspiers what's the usecase of having a health check for every barclamp?

> What is the justification for making a special exception for clusters?

it's not really a special exception as we named the ceph controller "storages" which is also away from the actual name of the feature.

about finding the right name, we have the Addons (which are also "features" in our repos.yml):
* ha (SLE HAE)
* storage (SES)

we have the barclamps:
* ceph
* pacemaker

we have the abstracted thing that they are:
* storages
* clusters

we also have the resources in rails, where it comes to a database model where we have multiple clusters for instance, thus needing a plural in the names. how would we call that? `pacemakers` or `cephs` ? I think in the design of the naming we favoured the rails side a bit more, to fit it in the framework, but maybe that was not the best decision.

jsuchome: > what's the usecase of having a health check for every barclamp?

See https://trello.com/c/r09uuhvj/1924-crowbar-health-check-api . But really, I'd like to work on this card (that's still to be refined) once we have it in the sprint and first finish this PR, that is part of a card that we have currently in doing. Renaming controller/model from `cluster` to `pacemaker` seems like something we can do as part of this PR, but broader discussion about how the API should be cleaned up belongs IMHO to that later one.

rsalevsky: I'm against the change to ceph or pacemaker because the API is not designed per barclamp is designed per component. So for now the APIs are very generic and should be able to handle more then just one barclamp. The barclamp specific implementation will follow later. Maybe the `clusters` name is not perfect but better then `has`. 

There was a lot of discussion going on some weeks back and we changed a lot but nobody complained about the `cluster` routes. Now it's nearly already too late because we communicated this already with the UI/UX team. Changing this would create discussions and we loose time again which we don't have. So there should be a very good reason for changing this.

aspiers: @rsalevsky What is a "component" exactly? I'm not aware of any clear definition of that, so it seems ambiguous to say things like "the API is [...] designed per component", and consequently it's hard to understand *why* the API should be designed per component. In contrast, barclamps are clearly defined, we all know what they are, and the vast majority of the Crowbar API so far has been designed per barclamp. So if you want us to start to deviate from that then I would really appreciate a clear explanation of *why* you want that.

> There was a lot of discussion going on some weeks back and we changed a lot but nobody complained about the cluster routes.

That's not really true; 19 days ago [I expressed some concerns with this approach and also asked whether I had missed some design discussions, and if so why I hadn't been pinged about them](https://github.com/crowbar/crowbar-ha/pull/138#issuecomment-242747314), which I think was a fair question considering that this is strongly related to HA.

Also I'm not convinced that renaming the controller from `clusters` to `pacemaker` would be highly disruptive to the UI/UX team - it's just a straightforward rename, as @jsuchome said.

aspiers: @MaximilianMeister:
> > I'm fine with moving it to api/pacemaker_controller.rb and modes/api/pacemaker.rb.
> 
> I think it's fine, but then we should maybe also rename the storage to ceph. I think it makes sense, but lets wait for some input from @rsalevsky

Yes, maybe that makes sense too - but see below.

> > Then we can establish a pattern whereby every barclamp has the possibility to provide a health status?
> 
> @aspiers what's the usecase of having a health check for every barclamp?

I explained this in [a comment on this PR 19 days ago](https://github.com/crowbar/crowbar-ha/pull/138#issuecomment-242747314).

> > What is the justification for making a special exception for clusters?
> 
> it's not really a special exception as we named the ceph controller "storages" which is also away from the actual name of the feature.

Well, then we have two special exceptions, but I don't think they automatically mutually justify each other.  There should be good and clear reasons for adding both, since AFAICS they are a fundamental deviation from Crowbar's design.  I'm not saying they're definitely wrong, just that I would like to understand what those reasons are.

> about finding the right name, we have the Addons (which are also "features" in our repos.yml):
> * ha (SLE HAE)
> * storage (SES)
> 
> we have the barclamps:
> * ceph
> * pacemaker
> 
> we have the abstracted thing that they are:
> * storages
> * clusters
> 
> we also have the resources in rails, where it comes to a database model where we have multiple clusters for instance, thus needing a plural in the names. how would we call that? `pacemakers` or `cephs` ? I think in the design of the naming we favoured the rails side a bit more, to fit it in the framework, but maybe that was not the best decision.

What data are you proposing to store in the database for these which isn't already covered by the proposals?

Another key question to ask when considering `pacemaker` vs. `cluster` is: is this code specific to Pacemaker, or generic to any clustering technology?  I.e. if we were to introduce other clustering technology, would this code work with that?  AFAICS, the answer is that it's specific to Pacemaker, which strongly suggests to me that `pacemaker` is the right name.

aspiers:
BTW, this PR is rapidly descending into a design discussion, which from previous experience we have learned is a **VERY BAD IDEA**.  Therefore we should continue this in mumble.

crowbar_framework/app/controllers/api/clusters_controller.rb
MaximilianMeister: > @aspiers what's the usecase of having a health check for every barclamp?
>> I explained this in a comment on this PR 19 days ago.

thanks, but I still don't understand why we need a health check for every barclamp, at least the comment doesn't mention any barclamp specific explanation

> What data are you proposing to store in the database for these which isn't already covered by the proposals?

I think we still don't have a use case for cluster or storage instances (db entries), this was just future groundwork for separating resources. right now we only have namespaced collective methods. but this could change, and right now it's not set in stone anyways.
I think we should mumble about that to clarify and adapt.

crowbar_framework/config/routes.d/cluster.routes
```diff
@@ -1,6 +1,6 @@
 namespace :api,
   constraints: ApiConstraint.new(2.0) do
-  resources :clusters,
+  resources :pacemaker,
```
MaximilianMeister: I would make a single resource from it like in https://github.com/crowbar/crowbar-ceph/pull/71 otherwise i think you have to call the controller pacemakers_controller. we can still refer to single clusters, but not from the database in a pacemakers table, but mapped from the proposals

crowbar_framework/app/controllers/api/pacemaker_controller.rb
```diff
@@ -14,27 +14,29 @@
...
-  api :GET, "/api/clusters", "List all HA clusters"
+class Api::PacemakerController < ApiController
+  api :GET, "/api/pacemaker", "List all pacemaker clusters"
   api_version "2.0"
   def index
     render json: [], status: :not_implemented
   end
 
-  api :GET, "/api/clusters/:id", "Show a single HA cluster"
+  api :GET, "/api/pacemaker/:id", "Show a single pacemaker cluster"
```
MaximilianMeister: shall we drop this for now, or do we want to keep it until we add the mapping of clusters to proposals?

jsuchome: let's drop it when we do not need it

crowbar_framework/config/routes.d/cluster.routes
jsuchome: Done

crowbar_framework/config/routes.d/cluster.routes
```diff
@@ -1,6 +1,6 @@
 namespace :api,
   constraints: ApiConstraint.new(2.0) do
-  resources :clusters,
+  resource :pacemaker, controller: :pacemaker
     only: [:index, :show] do
```
MaximilianMeister: the show can be dropped here as well

crowbar_framework/app/controllers/api/pacemaker_controller.rb
```diff
@@ -14,27 +14,22 @@
 # limitations under the License.
 #
 
-class Api::ClustersController < ApiController
-  api :GET, "/api/clusters", "List all HA clusters"
+class Api::PacemakerController < ApiController
```
jsuchome: So, I've changed it here... or do you want me to leave this change to later cleanup phase, @MaximilianMeister ?

MaximilianMeister: no :) thanks for changing it now

crowbar_framework/config/routes.d/cluster.routes
```diff
@@ -1,7 +1,7 @@
 namespace :api,
   constraints: ApiConstraint.new(2.0) do
-  resources :clusters,
-    only: [:index, :show] do
+  resource :pacemaker, controller: :pacemaker
+    only: [:index ] do
```
MaximilianMeister: maybe we should drop the index as well for now, as we don't implement it yet? that would be:

`resource :pacemaker, controller: :pacemaker, only: [] do`

MaximilianMeister:
+1

aspiers:
Gate failure [openstack-mkcloud#34512 (develcloud7 (5/openvswitch/-t -s/all_noreboot/crowbar/crowbar-ha:138:6bd23540c37d0f5d7959a6881d613468770caa6f:master 1))](https://ci.suse.de/job/openstack-mkcloud/34512/) failed at glance proposal:

```
[2016-09-20T14:17:36+00:00] FATAL: Mixlib::ShellOut::ShellCommandFailed: package[glance] (glance::common line 22) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '8'
---- Begin output of zypper-retry --non-interactive install --auto-agree-with-licenses openstack-glance ----
STDOUT: Loading repository data...
Reading installed packages...
Resolving package dependencies...
The following 30 NEW packages are going to be installed:
  libblas3 libboost_iostreams1_54_0 libboost_random1_54_0 libgfortran3 liblapack3 libquadmath0 librados2 librbd1 lttng-ust openstack-glance python-WSME python-automaton python-castellan python-glance python-glance_store python-networkx python-numpy python-os-brick python-oslo.rootwrap python-oslo.vmware python-osprofiler python-psutil python-semantic_version python-simplegeneric python-taskflow python-urllib3 python-xattr qemu-block-curl qemu-block-rbd qemu-tools
The following 3 recommended packages were automatically selected:
  python-numpy qemu-block-curl qemu-block-rbd
30 new packages to install.
Overall download size: 13.0 MiB. Already cached: 0 B. After the operation, additional 55.9 MiB will be used.
Continue? [y/n/? shows all options] (y): y
Retrieving package libboost_iostreams1_54_0-1.54.0-15.1.x86_64 (1/30),  39.8 KiB (103.1 KiB unpacked)
Abort, retry, ignore? [a/r/i/? shows all options] (a): a
STDERR: Media source 'http://192.168.253.10:8091/suse-12.2/x86_64/install' does not contain the desired medium
Problem occured during or after installation or removal of packages:
Installation aborted by user
Please see the above error message for a hint.
---- End output of zypper-retry --non-interactive install --auto-agree-with-licenses openstack-glance ----
Ran zypper-retry --non-interactive install --auto-agree-with-licenses openstack-glance returned 8
```

I'm guessing this was a transient issue with `clouddata.cloud.suse.de`, so rebuilding...

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -0,0 +1,59 @@
...
+      service_object.check_if_nodes_are_available founders
+
+      founders.each do |n|
+        ssh_retval = n.run_ssh_cmd("crm status 2>&1")
+        if ssh_retval[:exit_code] != 0
+          crm_failures[n.name] = ssh_retval[:stdout]
+          next
+        end
+        ssh_retval = n.run_ssh_cmd('crm status | grep -A 2 "^Failed Actions:"')
+        if ssh_retval[:exit_code] == 0
```
aspiers: I guess there's no reason not to fix this, right?

crowbar_framework/config/routes.d/cluster.routes
aspiers: Is the intention still to drop this?

crowbar_framework/app/models/api/pacemaker.rb
jsuchome: This is actually obsoleted comment, the actual code has `.zero?` already...

crowbar_framework/config/routes.d/cluster.routes
jsuchome: again, the comment is obsolete... :-(

jsuchome:
Finally, successful gating build. Can I get some +1?

MaximilianMeister:
not that it helps, but i am still +1

crowbar_framework/app/models/api/pacemaker.rb
aspiers: Ah. Stupid github!

crowbar_framework/app/models/api/cluster.rb
aspiers: > This is the same check we've done already in Cloud5-6 upgrade.

That doesn't make it right ;-)

> I'm not saying it's perfect, it is actually a notification that something was wrong and the output should be shown to user.

Yes I know that :-) I won't give a -1, but please check with `#sleha` gurus if there is a better approach.

crowbar_framework/app/controllers/api/pacemaker_controller.rb
```diff
@@ -0,0 +1,39 @@
+#
```
aspiers: What is this stupid Hound warning about? I keep seeing it.

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -42,12 +42,12 @@ def health_report
 
       founders.each do |n|
         ssh_retval = n.run_ssh_cmd("crm status 2>&1")
-        if ssh_retval[:exit_code] != 0
+        if (ssh_retval[:exit_code]).nonzero?
```
aspiers: This is fine but for future reference you don't need the extra `()` for this.

aspiers:
+1

PR #139, opened 2016-08-25
The cluster health check happens per cluster

rsalevsky:


nicolasbock:
+1

jsuchome:
How do we list clusters?

rsalevsky:
At the moment we can't but we anyway need to implement this in the model/controller later.

aspiers:
As per https://github.com/crowbar/crowbar-ha/pull/138#issuecomment-242747314 I think this design may be a mistake, and it would be good to discuss it before going further.

MaximilianMeister:
> As per #138 (comment) I think this design may be a mistake, and it would be good to discuss it before going further.

I also think we could do some better accumulated health checking. https://trello.com/c/r09uuhvj/986-crowbar-health-check-api would be great for that, but has to be integrated into the existing upgrade design (especially the [prechecks](https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L28)), and the UI team would need to know yesterday about it.

@jsuchome do you think we have enough time to work on that? It is likely that all of the points @aspiers mentioned are related to a smooth upgrade process for helping to eliminate bugs during the upgrade.

jsuchome:
@MaximilianMeister There are several mixed discussions here

1. I'm not against additional (non-HA) health checks
2. The HA cluster health check for 3.0 is what actually affects the upgrade, and that was already merged as https://github.com/crowbar/crowbar-core/pull/614 . (We can change it to be more generic, not just HA related)
3. Part of the discussion is about where in the master API do those checks belong. I do not have any preference here.

aspiers:
@MaximilianMeister wrote:
> the UI team would need to know yesterday about it.

Why?  It would not be exposed via the UI, only via the API.

MaximilianMeister:
> Why? It would not be exposed via the UI, only via the API.

yes, if it stays like it is now, but if we add other checks, the UI will maybe look different too... but I may be wrong

PR #141, opened 2016-09-13
enable API endpoint to check the ha repos

MaximilianMeister:
see https://github.com/crowbar/crowbar-ceph/pull/68 for the same implentation for ceph

### DIFF
--- a/crowbar_framework/app/controllers/api/clusters_controller.rb
+++ b/crowbar_framework/app/controllers/api/clusters_controller.rb
@@ -15,6 +15,8 @@
 #
 
 class Api::ClustersController < ApiController
+  before_action :set_cluster
+
   api :GET, "/api/clusters", "List all HA clusters"
   api_version "2.0"
   def index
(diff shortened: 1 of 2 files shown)


MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ncdOGOzP&#x2F;97-3-s47p11-implement-api-for-repo-checks-during-upgrade-upgrade-step-8"> (3) S47P11 Implement API for repo checks during upgrade (upgrade step 8)</a>

jsuchome:
+1

MaximilianMeister:
just fixed travis, I forgot to close a block ...

aspiers:
Similarly to what I said in https://github.com/crowbar/crowbar-ha/pull/138#discussion_r78732492, this code is clearly specific to SLE HA repos and hence Pacemaker, so I'm not convinced we should be adding it to a generic `cluster` controller when we could instead remain consistent with Crowbar's existing design and keep it as part of the Pacemaker barclamp.  -1 from me until we have a mumble discussion so I can understand if there are really good reasons to deviate from the existing design pattern.

MaximilianMeister:
> -1 from me until we have a mumble discussion so I can understand if there are really good reasons to deviate from the existing design pattern.

is the -1 for the design overall or specific to this change? we have it the same way in ceph right now, and I think we could change it afterwards even we merge this before

MaximilianMeister:
closing in favor of https://github.com/crowbar/crowbar-ha/pull/144

PR #142, opened 2016-09-16
Rename HAE to HA

dirkmueller:
This extra 'E' got dropped with SLE12, so remove it to avoid confusion.


rsalevsky:
+1

nicolasbock:
So we need github approvals _and_ lgtm approvals now?

aspiers:
@nicolasbock That should be discussed with the whole team on `cloud-devel`, or even better, the Crowbar list ...

PR #143, opened 2016-09-19
s390x enablement

dirkmueller:

PR #145, opened 2016-09-23
Switch to node[:crowbar_ohai] for getting libvirt uuid

rhafer:
node[:crowbar_ohai][:libvirt][:guestuuid] is maintained by crowbar's
ohai plugin to avoid the use of the architecture specific node[:dmi]
subtree.

Alternative to #143
Requires: crowbar/crowbar-core#677

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
@@ -133,16 +133,15 @@ when "libvirt"
     CrowbarPacemakerHelper.remote_nodes(node)
 
   all_nodes.each do |cluster_node|
-    manufacturer = cluster_node[:dmi][:system][:manufacturer] rescue "unknown"
-    unless %w(Bochs QEMU).include? manufacturer
+    unless cluster_node[:crowbar_ohai][:libvirt][:guest_uuid]
       message = "Node #{cluster_node[:hostname]} does not seem to be running in libvirt."
       Chef::Log.fatal(message)
       raise message
     end
 
     # We need to know the domain to interact with for each cluster member; it
-    # turns out that libvirt puts the domain UUID in DMI
-    domain_id = cluster_node[:dmi][:system][:uuid]
+    # turns out the domain UUID is accessible via ohai
+    domain_id = cluster_node[:crowbar_ohai][:libvirt][:guest_uuid]
 
     stonith_node_name = pacemaker_node_name(cluster_node)
 
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -547,13 +547,8 @@ class PacemakerService < ServiceObject
       end
       members.each do |member|
         node = NodeObject.find_node_by_name(member)
-        manufacturer = node[:dmi][:system][:manufacturer] rescue "unknown"
-        unless %w(Bochs QEMU).include? manufacturer
-          validation_error I18n.t(
-            "barclamp.#{bc_name}.validation.libvirt",
-            member: member
-          )
-        end
+        next if node[:crowbar_ohai][:libvirt][:guest_uuid]
+        validation_error I18n.t("barclamp.#{bc_name}.validation.libvirt", member: member)
       end
     else
       validation_error I18n.t(


rhafer:
Alternative to https://github.com/crowbar/crowbar-ha/pull/143
requires: https://github.com/crowbar/crowbar-core/pull/677

rhafer:
> Are there any architectures which use libvirt and don't have that subtree?

@aspiers: s390x :-) (we use libvirt/kvm there in the ci to some extend), not sure about arm.

rhafer:
@aspiers Updated the commit message. Hopefully that makes things clearer.

vuntz:
See my comment in the other PR for some small simplification here...

rhafer:
> See my comment in the other PR for some small simplification here...

Updated.

chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
```diff
@@ -133,16 +133,15 @@ def pacemaker_node_name(n)
     CrowbarPacemakerHelper.remote_nodes(node)
 
   all_nodes.each do |cluster_node|
-    manufacturer = cluster_node[:dmi][:system][:manufacturer] rescue "unknown"
-    unless %w(Bochs QEMU).include? manufacturer
+    unless cluster_node[:crowbar_ohai][:libvirt][:guest_uuid]
```
aspiers: Isn't this going to crash if `cluster_node[:crowbar_ohai][:libvirt]` is `nil`?

chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
```diff
@@ -133,16 +133,15 @@ def pacemaker_node_name(n)
...
       message = "Node #{cluster_node[:hostname]} does not seem to be running in libvirt."
       Chef::Log.fatal(message)
       raise message
     end
 
     # We need to know the domain to interact with for each cluster member; it
-    # turns out that libvirt puts the domain UUID in DMI
-    domain_id = cluster_node[:dmi][:system][:uuid]
+    # turns out the domain UUID is accessible via ohai
+    domain_id = cluster_node[:crowbar_ohai][:libvirt][:guest_uuid]
```
aspiers: Same question here.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -547,8 +547,7 @@ def validate_proposal_stonith stonith_attributes, members
       end
       members.each do |member|
         node = NodeObject.find_node_by_name(member)
-        manufacturer = node[:dmi][:system][:manufacturer] rescue "unknown"
-        unless %w(Bochs QEMU).include? manufacturer
+        unless node[:crowbar_ohai][:libvirt][:guest_uuid]
```
aspiers: And here.  Also, I think Hound was suggesting that

    next if node[:crowbar_ohai][:libvirt][:guest_uuid]

is a bit cleaner here.  In general I would agree, although I suppose there is the possibility to add other checks inside this loop, in which case it might or might not make sense.

chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
dirkmueller: same answer here

crowbar_framework/app/models/pacemaker_service.rb
rhafer: Updated. Though I am not sure which variant is really cleaner here.

dirkmueller:
@aspiers don't unassign yourself when you're the one requesting changes ;)

chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
aspiers: A.k.a. "what could possibly go wrong?" :-P  In light of Murphy's Law, I always prefer defensive programming (and this is by far one of the most common issues we experience in our recipes), but I guess you just volunteered to fix anything which goes wrong, so I'm OK with it ;-)

aspiers:
@dirkmueller Why not? Makes sense to me. Please see https://github.com/isaacs/github/issues/783#issuecomment-250482249

aspiers:
(but basically GitHub's review system currently sucks so there is no good solution here except maybe using Reviewable or Gitcolony or GerritHub, which I am beginning to think we should seriously consider)

chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
vuntz: Heh, aren't you the one wondering about my defensive programming style from time to time? :-) Be sure that I'll remember this comment and will link to it in the future!

PR #146, opened 2016-09-26
Check if the node is currently in some upgrade phase and do not sync with other nodes

jsuchome:
I'm not sure if this is the right place, and it is not tested yet. More like initial idea open for comments.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -31,6 +31,12 @@ module CrowbarPacemakerHelper
     node[:pacemaker][:founder]
   end
 
+  # Check if the node is currently in some upgrade phase
+  def self.being_upgraded?(node)
+    upgrade_step = node["crowbar_wall"]["crowbar_upgrade_step"] || "none"
+    upgrade_step == "prepare-os-upgrade"
+  end
+
   # Returns the number of corosync (or non-remote) nodes in the cluster.
   def self.num_corosync_nodes(node)
     return 0 unless cluster_enabled?(node)
--- a/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
@@ -60,6 +60,11 @@ module CrowbarPacemakerSynchronization
   def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout = 60)
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
     return if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    if CrowbarPacemakerHelper.being_upgraded?(node)
+      Chef::Log.debug("Node is being upgraded." \
+        "Skipping wait loop for cluster founder.")
+      return
+    end
 
     cluster_name = CrowbarPacemakerHelper.cluster_name(node)
 
@@ -127,6 +132,12 @@ module CrowbarPacemakerSynchronization
 
     node_count = CrowbarPacemakerHelper.cluster_nodes(node).length
     raise "No member in the cluster!" if node_count == 0
+
+    if CrowbarPacemakerHelper.being_upgraded?(node)
+      Chef::Log.debug("Node is being upgraded." \
+        "Skipping wait loop for all other cluster nodes.")
+      return
+    end
 
     Chef::Log.info("Checking if all cluster nodes have set #{mark} to #{revision}...")
 


vuntz:
Wondering out loud: since the plan is to make the first node to be upgraded the cluster "founder", shouldn't we let the node set the marks (but not block on the other nodes)?

jsuchome:
I don't know if we'd start with the cluster founder. So far, the only dependency we've thought about was starting with DRBD slave node.

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -31,6 +31,12 @@ def self.is_cluster_founder?(node)
     node[:pacemaker][:founder]
   end
 
+  # Check if the node is currently in some upgrade phase
+  def self.being_upgraded?(node)
+    upgrade_step = node["crowbar_wall"]["crowbar_upgrade_step"] || "none"
+    ["prepare-os-upgrade", "os-upgrading"].include? upgrade_step
```
jsuchome: BTW, we do not have any "os-upgrading" step so far. We need to figure out which steps have to be checked here.

aspiers:
That's correct. There's no intention to start with the founder; it has to be the DRBD slave. If DRBD doesn't exist in the cluster then we could start with the founder, but I don't think it matters whether we do.

vuntz:
Sure, but didn't we say we would make the first node we upgrade the founder (which possibly implies the transfer of the role from another node)?

jsuchome:
What would be the gain of setting marks in partial (one node only) chef client run? 

What if we run full chef client again once all controllers are upgraded (and nodes are not in upgrade 'mode' any more), so the marks would be set correctly?

vuntz:
> What would be the gain of setting marks in partial (one node only) chef client run?

So the idea would be to avoid later chef-client runs to fail because of sync marks (assuming that such runs might not happen at the same time on all nodes -- this could be true due to what we do in the upgrade process, or simply because of the periodic chef-client run).

That being said, it requires setting the sync marks not just on the founder, but on all nodes. So I'm not fully sure my point about the founder was fully correct wrt this.

The only part where it's correct is that we need the chef recipe to run some code that would run on founders only (like db syncs), so I think we definitely need to transfer the "founder" bit to the first node.

> What if we run full chef client again once all controllers are upgraded (and nodes are not in upgrade 'mode' any more), so the marks would be set correctly?

Sure, that's also an option, but this will slow down things.

aspiers:
> > What would be the gain of setting marks in partial (one node only) chef client run?
> 
> So the idea would be to avoid later chef-client runs to fail because of sync marks (assuming that such runs might not happen at the same time on all nodes -- this could be true due to what we do in the upgrade process

Correct - I don't think we'll ever do a run at the same time on all nodes... OK maybe we could do one right near the end.
But why would they fail anyway, if we're disabling sync checks during upgrade?  I don't see your point.

> or simply because of the periodic chef-client run

Periodic runs are disabled during upgrade.  It's a given that chef-client runs will happen again after the upgrade, of course :-)

> That being said, it requires setting the sync marks not just on the founder, but on all nodes. So I'm not fully sure my point about the founder was fully correct wrt this.

Are you worried about the sync mark revision numbers getting out of sync once we're post upgrade or something?  That seems like a valid concern - presumably revision numbers will be bumped by barclamp schema upgrades?  If so, could we deal with that by unconditionally setting sync mark revision numbers during upgrade mode, and only skipping the waits?

> The only part where it's correct is that we need the chef recipe to run some code that would run on founders only (like db syncs), so I think we definitely need to transfer the "founder" bit to the first node.

Yes, that's a very good point.

> > What if we run full chef client again once all controllers are upgraded (and nodes are not in upgrade 'mode' any more), so the marks would be set correctly?
> 
> Sure, that's also an option, but this will slow down things.

Would my above proposal address this?

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -95,6 +96,7 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
   def self.set_mark_if_founder(node, mark, revision)
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
     return unless CrowbarPacemakerHelper.is_cluster_founder?(node)
+    return if CrowbarPacemakerHelper.being_upgraded?(node)
```
jsuchome: @vuntz so am I right, based on the discussions, that we do not need this check here, in `set_mark_if_founder`?

rhafer:
> Are you worried about the sync mark revision numbers getting out of sync once we're post upgrade or something? That seems like a valid concern - presumably revision numbers will be bumped by barclamp schema upgrades?

@aspiers Currently they are not bumped during the schema upgrades. Why would they?

>  If so, could we deal with that by unconditionally setting sync mark revision numbers during upgrade mode, and only skipping the waits?

Isn't that what @vuntz suggested in the first comment. Though I am not sure I really understand why hat would be required.

Wouldn't it make sense to just reset the revisions on all nodes once the first controller node got upgraded? (I really hope this question doesn't make it too obvious how little a actually  still know about the upgrade process :-) )

jsuchome:
Reviews are still missing here, @vuntz , @aspiers @rhafer ...

aspiers:
> > Are you worried about the sync mark revision numbers getting out of sync once we're post upgrade or something? That seems like a valid concern - presumably revision numbers will be bumped by barclamp schema upgrades?
> 
> @aspiers Currently they are not bumped during the schema upgrades.

Are you sure?  `migrate_role` calls `role.save` which calls `RoleObject#increment_crowbar_revision!`.

> Why would they?

Because revision numbers are supposed to be bumped every time the proposal is changed.

> >  If so, could we deal with that by unconditionally setting sync mark revision numbers during upgrade mode, and only skipping the waits?
> 
> Isn't that what @vuntz suggested in the first comment.

Yes I think so.

> Though I am not sure I really understand why hat would be required.

Setting sync marks is necessary so that once we've finished the upgrade, periodic chef-client runs don't time out due to missing sync marks.  Skipping the waits is necessary because we're upgrading nodes one by one.

> Wouldn't it make sense to just reset the revisions on all nodes once the first controller node got upgraded?

Resetting revisions doesn't make sense; they're supposed to be monotonically increasing.  Do you mean resetting (as in, removing) sync marks?  And by "once the first controller node got upgraded", you mean after its `chef-client` run?  If we did that, then at the end of the upgrade the first node would be missing sync marks, which would cause its first periodic `chef-client` run to fail.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -112,6 +113,7 @@ def self.set_mark_if_founder(node, mark, revision)
   # See "Synchronization helpers" documentation
   def self.synchronize_on_mark(node, mark, revision, fatal = false, timeout = 60)
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
+    return if CrowbarPacemakerHelper.being_upgraded?(node)
```
aspiers: So based on the PR discussion I think this one is too early.  It should come after the mark has been set, but before the timeout loop.

jsuchome: I am certainly missing something, but which point of discussion leads to this?

rhafer:
> > @aspiers Currently they are not bumped during the schema upgrades.
> 
> Are you sure? migrate_role calls role.save which calls RoleObject#increment_crowbar_revision!.

Sorry. Yeah you're right. I mixed up sync_marks and revision numbers and wondered why sync_marks would be touched during schema migrations :/

> Resetting revisions doesn't make sense; they're supposed to be monotonically increasing. Do you mean resetting (as in, removing) sync marks? And by "once the first controller node got upgraded", you mean after its chef-client run?

Not "as in, removing". I was thinking about either setting them to the correct value (i.e. the current revison number of the barclamp) all at once. Or to reset all revisions and sync_marks to 0. But thinking about this more I don't think it gives any advantage over just setting the marks but skipping the waits during the upgrade.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
rhafer: Mainly @aspiers' reply (https://github.com/crowbar/crowbar-ha/pull/146#issuecomment-258047440) to my older comment leads to this conclusion.

jsuchome: So I've moved the check.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -128,6 +129,8 @@ def self.synchronize_on_mark(node, mark, revision, fatal = false, timeout = 60)
     node_count = CrowbarPacemakerHelper.cluster_nodes(node).length
     raise "No member in the cluster!" if node_count == 0
 
+    return if CrowbarPacemakerHelper.being_upgraded?(node)
```
rhafer: I think it would be good to also log something here when skipping the wait loop. Should make debugging this a bit easier.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -60,6 +60,7 @@ module CrowbarPacemakerSynchronization
   def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout = 60)
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
     return if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    return if CrowbarPacemakerHelper.being_upgraded?(node)
```
rhafer: I think it would be good to also log something here when skipping the wait loop. Should make debugging this a bit easier.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -60,6 +60,10 @@ module CrowbarPacemakerSynchronization
   def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout = 60)
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
     return if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    if CrowbarPacemakerHelper.being_upgraded?(node)
+      Chef::Log.debug("Node is in upgrade phase, not waiting for sync marc")
```
rhafer: How about ```The cluster is being upgraded. Skipping wait loop for cluster founder```

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -128,6 +132,11 @@ def self.synchronize_on_mark(node, mark, revision, fatal = false, timeout = 60)
     node_count = CrowbarPacemakerHelper.cluster_nodes(node).length
     raise "No member in the cluster!" if node_count == 0
 
+    if CrowbarPacemakerHelper.being_upgraded?(node)
+      Chef::Log.debug("Node is in upgrade phase, not synchronizing")
```
rhafer: How about ```The cluster is being upgraded. Skipping wait loop for all other cluster nodes.```

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
jsuchome: But the check is really for node, not for cluster. Other node in cluster might already be upgraded

rhafer: Which still somehow means that the cluster is being upgraded :). But my main concern wasn't about the cluster vs. node thing (I am fine with either). It was just a suggestion to improve the wording (and to fix the typo).

PR #147, opened 2016-09-29
Simple check for presence of HA-based configuration

jsuchome:
This is to be used by crowbar pre-checks, run before the upgrade

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -17,6 +17,15 @@
 module Api
   class Pacemaker < Tableless
     class << self
+      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_presence_check
+        unless repocheck["ha"]["available"]
+          return { errors: [I18n.t("api.pacemaker.ha_not_installed")] }
+        end
+        founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+        founders.empty? ? { errors: [I18n.t("api.pacemaker.ha_not_configured")] } : {}
+      end
+
       # Simple check if HA clusters report some problems
       # If there are no problems, empty hash is returned.
       # If this fails, information about failed actions for each cluster founder is
--- a/crowbar_framework/config/locales/pacemaker/en.yml
+++ b/crowbar_framework/config/locales/pacemaker/en.yml
@@ -18,6 +18,10 @@
 en:
   nav:
     pacemaker: 'Pacemaker'
+  api:
+    pacemaker:
+      ha_not_installed: 'HA add-on is not installed.'
+      ha_not_configured: 'No Pacemaker cluster is configured.'
   barclamp:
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -17,6 +17,20 @@
 module Api
   class Pacemaker < Tableless
     class << self
+      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_presence_check
+        ret = {}
```
MaximilianMeister: matter of taste but I think we could use tap here, which always returns the hash as we want it.
and currently you are overriding the array if both errors happen, I would suggest sth like:

```ruby
{}.tap do |ret|
  unless repocheck["ha"]["available"]
    ret[:errors] = []
    ret[:errors].push(I18n.t("api.pacemaker.ha_not_installed"))
    return ret
  end
  founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
  if founders.empty?
    ret[:errors] ||= []
    ret[:errors].push(I18n.t("api.pacemaker.ha_not_configured"))
  end
end
```

aspiers: That's strangely verbose.  Why not just this?

```ruby
unless repocheck["ha"]["available"]
  return { errors: [ I18n.t("api.pacemaker.ha_not_installed") ] }
end
founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
founders.empty? ? { errors: [ I18n.t("api.pacemaker.ha_not_configured") ] } : {}
```

jsuchome: Does this really matter that much?

jsuchome:
I do not have any special reason for that `tap`, I've added it based on @MaximilianMeister suggestion. In general, I do not find it better or worse with my original approach, Max's tap, or Adam's shorter one.  A little bit more or less lines, and all examples are readable ... then I just tend to adapt to the needs of reviewers to get the stuff done.

aspiers:
I'm not religiously against the use of `tap` but I would just like to understand if there is a reason for using it which I'm missing.

jsuchome:
> but I would just like to understand if there is a reason for using it which I'm missing.

I don't know, I believe it's matter of taste. As I wrote, I followed Max's recommendation. @MaximilianMeister ?

MaximilianMeister:
> I don't know, I believe it's matter of taste. As I wrote, I followed Max's recommendation. @MaximilianMeister ?

I don't mind either way is fine with me

> I'm not religiously against the use of tap but I would just like to understand if there is a reason for using it which I'm missing.

it was just a suggestion of how I would do it, doesn't mean that there is no better way of doing it

jsuchome:
Adapted according to @aspiers suggestions

MaximilianMeister:
gating error is unrelated, and this code is not used at all. we need it as well for backporting it to be used by the prechecks API

PR #148, opened 2016-10-05
fix corosync.conf values (bsc#1001164)

aspiers:
Our `corosync.conf.v2` template was based on a version from SLE HA which had some tunings accidentally dropped in the migration from SLE11.  In particular, some of the timeouts were too aggressive.


aspiers:
Alex Runge has [tested this in a PTF](https://bugzilla.suse.com/show_bug.cgi?id=1001164#c29), so it should be OK to merge after the freeze.

PR #149, opened 2016-10-06
Helpers for creating upgrade related location constraints

jsuchome:
Requires https://github.com/crowbar/crowbar-ha/pull/146

Usage in a recipe:

```
upgrade_location_name = upgraded_only_location_for clone_name
transaction_objects << "pacemaker_location[#{upgrade_location_name}]" if CrowbarPacemakerHelper.being_upgraded?(node)
```

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/definitions/upgraded_only_location_for.rb
+++ b/chef/cookbooks/crowbar-pacemaker/definitions/upgraded_only_location_for.rb
@@ -1,0 +1,26 @@
+#
+# Copyright 2016, SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+define :upgraded_only_location_for do
+  resource = params[:name]
+  location_name = "l-#{resource}-upgraded-only"
+  pacemaker_location location_name do
+    definition CrowbarPacemakerHelper.upgraded_only_location(location_name, resource)
+    action :update
+    only_if { CrowbarPacemakerHelper.being_upgraded?(node) }
+  end
+  location_name
+end
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -291,4 +291,12 @@ module CrowbarPacemakerHelper
 
     servers
   end
+
+  # Helper method to provides a convenient way to create Pacemaker
+  # location constraints which limit resources to only run on node that has already been upgraded
+  def self.upgraded_only_location(location, service)
+    # resource-discovery=exclusive means that Pacemaker will detect
+    # if a resource is running where it shouldn't be (i.e. on a pre-upgrade node), and then stop it
+    "location #{location} #{service} resource-discovery=exclusive rule 0: pre-upgrade eq false"
+  end
 end


jsuchome:
@aspiers I think you suggested another library for `self.upgraded_only_location`, but this looks like a correct place to me...

jsuchome:
[(13) Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2)](https://trello.com/c/2PMGauJx/118-13-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2)

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -291,4 +291,10 @@ def self.haproxy_servers_for_service(node, name, role, ports_key)
 
     servers
   end
+
+  # Helper method to provides a convenient way to create Pacemaker
+  # location constraints which limit resources to only run on node that has already been upgraded
+  def self.upgraded_only_location(location, service)
+    "location #{location} #{service} resource-discovery=exclusive rule 0: pre-upgrade eq false"
```
aspiers: To be clear, IIUC `resource-discovery=exclusive` means that Pacemaker will detect if a resource is running where it shouldn't be (i.e. on a pre-upgrade node), and then stop it:

http://clusterlabs.org/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/_deciding_which_nodes_a_resource_can_run_on.html

(Due to the existing controller-only constraint which has the same option enabled, it should only perform the discovery checks on controller nodes, not compute nodes.)

So I think we're right to include this, but it's worth a comment explaining it, since it's really not obvious.

aspiers:
> @aspiers I think you suggested another library for self.upgraded_only_location, but this looks like a correct place to me...

Yeah I was thinking it would be worth having a separate file for upgrade-specific stuff.  `helpers.rb` is already quite long.  Having said that it's not a big deal.

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
aspiers: *UPDATE*: @gao-yan thinks it should work even without this setting, but I think we should keep it anyway.

chef/cookbooks/crowbar-pacemaker/definitions/upgraded_only_location_for.rb
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+define :upgraded_only_location_for do
+  resource = params[:name]
+  location_name = "l-#{resource}-upgraded-only"
```
jsuchome: 1. I wanted to make it similar to the other definition we have (`openstack_pacemaker_controller_only_location_for.rb`) where we have the name also inside

2. Yes, it's confusing, I was meaning "helper" in a way it is "helping". I'll adapt the commit message

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
jsuchome: Thanks for the research. I've added the comment

aspiers:
@jsuchome It's a very small nitpick which doesn't require a change to this PR, but I forgot to say before - I think ideally in the future, pairs of commits like these would probably be squashed together. I don't think it makes sense to add a tiny method in one commit and then add its only consumer in a second commit, because the dependency is so strong and they are both such small changes. The second commit cannot function without the first, and the first only exists to serve the second.  For example it would be possible to revert the second commit and accidentally forget to revert the first.  

Another case of something like this would be adding a new method and its unit test - although in that case the argument for squashing into one commit is much stronger.

It's a stylistic decision so you may disagree (which is fine) - worth bearing in mind for the future though :-)

chef/cookbooks/crowbar-pacemaker/definitions/upgraded_only_location_for.rb
```diff
@@ -0,0 +1,26 @@
...
+  pacemaker_location location_name do
+    definition CrowbarPacemakerHelper.upgraded_only_location(location_name, resource)
+    action :update
+    only_if { CrowbarPacemakerHelper.being_upgraded?(node) }
+  end
+  location_name
```
aspiers: Hmm, ideally this should return `nil` if the node isn't being upgraded, but I'm not sure that's possible with Chef, since the `only_if` block's evaluation presumably happens in a different phase.

chef/cookbooks/crowbar-pacemaker/definitions/upgraded_only_location_for.rb
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+#
+
+define :upgraded_only_location_for do
+  resource = params[:name]
+  location_name = "l-#{resource}-upgraded-only"
+  pacemaker_location location_name do
+    definition CrowbarPacemakerHelper.upgraded_only_location(location_name, resource)
+    action :update
+    only_if { CrowbarPacemakerHelper.being_upgraded?(node) }
```
vuntz: The one thing I'm not fully sure about: is it fine if this runs eventually on all nodes? Or does it have to run only on the first controller? Or does it not matter because it will actually be run only on the first controller anyway?

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -291,4 +291,12 @@ def self.haproxy_servers_for_service(node, name, role, ports_key)
 
     servers
   end
+
+  # Helper method to provides a convenient way to create Pacemaker
+  # location constraints which limit resources to only run on node that has already been upgraded
+  def self.upgraded_only_location(location, service)
```
vuntz: For the record, the thread at https://github.com/crowbar/crowbar-openstack/pull/545/files#r82961539 suggests that we might need to change this to make it easier to call.

chef/cookbooks/crowbar-pacemaker/definitions/upgraded_only_location_for.rb
vuntz: Thanks, so I guess it's fine if it runs everywhere (as it will have been created on the first node earlier, so no race)

aspiers: That's right.

chef/cookbooks/crowbar-pacemaker/definitions/upgraded_only_location_for.rb
aspiers: No, I think it's a different scenario, because with the `pre-upgrade` constraint, there is a time (in normal proposal application, before *any* upgrade workflow happens) where we want to avoid the constraint from being created.  Whereas with the controller-only constraint, we always want it to exist, and it will be created by the founder at the earliest opportunity.

chef/cookbooks/crowbar-pacemaker/definitions/upgraded_only_location_for.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+define :upgraded_only_location_for do
+  resource = params[:name]
```
aspiers: As per https://github.com/crowbar/crowbar-openstack/pull/545/files#r83259814 I think we should wrap the body of this definition in an `if CrowbarPacemakerHelper.being_upgraded?(node)` conditional.  This will of course require passing `node` as a parameter.

jsuchome:
@aspiers @vuntz I've added 3rd commit that is an alternative to the previous two (I should probably create new PR instead...) 
That one adds helper modifying the list of given transaction objects, which is basically the only thing we want from newly created locations. I think it addresses all the problems raised so far.

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -291,4 +291,26 @@ def self.haproxy_servers_for_service(node, name, role, ports_key)
...
+  def self.add_upgraded_only_location(node, transaction_objects, resources)
+    return transaction_objects unless CrowbarPacemakerHelper.being_upgraded?(node)
+    location = "l-#{resource}-upgraded-only"
+    pacemaker_location location_name do
+      definition "location #{location} #{service} resource-discovery=exclusive " \
+        "rule 0: pre-upgrade eq false"
+      action :update
+    end
+    transaction_objects << "pacemaker_location[#{upgrade_location_name}]"
+    transaction_objects
```
vuntz: Generally speaking, I like this helper more than the other approach.

jsuchome:
Should be obsoleted by https://github.com/crowbar/crowbar-openstack/pull/562

PR #150, opened 2016-10-07
Fix undefined method `sort' for nil:NilClass

mjura:
It can happened that deployment["elements_expanded"][role_name]
can be nil. This will cause

"Failed to apply the proposal: exception before calling chef (undefined method `sort' for nil:NilClass)"

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -153,7 +153,8 @@ class PacemakerService < ServiceObject
         end
 
         expanded_nodes.sort!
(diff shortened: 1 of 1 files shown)


rsalevsky:
Feel free to merge once gating is green.

mjura:
@aspiers: I fixed commit message to:

```
Fix undefined method `sort' for nil:NilClass

It happened after reinstallation one of my compute nodes (which acts as pacemaker-remote).
The pacemaker barclamp couldn't apply anymore and it throws the error

"Failed to apply the proposal: exception before calling chef (undefined method `sort' for nil:NilClass)"
```

vuntz:
@aspiers are you fine with the updated commit message?

PR #152, opened 2016-10-14
Method for changing the cluster founder to given node

jsuchome:
This is supposed to be called from `models/api/upgrade.rb` after the first node in cluster is upgraded.

So... the logic should be correct, but the question is if the method belongs in this place or elsewhere.

We could split it into pieces (like `cluster_for_node` etc.) and put them into PacemakerServiceObject, but that does not seem right as well, because we do not need the service object for anything.

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -92,11 +92,17 @@ module Api
           Rails.logger.warning("No cluster founder found. Making #{name} the new founder anyway.")
         else
           old_founder[:pacemaker][:founder] = false
+          if old_founder[:drbd] && old_founder[:drbd][:rsc]
+            old_founder[:drbd][:rsc].each { |_, res| res[:master] = false }
+          end
           old_founder.save
         end
 
         # 3. mark given node as founder
         new_founder[:pacemaker][:founder] = true
+        if new_founder[:drbd] && new_founder[:drbd][:rsc]
+          new_founder[:drbd][:rsc].each { |_, res| res[:master] = true }
+        end
         new_founder.save
       end
 
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -66,6 +66,40 @@ module Api
         ret
       end
 
+      def set_node_as_founder(name)
+        # 1. find the cluster new founder is in
+        new_founder = NodeObject.find_node_by_name(name)
+        if new_founder.nil?
+          Rails.logger.error("Node #{name} not found!")
+          return false
+        end
+        unless new_founder[:pacemaker]
+          Rails.logger.error("Node #{name} does not have pacemaker setup")
+          return false
+        end
+        if new_founder[:pacemaker][:founder]
+          Rails.logger.debug("Node #{name} is already the cluster founder.")
+          return true
+        end
+
+        # 2. find the current cluster founder in the same cluster
+        cluster_env = new_founder[:pacemaker][:config][:environment]
+        old_founder = NodeObject.find(
+          "pacemaker_founder:true AND pacemaker_config_environment:#{cluster_env}"
+        ).first
+
+        if old_founder.nil?
+          Rails.logger.warning("No cluster founder found. Making #{name} the new founder anyway.")
+        else
+          old_founder[:pacemaker][:founder] = false
+          old_founder.save
+        end
+
+        # 3. mark given node as founder
+        new_founder[:pacemaker][:founder] = true
+        new_founder.save
+      end
+
       def repocheck
         Api::Node.repocheck(addon: "ha")
       end


crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -57,6 +57,25 @@ def health_report
         ret
       end
 
+      def set_node_as_founder(name)
+        # 1. find the cluster new founder is in
+        new_founder = NodeObject.find_node_by_name(name)
```
jsuchome: It's possible we would have the correct node object already in the caller of this method, so we could already pass it as argument...

vuntz: If that happens, it is actually important to either use the object from the caller, or to re-fetch the node object in the caller (because of the `save`).

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -57,6 +57,25 @@ def health_report
...
+        return true if new_founder[:pacemaker][:founder]
+
+        # 2. find the current cluster founder in the same cluster
+        cluster_env = new_node[:pacemaker][:config][:environment]
+        old_founder = NodeObject.find(
+          "pacemaker_founder:true AND pacemaker_config_environment:#{cluster_env}"
+        ).first
+
+        old_founder[:pacemaker][:founder] = false
+        old_founder.save
```
vuntz: Just to be safe, I'd handle the `old_founder.nil?` case.

crowbar_framework/app/models/api/pacemaker.rb
jsuchome: true

crowbar_framework/app/models/api/pacemaker.rb
jsuchome: Done (simply only)

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -57,6 +57,26 @@ def health_report
...
+        # 1. find the cluster new founder is in
+        new_founder = NodeObject.find_node_by_name(name)
+        return true if new_founder[:pacemaker][:founder]
+
+        # 2. find the current cluster founder in the same cluster
+        cluster_env = new_node[:pacemaker][:config][:environment]
+        old_founder = NodeObject.find(
+          "pacemaker_founder:true AND pacemaker_config_environment:#{cluster_env}"
+        ).first
+        return false if old_founder.nil?
```
jsuchome: Well... it's probably broken when there's no founder. But OK, I could set the new one.

jsuchome:
[(13) S49P16 Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2)](https://trello.com/c/2PMGauJx/118-13-s49p16-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2)

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -57,6 +57,32 @@ def health_report
         ret
       end
 
+      def set_node_as_founder(name)
+        # 1. find the cluster new founder is in
+        new_founder = NodeObject.find_node_by_name(name)
+        if new_founder[:pacemaker][:founder]
```
rhafer: Do we need to worry about ```new_founder.nil?``` here. Or the node not having the pacemaker attributes?

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -57,6 +57,32 @@ def health_report
...
+        end
+
+        # 2. find the current cluster founder in the same cluster
+        cluster_env = new_node[:pacemaker][:config][:environment]
+        old_founder = NodeObject.find(
+          "pacemaker_founder:true AND pacemaker_config_environment:#{cluster_env}"
+        ).first
+
+        if old_founder.nil?
+          Rails.logger.warning("No cluster founder found. Setting #{name} anyway.")
```
rhafer: I'd suggest ```" ... . Making #{name} the new founder anyway."``` here

rhafer:
Testing failed because of an unrelated issue (while trying to test ceph). Tempest succeded so I'd say is fine to merge this. @aspiers  or  @vuntz cloud you please merge?

jsuchome:
I've added new commit for handing DRBD configuration (originally coming from https://github.com/crowbar/crowbar-ha/pull/155).

jsuchome:
@rhafer @aspiers so I have approvals from you and even green gating, but unfortunatelly new commit was added meanwhile.. are you fine with that one?

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -57,6 +57,42 @@ def health_report
         ret
       end
 
+      def set_node_as_founder(name)
```
rsalevsky: I would just call it `founder_node`.

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -83,11 +83,13 @@ def set_node_as_founder(name)
           Rails.logger.warning("No cluster founder found. Making #{name} the new founder anyway.")
         else
           old_founder[:pacemaker][:founder] = false
+          old_founder[:drbd][:rsc].each { |_, res| res[:master] = false } if old_founder[:drbd]
```
aspiers: Can you be sure that `old_founder[:drbd][:rsc]` will always exist if `old_founder[:drbd]` does?  Just checking ...

aspiers: Also please fix the commit message to explain *why* the change is necessary. I wish I didn't have to ask for this in each PR :-/

jsuchome:
@aspiers commit messages adapted. But please do not talk about the need for reverse engineering. When I added the commit with DRBD change, I referenced original PR that was handling it separately

aspiers:
@jsuchome Thanks a lot, that's much better!  I was probably being a bit over-dramatic by referring to "reverse engineering", but I was still talking about the need for commit messages to be a self-contained explanation for the changes, rather than relying on external explanations such as info in the PR.  This relates to [the second point here](https://wiki.openstack.org/wiki/GitCommitMessages#Information_in_commit_messages), but it also eliminates the need to read the entire PR history when doing a review.  If we can follow these guidelines by default then it will speed up code reviews a lot.

jsuchome:
Gating has failed in cct, which is not relevant to this PR (as the newly added methods are not used anywhere so far). Could we merge?

PR #153, opened 2016-10-18
Helper to modify given transaction_objects by adding location constraint

jsuchome:
for nodes that are being upgraded. Such constraint ensures that
the services are started only on node that does not have pre-upgrade
attribute.

This is an alternative to https://github.com/crowbar/crowbar-ha/pull/149

Example usage from the recipe:
```
transaction_objects = CrowbarPacemakerHelper.add_upgraded_only_location(
  node, transaction_objects, clone_name)
```

Should be obsoleted by https://github.com/crowbar/crowbar-openstack/pull/562

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -291,4 +291,19 @@ module CrowbarPacemakerHelper
 
     servers
   end
+
+  # Helper method that adds a Pacemaker location constraint to the given list of transation objects.
+  # Such constraint limits resources to only run on node that has already been upgraded.
+  def self.add_upgraded_only_location(node, transaction_objects, resource)
+    return transaction_objects unless CrowbarPacemakerHelper.being_upgraded?(node)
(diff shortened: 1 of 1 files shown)


chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -291,4 +291,19 @@ def self.haproxy_servers_for_service(node, name, role, ports_key)
...
     servers
   end
+
+  # Helper method that adds a Pacemaker location constraint to the given list of transation objects.
+  # Such constraint limits resources to only run on node that has already been upgraded.
+  def self.add_upgraded_only_location(node, transaction_objects, resource)
+    return transaction_objects unless CrowbarPacemakerHelper.being_upgraded?(node)
+    return transaction_objects unless CrowbarPacemakerHelper.is_cluster_founder?(node)
+    location = "l-#{resource}-upgraded-only"
+    pacemaker_location location do
```
rhafer: I am not entirely sure (didn't try) but what might work here is:

```
l = Chef::Resource::Pacemaker::Location.new(location, run_context)
l.defintion(".....")
l.run_action(:update)
```

PR #154, opened 2016-10-19
Do not use node.ipaddress, but IP from the 'admin' range

jsuchome:
With 'ipaddress', it sometimes (??) uses the "public" IP instead... 
Specifically, after the upgrade, when running chef-client on SP2 node, and this results in having differnt IP's in the config of upgraded and non-upgraded node

### DIFF
--- a/chef/cookbooks/drbd/providers/resource.rb
+++ b/chef/cookbooks/drbd/providers/resource.rb
@@ -31,6 +31,9 @@ action :create do
   raise "Remote node #{remote_host} not found!" if remote_nodes.empty?
   remote = remote_nodes.first
 
+  ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+  remote_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(remote, "admin").address
+
   drbd_resource_template = template "/etc/drbd.d/#{name}.res" do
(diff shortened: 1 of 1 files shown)


chef/cookbooks/drbd/providers/resource.rb
```diff
@@ -31,6 +31,9 @@
   raise "Remote node #{remote_host} not found!" if remote_nodes.empty?
   remote = remote_nodes.first
 
+  ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
```
abel-navarro: is this getting a network range instead of a single IP?

jsuchome: No, it is single IP, check the usage in other recipes

vuntz: For the record, you can remove the `Chef::Recipe::` prefix

jsuchome:
I'm really confused by that gating error. It looks related, because it fails right in the database resources creation, but I do not see any drbd related errors...

rhafer:
The gating failure is unrelated to this change and it got already past the database deployment. So I think it's fine to merge this.

@aspiers or @vuntz could you please merge this. Seems I am lacking some github karma for that.

PR #155, opened 2016-10-25
Adapt the DRBD settings to the changes of cluster founder.

jsuchome:
During the upgrade, we do not want to reconfigure DRBD.
So we need to explicitely adapt the settings in node[:drbd] to
the changes of cluster founder, otherwise they would look as inconsistent
with the node setup.

This is what I want to prevent: https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/providers/drbd.rb#L56

UPDATE: likely obsoleted by second commit of https://github.com/crowbar/crowbar-ha/pull/152

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -69,6 +69,35 @@ module Api
       def repocheck
         Api::Node.repocheck(addon: "ha")
       end
+
+      # Adapt DRBD information to possible changes of cluster founder
+      def adapt_drbd_settings(founder_name)
+        founder = NodeObject.find_node_by_name(founder_name)
+
+        # nothing to be done if there's no DRBD setup
+        return true unless founder[:drbd]
+
+        modified = false
+        founder[:drbd][:rsc].each do |_, resource|
+          unless resource[:master]
+            resource[:master] = true
+            modified = true
+          end
+        end
+        return true unless modified
+
+        founder.save
+
+        # adapt the info for remaining node
+        cluster_env = founder[:pacemaker][:config][:environment]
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -69,6 +69,35 @@ def health_report
       def repocheck
         Api::Node.repocheck(addon: "ha")
       end
+
+      # Adapt DRBD information to possible changes of cluster founder
+      def adapt_drbd_settings(founder_name)
```
rhafer: AFAICS this message will only ever called after the founder was changed using ``` set_node_as_founder``` from https://github.com/crowbar/crowbar-ha/pull/152. Likely even from within that function, right?

Wouldn't it be a lot easier (and probably a little more efficient) to pass whole node objects into this function (we load the nodes already in ```set_node_founder```). It would also save yet another ```node.save``` call.

As the whole purpose of this function is to adjust  the ```master``` attribute of the drbd resources after the founder node change I think the name is also somewhat missleading.

jsuchome: Yes, I've thought about it. I wanted to make it universally usable, that's why I chose this approach.

But I could call it directly from `set_node_as_founder` from Api::Pacemaker with node object.

jsuchome: @rhafer so I've adapted https://github.com/crowbar/crowbar-ha/pull/152 That would change this PR unnecessary

aspiers: I just merged #152 so we can close this now?

PR #156, opened 2016-10-26
[stable/3.0] fix corosync.conf values (bsc#1001164)

aspiers:
Our corosync.conf.v2 template was based on a version from SLE HA which
had some tunings accidentally dropped in the migration from SLE11.
In particular, some of the timeouts were too aggressive.

(cherry picked from commit c1d9c8f4e25e74a4cb5413b9f3e6337c7e1ff672)


rsalevsky:
Gating and travis failed.

vuntz:
(btw, the new gating passed, but old gating was using a different name, so didn't get overwritten)

PR #158, opened 2016-11-28
[UNTESTED] don't restart corosync when corosync.conf changes (bsc#1001164)

aspiers:
If the config parameters are changed, it's too risky to just restart the cluster - this could happen on all cluster nodes at a similar time and cause a significant outage.  Fortunately it's possible to instead reload the config whilst keeping `corosync` running, via `corosync-cfgtool -R`.


aspiers:
@matelakat Did you test this one or did you backport and test with SOC6?

matelakat:
superseded by https://github.com/crowbar/crowbar-ha/pull/160

PR #159, opened 2016-11-29
don't restart corosync when corosync.conf changes (bsc#1001164)

matelakat:

### DIFF
--- a/chef/cookbooks/corosync/recipes/config.rb
+++ b/chef/cookbooks/corosync/recipes/config.rb
@@ -56,8 +56,17 @@ template "/etc/corosync/corosync.conf" do
     transport: node[:corosync][:transport]
   )
 
-  service_name = node[:pacemaker][:platform][:service_name] rescue nil
-  if service_name
-    notifies :restart, "service[#{service_name}]"
-  end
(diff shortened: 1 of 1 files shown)


matelakat:
[Backport corosync config reload bug to 3.0](https://trello.com/c/ScT5hFYp/199-backport-corosync-config-reload-bug-to-3-0)

matelakat:
This has been manually tested, and it is indeed reloading the configuration nicely. The only issue we could have with this is the following:
- Let's assume we would like to change `corosync.conf` from `A` to `B`:
- node1, node2, node3 all has config `A`
- chef kicks off, finishes first on node1 - this means it's triggering a reload, it would mean that at that point the cluster's configuration would be `B`, `A`, `A`, because the command triggers a config reload on all the nodes.

So the issues:
- config will be reload `n` times, where `n` is the number of nodes in the cluster
- inconsistent configurations might present

Best would be to trigger a reload only once per cluster.

@aspiers - ideas?

matelakat:
Looking at the CI it seems that reload corosync.conf can fail - I guess corosync is not installed at that time yet. This might cause an issue during deployment.

matelakat:
superseded by #161

PR #160, opened 2016-12-02
don't restart corosync when corosync.conf changes (bsc#1001164)

matelakat:

### DIFF
--- a/chef/cookbooks/corosync/recipes/config.rb
+++ b/chef/cookbooks/corosync/recipes/config.rb
@@ -56,8 +56,27 @@ template "/etc/corosync/corosync.conf" do
     transport: node[:corosync][:transport]
   )
 
-  service_name = node[:pacemaker][:platform][:service_name] rescue nil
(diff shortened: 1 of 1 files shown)


matelakat:
superseeds #158

chef/cookbooks/corosync/recipes/config.rb
```diff
@@ -56,8 +56,27 @@
...
+  # If the config parameters are changed, it's too risky to just
+  # restart the cluster - this could happen on all cluster nodes at a
+  # similar time and cause a significant outage.  Fortunately it's
+  # possible to instead reload the config whilst keeping corosync running,
+  # via corosync-cfgtool -R.
+end
+
+execute "reload corosync.conf" do
+  # corosync-cfgtool -R reloads the configuration across ALL the nodes in the
+  # cluster - each reloading it's own local configuration file. This means that
```
aspiers: "its" ;-)

PR #161, opened 2016-12-02
[3.0] don't restart corosync when corosync.conf changes (bsc#1001164)

matelakat:

(cherry picked from commit d569623f863da198753cdc81cb339a028b09e62c)

backport of #160

### DIFF
--- a/chef/cookbooks/corosync/recipes/config.rb
+++ b/chef/cookbooks/corosync/recipes/config.rb
@@ -56,8 +56,27 @@ template "/etc/corosync/corosync.conf" do
     transport: node[:corosync][:transport]
   )
 
-  service_name = node[:pacemaker][:platform][:service_name] rescue nil
-  if service_name
-    notifies :restart, "service[#{service_name}]"
-  end
+  # If the config parameters are changed, it's too risky to just
+  # restart the cluster - this could happen on all cluster nodes at a
+  # similar time and cause a significant outage.  Fortunately it's
+  # possible to instead reload the config whilst keeping corosync running,
+  # via corosync-cfgtool -R.
+end
+
+execute "reload corosync.conf" do
+  # corosync-cfgtool -R reloads the configuration across ALL the nodes in the
+  # cluster - each reloading its own local configuration file. This means that
+  # running this command on a cluster with n nodes results in n^2 reloads.
+  # Given that the reloads might happen while some of the nodes still having
+  # the old configuration, this approach is not valid for updating nodelist
+  # or votequorum parameters, but should work for amending token timeouts.
+  # Ideally we would synchronize between all nodes after the config files are
+  # updated, and then just run the reload on a single node, but we can only
+  # achieve that when the proposal is being applied, and we need something
+  # which works with the convergence runs.
+  command "corosync-cfgtool -R"
+  user "root"
+  group "root"
+  action :nothing
+  subscribes :run, "template[/etc/corosync/corosync.conf]", :delayed
 end


matelakat:
This is a backport of #160

matelakat:
[Backport corosync config reload bug to 3.0](https://trello.com/c/ScT5hFYp/199-backport-corosync-config-reload-bug-to-3-0)

aspiers:
Travis failures :-/

Itxaka:
`Gem::InstallError: rack requires Ruby version >= 2.2.2.`

Seems that we need to pin rack to 1.6.5

Itxaka:
https://github.com/crowbar/crowbar-ha/pull/162 should fix this

chef/cookbooks/corosync/recipes/config.rb
```diff
@@ -56,8 +56,17 @@
...
-  service_name = node[:pacemaker][:platform][:service_name] rescue nil
-  if service_name
-    notifies :restart, "service[#{service_name}]"
-  end
+  # If the config parameters are changed, it's too risky to just
+  # restart the cluster - this could happen on all cluster nodes at a
+  # similar time and cause a significant outage.  Fortunately it's
+  # possible to instead reload the config whilst keeping corosync running,
+  # via corosync-cfgtool -R.
+end
```
aspiers: Please can you add our discoveries from the conversation with @gao-yan to this comment?

- Running `corosync-cfgtool -R` actually causes a config reload on *all* cluster nodes, not just the local host.  So the reload will happen O(n^2) times, but we think that's generally OK for the following reasons ...
- Each node will reload the config from its own `corosync.conf`.
- This PR doesn't implement any cluster-wide orchestration of the reload, so if left to Chef convergence rather than re-applying the Pacemaker proposal, each node will run this command at some random point within a (typically) 15-minute window.
- When each node runs this command, it is expected that all the corosync.conf files match.  However as long as the config changes don't touch things like `nodelist` and `votequorum` then it should be OK; discrepancies between nodes in (say) the corosync token timeout should not cause problems.

Ideally we would synchronize between all nodes after the config files are updated, and then just run the reload on a single node, but we can only achieve that when the proposal is being applied, and we need something which works with the convergence runs.

matelakat: pincking it up...

matelakat:
@aspiers Are we good with the additional changes? Also a question from me: Let's say we install a cluster with several nodes. If I understand it correctly during installation, after each corosync config file has been updated, we'll reload the config. Fine. What happens if for some reason Node(1)..Node(n-1) has already joined the cluster, but for some reason chef hasn't installed corosync on Node(n) yet, and let's say that this situation applies long enough for the cluster to fence Node(n) - is it a possible scenario, or am I getting it wrong?

chef/cookbooks/corosync/recipes/config.rb
aspiers: Thanks, looks good now (except for the "it's" typo which is not a dealbreaker ;-)

matelakat:
> @aspiers Are we good with the additional changes? Also a question from me: Let's say we install a cluster with several nodes. If I understand it correctly during installation, after each corosync config file has been updated, we'll reload the config. Fine. What happens if for some reason Node(1)..Node(n-1) has already joined the cluster, but for some reason chef hasn't installed corosync on Node(n) yet, and let's say that this situation applies long enough for the cluster to fence Node(n) - is it a possible scenario, or am I getting it wrong?

We discussed this, and it won't be a problem, as the node which has not joined the cluster yet, cannot be fenced.

PR #163, opened 2016-12-06
Do the translation of crowbar-specific STONITH config in the rails app

vuntz:
Doing it in the recipes is an issue: as they're default attributes, it means every time we need to look at these attributes, we need to run the recipes. And we will want to not run these recipes all the time in the future. For instance, https://github.com/crowbar/crowbar-core/pull/747

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
@@ -46,125 +46,19 @@ end
 case node[:pacemaker][:stonith][:mode]
 when "sbd"
   include_recipe "crowbar-pacemaker::sbd"
-
-# Need to add the hostlist param for shared
 when "shared"
-  params = node[:pacemaker][:stonith][:shared][:params]
-
-  if params.respond_to?("to_hash")
-    params = params.to_hash
-  elsif params.is_a?(String)
-    params = ::Pacemaker::Resource.extract_hash(" params #{params}", "params")
-  else
-    message = "Unknown format for STONITH shared parameters: #{params.inspect}."
-    Chef::Log.fatal(message)
-    raise message
-  end
-
-  all_nodes = CrowbarPacemakerHelper.cluster_nodes(node) + \
-    CrowbarPacemakerHelper.remote_nodes(node)
-
-  member_names = all_nodes.map { |n| pacemaker_node_name(n) }
-  params["hostlist"] = member_names.join(" ")
-
-  node.default[:pacemaker][:stonith][:shared][:params] = params
-
-# Crowbar is using FQDN, but pacemaker seems to only know about the hostname
-# without the domain (and hostnames for remote nodes are not real "hostnames",
-# but primitive names), so we need to translate this here
+  # Nothing to do, translation done in the rails app
 when "per_node"
-  nodes = node.default[:pacemaker][:stonith][:per_node][:nodes]
-  new_nodes = {}
-  domain = node[:domain]
-
-  all_nodes = CrowbarPacemakerHelper.cluster_nodes(node) + \
-    CrowbarPacemakerHelper.remote_nodes(node)
-
-  nodes.keys.each do |fqdn|
-    cluster_node = all_nodes.find { |n| fqdn == n[:fqdn] }
-    next if cluster_node.nil?
-
-    stonith_node_name = pacemaker_node_name(cluster_node)
-    new_nodes[stonith_node_name] = nodes[fqdn].to_hash
-  end
-
-  node.default[:pacemaker][:stonith][:per_node][:nodes] = new_nodes
-
-# Translate IPMI stonith mode from the barclamp into something that can be
-# understood from the pacemaker cookbook
+  # Nothing to do, translation done in the rails app
 when "ipmi_barclamp"
-  node.default[:pacemaker][:stonith][:mode] = "per_node"
-  node.default[:pacemaker][:stonith][:per_node][:agent] = "external/ipmi"
-  node.default[:pacemaker][:stonith][:per_node][:nodes] = {}
-
-  all_nodes = CrowbarPacemakerHelper.cluster_nodes(node) + \
-    CrowbarPacemakerHelper.remote_nodes(node)
-
-  all_nodes.each do |cluster_node|
-    unless cluster_node.key?("ipmi") && cluster_node[:ipmi][:bmc_enable]
-      message = "Node #{cluster_node[:hostname]} has no IPMI configuration from IPMI barclamp; another STONITH method must be used."
-      Chef::Log.fatal(message)
-      raise message
-    end
-
-    stonith_node_name = pacemaker_node_name(cluster_node)
-    bmc_net = Barclamp::Inventory.get_network_by_type(cluster_node, "bmc")
-
-    params = {}
-    params["hostname"] = stonith_node_name
-    params["ipaddr"] = bmc_net.address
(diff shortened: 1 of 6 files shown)


vuntz:
I just realized we probably need a revision bump to do the changes in a migration too.

vuntz:
@aspiers for reference, that's the bit that was missing for fixing the compute-ha job with https://github.com/crowbar/crowbar-core/pull/747

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -681,6 +681,9 @@ def prepare_stonith_attributes(role_attributes, remote_nodes, member_nodes, remo
     cluster_nodes = member_nodes + remote_nodes
     stonith_attributes = role_attributes["stonith"]
 
+    # still make the original mode available
```
aspiers: I don't get this commit at all.  Please can you expand the commit message to give some context?

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -54,7 +54,7 @@
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 32,
+      "schema-revision": 100,
```
aspiers: Probably a stupid question but why 100?

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -676,20 +677,9 @@ def validate_proposal_after_save proposal
...
-    remotes ||= []
-    if remotes.include?(node.name)
-      "remote-#{node["hostname"]}"
-    else
-      node["hostname"]
-    end
-  end
-
-  def prepare_stonith_attributes(role, remote_nodes, member_nodes, remotes, members)
+  def prepare_stonith_attributes(role_attributes, remote_nodes, member_nodes, remotes, members)
```
aspiers: OK, but IIUC ideally this method signature refactoring would have been done in a separate commit preceding the one introducing the migration, to keep the commits as readable as possible.

chef/data_bags/crowbar/template-pacemaker.json
vuntz: We agreed to bump to 100 for all barclamps for the 4.0 branch a while ago.

chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
```diff
@@ -47,125 +47,23 @@ def pacemaker_node_name(n)
 when "sbd"
   include_recipe "crowbar-pacemaker::sbd"
 
-# Need to add the hostlist param for shared
+# Nothing to do, translation done in the rails app
```
vuntz: done

crowbar_framework/app/models/pacemaker_service.rb
vuntz: I expanded the message. Hopefully it's clearer?

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -668,5 +671,97 @@ def validate_proposal_after_save proposal
...
     super
   end
-end
 
+  private
+
+  def pacemaker_node_name(node, remotes)
+    remotes ||= []
+    if remotes.include?(node.name)
+      "remote-#{node["hostname"]}"
```
vuntz: AFAIK, it's not duplicated in the rails app. There's a similar method in the cookbook, though.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -668,5 +671,97 @@ def validate_proposal_after_save proposal
...
+    else
+      node["hostname"]
+    end
+  end
+
+  def prepare_stonith_attributes(role, remote_nodes, member_nodes, remotes, members)
```
vuntz: I guess we don't have the same definition of "very long" ;-) We can create helper methods for each mode, but I feel it makes the whole thing less readable. How much do you care about it?

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Done.

vuntz:
Gating needs https://github.com/SUSE-Cloud/automation/pull/1571 to be green

crowbar_framework/app/models/pacemaker_service.rb
aspiers: What?! 90 lines is *disgustingly* long for a method.  Anything greater than one page (i.e. 25 lines) is already bad and has a significant negative impact on legibility, data scoping, testability, modular design etc. etc.  [This website suggests asking questions when methods grow longer than *10* lines.](https://sourcemaking.com/refactoring/smells/long-method)

IMNSHO long methods and bad naming are the top two cardinal sins of programmers.  At least 80% of code I read could be significantly improved by keeping methods short and naming things well.  It baffles me that so often I see even really experienced programmers not doing this.  And my "O" is Not So "H" because it's not really my opinion - it comes from programmers who are WAY cleverer and more experienced than me, e.g. https://youtu.be/8bZh5LMaSmE?t=34s (*strongly* recommend watching the whole video if you haven't seen it already).

So as you can see, you struck a nerve, and I care about this a lot :-)

In this particular case, I do concede that the method mainly consisting of a `case` structure makes the length less bad that a 90-line method would normally be.  But there would be several immediate benefits to splitting it up nevertheless:

- clarify which variables are needed within the `when` bodies (for example, `role` isn't)
- make unit testing of each case cleaner
- make the list of cases immediately visible on one screen
- discourage programmers from growing a huge method even further (this is a really important one, just look at the monstrosity which is `ServiceObject#apply_role`)
- (long-term) open up the possibility for refactoring according to the Open/Closed Principle using a dispatcher and subclasses for each STONITH type, and in fact moving to this from a huge conditional is almost exactly the same redesign shown in the youtube talk linked above!

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Ah OK, yes thanks, this is much clearer now!  However the name `crowbar_mode` is pretty vague, since everything involved here is Crowbar-related.  How about `ui_mode` or `proposal_mode` or similar?

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -667,5 +692,105 @@ def validate_proposal_after_save proposal
...
+
+      hypervisor_ip = stonith_attributes["libvirt"]["hypervisor_ip"]
+      hypervisor_uri = "qemu+tcp://#{hypervisor_ip}/system"
+
+      cluster_nodes.each do |cluster_node|
+        stonith_node_name = pacemaker_node_name(cluster_node, remotes)
+
+        # We need to know the domain to interact with for each cluster member; it
+        # turns out the domain UUID is accessible via ohai
+        domain_id = cluster_node["crowbar_ohai"]["libvirt"]["guest_uuid"]
```
jsuchome: I've just hit a problem when my migration failed because `cluster_node["crowbar_ohai"]` does not contain `libvirt` key. Not sure if my setup was so broken or if we should adapt this code for such eventuality.

dirkmueller: That means the ohai plugins weren't loaded. This sounds like  a bug elsewhere

jsuchome: At which point should they be run? This seems to be easily reproducible during the upgrade, right after admin server is upgraded and when crowbar-init is trying to run migrations for new crowbar

PR #164, opened 2016-12-06
Do not require crowbar-pacemaker cookbook to run when we configure HA in openstack

vuntz:
We make it possible to not run the crowbar-pacemaker cookbook when we run some OpenStack cookbook that use HA and require some specific setup (for drbd or haproxy, for instance).

This depends on and needs to be merged at the same time as: https://github.com/crowbar/crowbar-openstack/pull/635

Note: this includes a patch similar to https://github.com/crowbar/crowbar-ha/pull/116

PR #165, opened 2016-12-07
crowbar-pacemaker: Do not start or restart drbd service (bsc#971771)

vuntz:
It should only be started by pacemaker (to avoid ownership conflict with
systemd), and therefore we should not require drbd to be running to do
the setup.

This is a followup of 52d11b7 and is required to properly fix
https://bugzilla.suse.com/show_bug.cgi?id=971771

(the three other commits are required to keep things working)

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/drbd_setup.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/drbd_setup.rb
@@ -64,8 +64,6 @@ include_recipe "drbd::install"
 include_recipe "drbd::config"
 
 # Disable drbd as it should never be started on boot (pacemaker should do it)
-# However, start it because on initial setup, we currently require drbd to be
-# running.
 service "drbd" do
-  action [:disable, :start]
+  action :disable
 end
--- a/chef/cookbooks/drbd/providers/resource.rb
+++ b/chef/cookbooks/drbd/providers/resource.rb
@@ -61,6 +61,12 @@ action :create do
   end
   p.run_action(:run)
 
+  drbdadm_up = execute "drbdadm up #{name}" do
+    only_if { drbd_resource_template.updated_by_last_action? }
+    action :nothing
+  end
+  drbdadm_up.run_action(:run)
+
   # claim primary based off of master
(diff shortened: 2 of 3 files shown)


vuntz:
This doesn't work because we use drbd-overview to check the status of the drbd resource, but of course, it doesn't return anything when drbd is not running... So we need more rework here.

aspiers:
Is this related to #116?

vuntz:
> Is this related to #116?

Yes. Sorry, I had the initial commit in https://github.com/crowbar/crowbar-ha/pull/164 and mentioned that PR there, and then decided to move the drbd bits to its proper PR but failed to mention this one...

vuntz:
@aspiers this is rather small, and I guess we really want this, so a review would be great :-)

vuntz:
> The code changes look fine, but the commit messages and the empty PR description are lacking an explanation for why these changes are necessary.

I agree with the PR description, but I think it's unfair to say that the commit messages don't explain why: the title is always the "what" and there's always a "why" message in the body.

> I think the real motivation for the changes is to fix bsc#971771 (soft shutdown of HA controller node results in premature DRBD shutdown), so at very least the first commit (and PR description) should reference that. This was in the original PR #116 but got lost during the rewrite.

This was actually merged as https://github.com/crowbar/crowbar-ha/commit/52d11b7d94cc0f0eb14be3462b7cd2d6814e41ac ; but I agree that the bug can be referenced again; let me do that.

vuntz:
Force-pushed with new commit messages but no code change. It's a shame that this re-triggers the CI...

aspiers:
> I agree with the PR description, but I think it's unfair to say that the commit messages don't explain why: the title is always the "what" and there's always a "why" message in the body.

The first commit (the most important one) was missing a "why", but it's good now :-)  Thanks.

> Force-pushed with new commit messages but no code change. It's a shame that this re-triggers the CI...

We should be able to fix this by tweaking `github-status.rb`.  IIRC I had a conversation with github about this quite a while ago.  In this case I think it's OK to merge before waiting for gating, if you want.

aspiers:
> IIRC I had a conversation with github about this quite a while ago.

I found the email and submitted isaacs/github#855 as a result.

PR #168, opened 2016-12-09
upgrade: Look for one more upgrade step when checking if node is being upgraded

jsuchome:
New "done_os_upgrade" step identifies the situation when node
has been upgraded, but we still want to skip synchronization,
because other nodes in cluster may still have the old system.

It's introduced by https://github.com/crowbar/crowbar-core/pull/840


jsuchome:
> I guess this needs to be merged before crowbar/crowbar-core#840 ?

Yes

PR #169, opened 2016-12-19
Decrease amount of node saves

vuntz:
Saving is not cheap, and most of the time, it's not needed (as no attribute changed, or because we will already save later).

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/mutual_ssh.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/mutual_ssh.rb
@@ -22,27 +22,26 @@
 # save them.  This is required so that hb_report can gather logs from
 # all cluster members.
 
-access_keys = {}
-
-node["provisioner"]["access_keys"].strip.split("\n").each do |key|
-  key.strip!
-  unless key.empty?
-    nodename = key.split(" ")[2]
-    access_keys[nodename] = key
-  end
-end
+cluster_access_keys = []
 
 CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
   pkey = nil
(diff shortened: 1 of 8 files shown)


nicolasbock:
There seems to be another race condition during the pacemaker setup. Or something is failing :(

chef/cookbooks/crowbar-pacemaker/recipes/mutual_ssh.rb
```diff
@@ -22,27 +22,26 @@
...
+cluster_access_keys = []
 
 CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
   pkey = nil
   if cluster_node[:crowbar][:ssh]
     pkey = cluster_node[:crowbar][:ssh][:root_pub_key]
   end
-  if !pkey.nil? && cluster_node.name != node.name && !access_keys.values.include?(pkey)
-    access_keys[cluster_node.name] = pkey
+  if !pkey.nil? && cluster_node.name != node.name && !access_keys.include?(pkey)
```
Itxaka: `access_keys` does not exists anymore, is now `cluster_access_keys`

Itxaka: Maybe this is the issue that breaks the CI job `00:29:53.340 Cannot find a resource for access_keys on suse version 12.2`

nicolasbock:
@Itxaka nice catch! That explains the gating failure.

nicolasbock:
The tempest failure in `TestVolumeBootPatternV2` is failing in all HA builds right now. @aplanas and I are looking into it.

vuntz:
Was completed in https://github.com/crowbar/crowbar-ha/pull/210

PR #170, opened 2016-12-21
Reimplement sync marks without node searches

vuntz:
This has several benefits, as highlighted in the commit messages.

Submitting now for early reviews, but this is not tested nor ready.


chef/cookbooks/crowbar-pacemaker/libraries/cibattribute.rb
```diff
@@ -0,0 +1,114 @@
...
+  def self.get(node, attribute, default)
+    # This is what we want:
+    # $ crm_attribute --name=crowbar_sync-pacemaker_setup --node=d52-54-77-77-77-2 --query --default=0
+    # Could not map name=d52-54-77-77-77-2 to a UUID
+    # $ crm_attribute --name=crowbar_sync-pacemaker_setup --node=d52-54-77-77-77-02 --query --default=0
+    # scope=nodes  name=crowbar_sync-pacemaker_setup value=0
+    validate_node(node)
+    validate_attribute(attribute)
+    validate_value(default)
+    output = `crm_attribute --name=#{attribute} --node=#{node} --query --default=#{default} 2>&1`
```
abel-navarro: I'm a bit concerned about using `crm_attribute` here. If one node is temporarily not participating in the HA cluster (i.e. `crm node standby` or the node is fenced) then this call will return an error. The former approach doesn't depend on the availability of the cluster nodes, so in that sense it's more resilient.

I'd love to hear @aspiers opinion about this.

aspiers:
Only just noticed this.  I wonder why I wasn't requested to review before.

vuntz:
Obsoleted by https://github.com/crowbar/crowbar-ha/pull/217

PR #175, opened 2017-01-18
Fallback to dmi[:system][:uuid] if new ohai attributes are not available

jsuchome:
Using new ohai attribute was introduced in commit
3b4cb98502eaa07543bc401ec6ea99e3959549ab.

During 6-7 upgrade, we do not have these new attributes yet when we
are running migrations.


jsuchome:
This is an alternative to https://github.com/crowbar/crowbar-core/pull/1002

jsuchome:
Closing, as the alternative approach was used.

PR #176, opened 2017-01-20
[stable/3.0] Disable apache2 service when deployed with HA

jsuchome:
This is an attempt to fix https://bugzilla.suse.com/show_bug.cgi?id=1021106. Comments are welcomed.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
@@ -101,3 +101,8 @@ end
 # Override service provider for apache2 resource defined in apache2 cookbook
 resource = resources(service: "apache2")
 resource.provider(Chef::Provider::CrowbarPacemakerService)
+
+# Disable apache2 as it should be only managed by pacemaker now
+service "apache2" do
+  action :disable
+end


jsuchome:
[apache service not configured correctly with HA (not handled by pacemaker)](https://trello.com/c/1b5iNWF7/273-apache-service-not-configured-correctly-with-ha-not-handled-by-pacemaker)

chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
```diff
@@ -101,3 +101,8 @@
 # Override service provider for apache2 resource defined in apache2 cookbook
 resource = resources(service: "apache2")
 resource.provider(Chef::Provider::CrowbarPacemakerService)
+
+# Disable apache2 as it should be only managed by pacemaker now
+service "apache2" do
+  action :disable
+end
```
vuntz: This is unfortunately too fragile, as something could come later on in the run-list that would re-enable it :/

vuntz:
That being said, we have this code already:

```ruby
# Override service provider for apache2 resource defined in apache2 cookbook
resource = resources(service: "apache2")
resource.provider(Chef::Provider::CrowbarPacemakerService)
```
This should do all the magic that is required for us, as any call to `:enable` will effectively disable the service (because that's what the provider is doing).

So something may not be working here?

jsuchome:
``resource = resources(service: "apache2")``

I think the point is that it was already enabled before this code

chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
jsuchome: But isn't this what we're doing for drbd?

vuntz:
> I think the point is that it was already enabled before this code

In the compile phase? (could be, in which case we need the attribute like in your other PR)

vuntz:
I believe this is the right fix: https://github.com/crowbar/crowbar-core/pull/1018 (works for me)

PR #177, opened 2017-01-26
upgrade: Do not ignore founder change when the node is founder

jsuchome:
It is possible that (e.g. due to some previous error) node is marked
as founder but drbd master value is not correctly set.

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -79,7 +79,6 @@ module Api
         end
         if new_founder[:pacemaker][:founder]
           Rails.logger.debug("Node #{name} is already the cluster founder.")
-          return true
         end
 
         # 2. find the current cluster founder in the same cluster


crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -79,7 +79,6 @@ def set_node_as_founder(name)
         end
         if new_founder[:pacemaker][:founder]
           Rails.logger.debug("Node #{name} is already the cluster founder.")
-          return true
```
rhafer: I am a bit confused. If ```new_founder[:pacemaker][:founder]``` is already the founder. Won't it mean, that after the search on line 85 ```old_founder``` points to the same node as ```new_founder``` ?

If that is true, couldn't we just skip the whole block from line 85 to 100 in that case (and save us a chef search and a node.save) . E.g. by putting it into an ```else``` branch of the ```if new_founder[:pacemaker][:founder]```

jsuchome: But it's possible that the drbd data are incorrectly set for both of nodes (it happened to me). In that case we should still try to write them to both nodes.

Which means, we should change the search for `old_founder` to `other_node` and rather do something like
```
other_node = NodeObject.find(
   "pacemaker_config_environment:#{cluster_env} AND NOT name:#{new_founder.name}"
)
```

I think

rhafer: Yeah, then it makes sense again.

jsuchome:
Seems like this is not really needed

PR #179, opened 2017-02-08
Rework haproxy config for stickiness and balance strategy

vuntz:
- We enable stickiness based on SSL sessions
- We add the ability to do stickiness based on cookies
- We allow to define the balance strategy on a per-resource basis
- We stop overriding the default balance strategy (and therefore use roundrobin by default)


vuntz:
> Speaking out of memory, I had trouble having complex configuration in a listen section in haproxy config file.

This works fine here. Just need to validate the SSL case, but everything else seems good.

vuntz:
So overall, this looks good, except for horizon with ssl: there, the browser is creating multiple connections, so the ssl session id affinity doesn't work as expected. Easy workaround is to use the source balance strategy in that case. But I'll play a bit more with it.

sjamgade:
[One change](https://github.com/crowbar/crowbar-ha/pull/179/commits/10ee491ed9b4a7afa2bc2ad8ab81c90dce81c6b7) has been cherry-picked in  https://github.com/crowbar/crowbar-ha/pull/193, do consider reviewing that one.

matelakat:
re-running a build on a separate PR to see if ssl is still broken.

matelakat:
Given that the tests passed in #208 - including the SSL tests, I am merging this PR.

matelakat:
Ah, I cannot merge it in the lack of passing CI.

matelakat:
@nicolasbock @dirkmueller could you please transfer your +1 to #208 so we can get this change merged please?

PR #183, opened 2017-02-09
pacemaker: move set_maintenance_mode into pacemaker helper

Itxaka:
Useful to expose it to other cookbooks

Use case: crowbar/crowbar-core#563

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/maintenance_mode_helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/maintenance_mode_helpers.rb
@@ -56,5 +56,20 @@ module CrowbarPacemaker
       # to an empty Chef::Node::Attribute if not already set to true.
       node.default[:pacemaker][:maintenance_mode][$$][:via_chef] == true
     end
+
+    def set_maintenance_mode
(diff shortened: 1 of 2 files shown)


Itxaka:
@aspiers as commented by you on https://github.com/crowbar/crowbar-core/pull/563#discussion_r92816069 I was duplicating this code to be used on the updater cookbook as it was not publicy available. With this we can reuse that code and hopefully move further the other PR :D

aspiers:
Yes thanks for the reminder - I had totally forgotten I wrote that comment, but coincidentally I stumbled across it just at the same time you were writing the explanation :)

aspiers:
BTW that (i.e. my poor memory, but other reasons too) is why it's good practice to mention dependencies between two PRs in both PR's description field ;-)  I have updated both PR descriptions to include this now.

Itxaka:
Actually is linked to the other PR but github does not really highlight that really well and gets kind of hidden

aspiers:
@Itxaka I just realised that this PR is not sufficient for that use case, because it will not work with remote nodes.  However this PR is a good first start, so I'm still +1 on it.  I'm also updating crowbar/crowbar-core#563 accordingly.

PR #185, opened 2017-03-13
upgrade: Provide a better error

Itxaka:
linked to https://github.com/crowbar/crowbar-core/pull/1155
this provides an error message that syncs the node with the actual error
in order to show it properly to the user on the error message

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -53,12 +53,14 @@ module Api
         founders.each do |n|
           ssh_retval = n.run_ssh_cmd("crm status 2>&1")
           if (ssh_retval[:exit_code]).nonzero?
-            crm_failures[n.name] = ssh_retval[:stdout]
+            crm_failures[n.name] = "#{n.name}: #{ssh_retval[:stdout]}"
+            crm_failures[n.name] << " #{ssh_retval[:stderr]}" unless ssh_retval[:stderr].blank?
             next
           end
           ssh_retval = n.run_ssh_cmd('crm status | grep -A 2 "^Failed Actions:"')
           if (ssh_retval[:exit_code]).zero?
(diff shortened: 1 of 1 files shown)


Itxaka:
[(2) S56P9 upgrade: Cleanup error messages](https://trello.com/c/k7Kg1D6Q/271-2-s56p9-upgrade-cleanup-error-messages)

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -53,12 +53,12 @@ def health_report
         founders.each do |n|
           ssh_retval = n.run_ssh_cmd("crm status 2>&1")
           if (ssh_retval[:exit_code]).nonzero?
-            crm_failures[n.name] = ssh_retval[:stdout]
+            crm_failures[n.name] = "#{n.name}: #{ssh_retval[:stdout]}"
```
skazi0: Please add stderr to the output as well.
Example when node is not reachable:
1. without stderr
zrzut ekranu z 2017-03-14 10-31-11
2. with stderr
zrzut ekranu z 2017-03-14 10-34-53

Itxaka: Nice, great catch @skazi0 :+1:

Itxaka: added stderr unless its empty

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -53,12 +53,14 @@ def health_report
...
+            crm_failures[n.name] << " #{ssh_retval[:stderr]}" unless ssh_retval[:stderr].empty?
```
jsuchome: Can't it be `nil`? Please make sure, I think it could be actually...

Itxaka: good, point, it could be nil. should be fixed now

PR #186, opened 2017-03-20
fix remote primitive existence test

aspiers:
**Still needs manual testing**

Fix a corner case where a remote primitive was incorrectly being reported as existing, due to `#crm_configure_show` returning a string like `node remote-d52-54-77-77-77-03:remote`, causing `#exists?` to return true.

This corner case arose during upgrade of a controller node, during `crowbar_join` after all primitives except for vips, drbd, and stonith resources were removed in anticipation of a switch to systemd agents.  At that point, the :update action of the remote primitive failed because it didn't know how to handle an object of type `node`, and this action happens before the containing transaction which would have recreated the primitive.

Since remote primitives use the `ocf:pacemaker:remote` agent, it probably makes sense to exclude them from this removal of primitives which happens during the upgrade, thereby eliminating this corner case.  However there may be other corner cases, so we should still ensure that `#exists?` always returns the correct value.

### DIFF
--- a/chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
+++ b/chef/cookbooks/pacemaker/libraries/pacemaker/cib_object.rb
@@ -34,10 +34,21 @@ module Pacemaker
           end
 
           if lines_for_def.empty?
-            # in case our attempt to only extract the definition we're
-            # interested in failed, play it safe and return the output
+            # In case our attempt to only extract the definition we're
+            # interested in failed, return nil to indicate the object
+            # doesn't exist.  This is particularly important in the
+            # corner case where an object like
+            #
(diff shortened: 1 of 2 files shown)


aspiers:
Hmm, need to fix the unit tests.

aspiers:
@jsuchome Travis fixed.

aspiers:
@jsuchome Forced an mkcloud rebuild since it was failing due to [`nova: Attribute ipaddress is not defined!` (Trello)](https://trello.com/c/vLiPktfn/315-nova-attribute-ipaddress-is-not-defined)

jsuchome:
Thanks. I've seen this error, I was not sure if it was already fixed

PR #188, opened 2017-03-22
drbd: Set node-id in resource config

cmurphy:
Some of the failing mkcloud HA tests (e.g. [1])) have the following chef
failure:

  Failure: (119) No valid meta-data signature found.

accompanied by an alert in the logs:

  ambiguous node id: meta-data: 0, config: 1

This patch sets the node IDs explicitly in the resource config file as
recommended in this mailing list post[2] and documented for drbd
9.0[3] to attempt to prevent these meta-data errors. It also adds a
connection stanza as that may also be affecting drbd's ability to
generate metadata[4]. Protocol C indicates "Writes to the DRBD device
complete as soon as they have reached the local and all remote disks"
and is the protocol used in the example configuration file in the
documentation. We could consider making this configurable later.

[1] https://ci.suse.de/job/openstack-mkcloud/56301/consoleText
[2] https://docs.linbit.com/doc/users-guide-90/re-drbdconf/
[3] http://lists.linbit.com/pipermail/drbd-dev/2017-January/003822.html
[4] http://lists.linbit.com/pipermail/drbd-dev/2017-January/003823.html

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -387,6 +387,13 @@ class PacemakerService < ServiceObject
     role.default_attributes["drbd"]["common"]["net"] ||= {}
     role.default_attributes["drbd"]["common"]["net"]["shared_secret"] = \
       role.default_attributes["pacemaker"]["drbd"]["shared_secret"]
+    # set node IDs for drbd metadata
+    member_nodes.each do |member_node|
+      member_node[:drbd] ||= {}
+      member_node[:drbd][:local_node_id] = member_node[:pacemaker][:founder] ? 0 : 1
+      member_node[:drbd][:remote_node_id] = member_node[:pacemaker][:founder] ? 1 : 0
+      member_node.save
+    end
 
     # translate crowbar-specific stonith methods to proper attributes
     prepare_stonith_attributes(role.default_attributes["pacemaker"],
--- a/chef/cookbooks/drbd/providers/resource.rb
+++ b/chef/cookbooks/drbd/providers/resource.rb
@@ -42,9 +42,11 @@ action :create do
       device: device,
       disk: disk,
       local_hostname: node.hostname,
+      local_node_id: node[:drbd][:local_node_id],
       local_ip: ip,
       port: port,
       remote_hostname: remote.hostname,
+      remote_node_id: node[:drbd][:remote_node_id],
       remote_ip: remote_ip
     )
     owner "root"
--- a/chef/cookbooks/drbd/templates/default/resource.erb
+++ b/chef/cookbooks/drbd/templates/default/resource.erb
@@ -5,9 +5,18 @@ resource <%= @resource %> {
   disk      <%= @disk %>;
   meta-disk internal;
   on <%= @local_hostname %> {
+    node-id   <%= @local_node_id %>;
     address   <%= @local_ip %>:<%= @port %>;
   }
   on <%= @remote_hostname %> {
+    node-id   <%= @remote_node_id %>;
     address   <%= @remote_ip %>:<%= @port %>;
+  }
+  connection {
+    host <%= @local_hostname %>  port <%= @port %>;
+    host <%= @remote_hostname %> port <%= @port %>;
+    net {
+      protocol C;
+    }
   }
 }


jsuchome:
During upgrade, we're creating new metadata (https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb#L39).
So I thin k it should not affect us

cmurphy:
Rebuilding with http://clouddata.cloud.suse.de/mkcloud/crowbar:crowbar-openstack:876:557b5953b03c2b60489a261de28910360032ea67:master/ added to UPDATEREPOS to see if the attribute error goes away

gao-yan:
And it's also mentioned in the chapter of "Migrating from DRBD 8 to DRBD 9" in our HA documentation:
http://docserv.nue.suse.com/documents/SLE-HA12-SP2/SLE-HA-guide/single-html/#sec.ha.drbd.migrate

"...
5. Recreate the metadata based on the new configuration
..."

cmurphy:
@s-t-e-v-e-n-k since you mentioned you would look into these gate failures in your morning I'll leave you with my notes. I think the gate failure on Sayali's original PR (https://github.com/crowbar/crowbar-openstack/pull/868) were related to DRBD. I'm attempting to run this PR co-gated with my resubmission of Sayali's (https://github.com/crowbar/crowbar-openstack/pull/876). As I'm leaving the office it looks like the crowbar-openstack PR managed to get past the database and rabbitmq proposals (https://ci.suse.de/job/openstack-mkcloud/57094/) which is where it usually breaks so I'm not convinced this crowbar-ha PR is needed. Interestingly, in my own mkcloud instance this crowbar-ha change fixes things for me without needing the crowbar-openstack change, so I'm not sure what's up with that. It seems to be the ceilometer proposal that breaks without the crowbar-openstack PR.

cmurphy:
Seems like my guess was wrong, this is causing different problems:

```
kernel: [  493.414873] drbd postgresql d52-54-77-77-01-02: Peer presented a node_id of 0 instead of 1
kernel: [  493.414888] drbd postgresql d52-54-77-77-01-02: conn( Connecting -> NetworkFailure )
kernel: [  493.439043] drbd postgresql d52-54-77-77-01-02: Connection closed
```

and the latest hacloud failures seem to be more related to the vip resources than drbd.

aspiers:
@cmurphy I have just realised that the change in this PR was wrong, which might explain the new problems you just described (sorry I didn't spot it yesterday when I reviewed it!).  The change is always setting the id of the *local* node (i.e. the one which is running the recipe via `chef-client`) to 0, and the remote (other) node to 1.  So both nodes think that they have id 0 and that the other one has id 1.  This is obviously wrong, and would perfectly explain why you are seeing `Peer presented a node_id of 0 instead of 1`!

So I suggest that you try this change again but first fix it so that the two config files are consistent with each other.  This would perhaps be best done by changing the Rails app to set `local-node-id` and `remote-node-id` attributes on each node, e.g. setting the founder node's `local-node-id` to `0` and the other to `1`, and vice-versa for `remote-node-id`.  An appropriate place to do this might be somewhere around https://github.com/crowbar/crowbar-ha/blob/master/crowbar_framework/app/models/pacemaker_service.rb#L384.  Then the recipe could simply pass these attributes to the template.

FYI a bit of history might be helpful here: the idea behind this approach is that we are trying to gradually move the smarts out of Chef cookbooks back into Crowbar.  Previously we might have done a [Chef node search](https://docs.chef.io/chef_search.html) in the recipe to calculate the local and remote node id for each node, but these days we are trying to eliminate use of Chef search so that we can ditch the Chef server entirely and only use Chef as a dumb per-node configuration manager.  So any code which relates to multiple nodes should be done in Crowbar.  This helps decouple Crowbar from Chef and could in theory let us ditch Chef altogether one day, although that's probably just a pipe dream ...

Having said all that, @vuntz may want to wade in here and suggest a better approach than mine :-)

cmurphy:
Thanks for the detailed explanation @aspiers. I'm happy to keep working on this but it seems like now our hacloud gate checks are failing for an entirely different reason, but they at least seem to make it past this drbd setup step, so I'm not sure that this change is needed.

cmurphy:
@aspiers I've updated this pull request with the changes you suggested. It's hard to say whether this is the answer to bsc#1003626 because the failures are intermittent but I think it at least doesn't hurt and it gets rid of the "ambiguous metadata" warnings in syslog.

dirkmueller:
does this change affect update (non-upgrade, e.g. maintenance update for customers installed with GM7)?

cmurphy:
@dirkmueller It shouldn't, because a drbd resource only ever gets created once, it won't be modified after that: https://github.com/crowbar/crowbar-ha/blob/08a97f42d50229c0d01dc3138fb76b92b217dfa7/chef/cookbooks/drbd/recipes/resource.rb#L24

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -387,6 +387,18 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
     role.default_attributes["drbd"]["common"]["net"] ||= {}
     role.default_attributes["drbd"]["common"]["net"]["shared_secret"] = \
       role.default_attributes["pacemaker"]["drbd"]["shared_secret"]
+    # set node IDs for drbd metadata
+    member_nodes.each do |member_node|
+      member_node[:pacemaker][:drbd] = {}
```
vuntz: Can you change this to:
```
member_node[:drbd] ||= {}
```
and then use:
```
member_node[:drbd][:local_node_id] = 0
```
?

Rationale: the drbd cookbook doesn't know anything about `:pacemaker` attributes.

vuntz:
You can consider me as +1 after the change I requested.

crowbar_framework/app/models/pacemaker_service.rb
aspiers: I agree.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -387,6 +387,18 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+      if member_node[:pacemaker][:founder]
```
aspiers: Just a style thing, but I'd write this as

    member_node[:pacemaker][:drbd][:local_node_id] = \
        member_node[:pacemaker][:founder] ? 0 : 1
    member_node[:pacemaker][:drbd][:remote_node_id] = \
        member_node[:pacemaker][:founder] ? 1 : 0

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -387,6 +387,13 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
     role.default_attributes["drbd"]["common"]["net"] ||= {}
     role.default_attributes["drbd"]["common"]["net"]["shared_secret"] = \
       role.default_attributes["pacemaker"]["drbd"]["shared_secret"]
+    # set node IDs for drbd metadata
+    member_nodes.each do |member_node|
+      member_node[:drbd] = {}
```
vuntz: Please use `||=` and not `=` to be on the safe side. Right now, you maybe losing attributes when re-applying and that can cause some really bad things.

cmurphy: Thanks for catching that

aspiers:
@vuntz I think you should be able to remove your -1 on this now.

PR #189, opened 2017-03-31
Write node's IP address into corosync.conf instead of subnet

jsuchome:
If there are more addresses in the subnet, corosync could get
confused and not find correct address to bind to.
So let's try to use node's IP instead (our corosync.conf files are node specific anyway)

See https://bugzilla.suse.com/show_bug.cgi?id=1030437 for context.

~~This is not tested.~~ UPDATE: tested already few times.

I know that in that bug mentioned above there's a patch for corosync proposed, but i was hoping to maybe have a simpler one on our side.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
@@ -26,6 +26,7 @@ if node[:corosync][:authkey].nil?
 end
 
 node[:corosync][:cluster_name] = CrowbarPacemakerHelper.cluster_name(node)
+node[:corosync][:bind_addr] = Barclamp::Inventory.get_network_by_type(node, "admin").address
 
 include_recipe "crowbar-pacemaker::quorum_policy"
 include_recipe "crowbar-pacemaker::stonith"
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -363,7 +363,6 @@ class PacemakerService < ServiceObject
     admin_net = Chef::DataBag.load("crowbar/admin_network") rescue nil
 
     role.default_attributes["corosync"] ||= {}
-    role.default_attributes["corosync"]["bind_addr"] = admin_net["network"]["subnet"]
 
     role.default_attributes["corosync"]["mcast_addr"] = role.default_attributes["pacemaker"]["corosync"]["mcast_addr"]
     role.default_attributes["corosync"]["mcast_port"] = role.default_attributes["pacemaker"]["corosync"]["mcast_port"]


chef/cookbooks/corosync/recipes/config.rb
```diff
@@ -38,6 +38,8 @@
   raise "Members have to be defined when using \"udpu\" transport!"
 end
 
+my_ipaddress = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
```
vuntz: You can remove `Chef::Recipe::` here.

vuntz: And actually, the biggest issue: this cookbook is not supposed to know anything about crowbar. The way you should deal with it is that in the crowbar-pacemaker cookbook, at the beginning of the default.rb recipe (I think), you should set the `bind_addr` attribute to this address.

This also means you can avoid setting the attribute in the pacemaker_service.rb file of the rails app.

jsuchome:
@vuntz like this?

Itxaka:
Applied this to https://ci.suse.de/job/openstack-mkcloud/58474/console (cloud-mkcloud7-job-upgrade-disruptive-ha-x86_64)

For the moment it seems to work, corosync is up and running with the proper bind_addr and the config was set properly:

```
root@crowbar:~ # knife node show d52-54-77-77-01-02.va1.cloud.suse.de -a corosync.bind_addr
corosync.bind_addr:  192.168.253.84
root@crowbar:~ # knife node show d52-54-77-77-01-01.va1.cloud.suse.de -a corosync.bind_addr
corosync.bind_addr:  192.168.253.81

```

jsuchome:
I've added the patch to this job https://ci.suse.de/job/openstack-mkcloud/58485/console and it seems it helped

jsuchome:
[(5) S58P6 Make Upgrade jobs green](https://trello.com/c/DJMzh2sT/392-5-s58p6-make-upgrade-jobs-green)

jsuchome:
[(5) S58P1 bsc#1030437 - [upgrade] Changing the pre-upgrade role for node failed](https://trello.com/c/c7DvdiFl/384-5-s58p1-bsc-1030437-upgrade-changing-the-pre-upgrade-role-for-node-failed)

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -26,6 +26,7 @@
 end
 
 node[:corosync][:cluster_name] = CrowbarPacemakerHelper.cluster_name(node)
+node[:corosync][:bind_addr] = Barclamp::Inventory.get_network_by_type(node, "admin").address
```
vuntz: Should be `node.set[:default][:bind_addr]`, but given that the line above has the same issue, let's ignore this.

aspiers: This aspect is being fixed in #190.

PR #190, opened 2017-04-03
Write node's address into corosync.conf instead of subnet IP

jsuchome:
If there are more addresses in the subnet, corosync could get
confused and not find correct address to bind to.
So let's try to use node's IP instead.

See bsc#1030437 for context.

(cherry picked from commit de1dfc6fc99570126c53128a3a66380ce9e59f5b)

Forward-port of https://github.com/crowbar/crowbar-ha/pull/189

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
@@ -26,6 +26,7 @@ if node[:corosync][:authkey].nil?
 end
 
 node[:corosync][:cluster_name] = CrowbarPacemakerHelper.cluster_name(node)
+node[:corosync][:bind_addr] = Barclamp::Inventory.get_network_by_type(node, "admin").address
 
 include_recipe "crowbar-pacemaker::quorum_policy"
 include_recipe "crowbar-pacemaker::stonith"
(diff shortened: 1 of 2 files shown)


Itxaka:
[(5) S58P1 bsc#1030437 - [upgrade] Changing the pre-upgrade role for node failed](https://trello.com/c/c7DvdiFl/384-5-s58p1-bsc-1030437-upgrade-changing-the-pre-upgrade-role-for-node-failed)

aspiers:
Actually, please can you also provide a commit message for the `node.set` explaining *why* the `node.set` is required?  Preferably something more explicit than "@vuntz [told me to do it](https://github.com/crowbar/crowbar-ha/pull/189#discussion_r109356340)" ;-)

jsuchome:
> Actually, please can you also provide a commit message for the node.set explaining why the node.set 
> is required? Preferably something more explicit than "@vuntz told me to do it" ;-)

I do not have anything better :-(

aspiers:
> I do not have anything better :-(

Well it's not good to make changes blindly, even if they come from @vuntz.  I think it's related to the attribute precedence, but let's check with him.

aspiers:
Actually looks like it's more of a style thing than a behavioural change: https://docs.chef.io/attributes.html#examples

jsuchome:
@aspiers do you still require change of commit message for the 2nd commit?
Or I could just remove that commit and make this pure forward-port.

aspiers:
@jsuchome We just agreed on IRC to use `Node#normal`, since `node[:foo] = 'bar'` will break on Chef 11, and `node.set` will break on Chef 14, if we ever reach it.  So all that's required here is to switch to `node.normal` and include this explanation of why in the commit message.

PR #198, opened 2017-04-28
haproxy: Add sync marks

cmurphy:
Without this patch, the keystone proposal often causes a crm resource
failure when it tries to start the haproxy service under pacemaker:

  Failed Actions:
  * haproxy_start_0 on d52-54-77-77-01-02 'not installed' (5): call=64, status=Not installed, exitreason='none',
      last-rc-change='Fri Apr 28 09:53:22 2017', queued=0ms, exec=92ms

When we examine the zypper history of both controller nodes, we can see
the haproxy package may be installed by 09:53:22 on one of the nodes,
but is only installed a few seconds later on the other node, leading
pacemaker to report a failure. This patch adds sync marks (modeled after
the apache recipe in the same cookbook) to ensure pacemaker does not try
to start haproxy until both nodes are ready.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
@@ -35,6 +35,18 @@ include_recipe "haproxy::setup"
 
 cluster_name = CrowbarPacemakerHelper.cluster_name(node)
 
+# Wait for all nodes to reach this point so we know that all nodes will have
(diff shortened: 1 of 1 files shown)


cmurphy:
@aplanas thanks - also needed to s/apache/haproxy/ :)

cmurphy:
@aspiers I think the apache errors are related to https://trello.com/c/P2kCVkiQ which is something different :(

PR #199, opened 2017-06-08
pacemaker: Use discovered BMC addr (bsc#1035215)

cmurphy:
In the IPMI barclamp, use_dhcp is set to true,
ignore_address_suggestions is irrelevant and the barclamp must use the
address discovered from the BMC interfaces. Since we're relying on dhcp
for this, we can't trust our database of allocated IPs to give us the
right address. This can cause the address for the STONITH device to be
either unset or incorrect, even when the IPMI barclamp reports the
correct BMC address. This patch looks up the use_dhcp value for the node
and, if set to true, uses the discovered address stored in the node
rather than the address allocated in the BMC network hash.

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -754,7 +754,12 @@ class PacemakerService < ServiceObject
 
         params = {}
         params["hostname"] = stonith_node_name
-        params["ipaddr"] = bmc_net["address"]
+        # If the bmc is using dhcp, we can't trust the crowbar bmc network to know it
+        params["ipaddr"] = if cluster_node["ipmi"]["use_dhcp"]
+          cluster_node["crowbar_wall"]["ipmi"]["address"]
+        else
+          bmc_net["address"]
+        end
         params["userid"] = cluster_node["ipmi"]["bmc_user"]
         params["passwd"] = cluster_node["ipmi"]["bmc_password"]
 


crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -754,7 +754,12 @@ def prepare_stonith_attributes(role_attributes, remote_nodes, member_nodes, remo
 
         params = {}
         params["hostname"] = stonith_node_name
-        params["ipaddr"] = bmc_net["address"]
+        # If the bmc is using dhcp, we can't trust the crowbar bmc network to know it
+        params["ipaddr"] = if cluster_node["ipmi"]["use_dhcp"]
+          cluster_node["crowbar_wall"]["ipmi"]["address"]
```
vuntz: Could we simply use this attribute all the time?

cmurphy: I think so, yes. If the BMC interface doesn't know its own address then there would be a whole different set of problems. I'll update.

cmurphy: Actually, I think not. Looking at https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/ipmi/recipes/ipmi-configure.rb when it sets the interface's IP address it doesn't update crowbar_wall, so it would not have that information later if the IP address wasn't found during discovery, or if ignore_address_suggestions was true an the IP address was updated, the information in crowbar_wall would be wrong.

cmurphy:
@vuntz do you agree with my reasoning about not using the crowbar_wall attribute all the time?

PR #200, opened 2017-06-12
[3.0] pacemaker: Use discovered BMC addr (bsc#1035215)

cmurphy:
In the IPMI barclamp, use_dhcp is set to true,
ignore_address_suggestions is irrelevant and the barclamp must use the
address discovered from the BMC interfaces. Since we're relying on dhcp
for this, we can't trust our database of allocated IPs to give us the
right address. This can cause the address for the STONITH device to be
either unset or incorrect, even when the IPMI barclamp reports the
correct BMC address. This patch looks up the use_dhcp value for the node
and, if set to true, uses the discovered address stored in the node
rather than the address allocated in the BMC network hash.

Note that this is not a direct cherry-pick, as the logic moved from the chef
recipe to the crowbar model between SOC6 and SOC7.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/stonith.rb
@@ -111,7 +111,13 @@ when "ipmi_barclamp"
 
     params = {}
     params["hostname"] = stonith_node_name
-    params["ipaddr"] = cluster_node[:crowbar][:network][:bmc][:address]
+    # If the bmc is using dhcp, we can't trust the crowbar bmc network to know it
+    params["ipaddr"] = if cluster_node[:ipmi][:use_dhcp]
+      cluster_node[:crowbar_wall][:ipmi][:address]
+    else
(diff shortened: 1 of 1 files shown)


cmurphy:
mkcloud job is failing with

```
/root/github.com/SUSE-Cloud/automation/scripts/github-status/github-status.rb -r crowbar/crowbar-ha -a set-status -s success -t https://ci.suse.de/job/openstack-mkcloud/68229/ -c 8ea1629ae9537b362b8e1dcabb8f74f395813d0f --context suse/mkcloud -m 'PR gating succeeded'
/usr/lib64/ruby/gems/2.1.0/gems/octokit-4.7.0/lib/octokit/response/raise_error.rb:16:in `on_complete': POST https://api.github.com/repos/crowbar/crowbar-ha/statuses/8ea1629ae9537b362b8e1dcabb8f74f395813d0f: 403 - API rate limit exceeded for susecloudapi. // See: https://developer.github.com/v3/#rate-limiting (Octokit::TooManyRequests)
```

which I think can be counted as a pass.

PR #201, opened 2017-06-19
drbd: Remove deprecated option for initial DRBD sync

jsuchome:
DRBD manual says "--force option replaces --overwrite-data-of-peer".
Although it was deprecated in 8.4 already, drbd startup started failing
only now when we switched to SLES12SP3.


jsuchome:
[#cloud8 Make OpenStack barclamps work for SP3](https://trello.com/c/DxMDobCg/33-cloud8-make-openstack-barclamps-work-for-sp3)

PR #202, opened 2017-06-27
[UNTESTED] make synchronization failure messages more helpful

aspiers:
Chef sync_mark synchronization failures are one of the most frequent source of problems and confusion in SOC.  We constantly receive support requests where the reported error is of a node timing out whilst waiting for another node, because users don't realise that that is not the underlying cause of the error so they don't check the correct log file.

So tweak the error messages to make it more obvious what's going on, and to encourage users to check the correct log files to find out and report the real problem, not just a symptom of it.

We choose to extend the log messages with reference to specific non-FQDN hostnames, as the FQDN versions would make the messages unnecessarily long.  If it's type A synchronization (i.e. "wait for founder"), tell the user which node is the founder.  If it's type B ("wait for all other nodes") then tell the user which nodes we're still waiting for.  A nice side-effect of this last change is that we calculate the remaining nodes via set subtraction rather than by length comparison, which should be more accurate.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
@@ -56,6 +56,17 @@ require "timeout"
 
 module CrowbarPacemakerSynchronization
 
+  def self.founder_at_revision(founder, cluster_name, mark, revision)
+    return false if founder.nil?
+    current_rev = \
+      begin
+        founder[:pacemaker][:sync_marks][cluster_name][mark]
+      rescue NoMethodError
+        nil
+      end
+    current_rev == revision
+  end
+
   # See "Synchronization helpers" documentation
   def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout = 60)
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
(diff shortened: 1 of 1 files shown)


jsuchome:
Looks promising!

Itxaka:
Looking good and will be really helpful!

Should we also include the cluster name in the messages in case there is more than 1 cluster in the environment?

aspiers:
@Itxaka Nice idea thanks!

aspiers:
Looks like I finally got a version which doesn't break everything :-)

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -130,24 +150,33 @@ def self.synchronize_on_mark(node, mark, revision, fatal = false, timeout = 60)
...
+    Chef::Log.info("Checking if all #{cluster_name} cluster nodes have " \
+                   "set #{mark} to #{revision}...")
 
+    nodes_with_mark_set = []
     begin
       Timeout.timeout(timeout) do
         while true
-          search_result = []
-          Chef::Search::Query.new.search(:node, "pacemaker_config_environment:#{node[:pacemaker][:config][:environment]} AND pacemaker_sync_marks_#{cluster_name}_#{mark}:#{revision}") { |o| search_result << o }
+          nodes_with_mark_set = []
```
Itxaka: Arent you emptying the var on every loop here?

aspiers: Yes, because the Array gets re-populated each time the search query is run.  If that bothers you  then let me know why!  It seems very unlikely but *maybe* not impossible that a node could have the desired sync mark set at one point, and then later have it removed, in which case we wouldn't want the other nodes to proceed until it's put back.  However a more likely scenario might be that somehow one of the nodes' sync marks gets a *higher* revision number, in which case it would no longer be returned by this search query (which only returns nodes matching the exact revision number given).  In that case maybe it's safer to block all nodes from proceeding, since then the cluster is out of sync, and not blocking would probably lead to undefined behaviour.

aspiers:
@Itxaka Replied.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
Itxaka: Thanks!

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -68,29 +79,38 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
...
           sleep(5)
         end # while true
       end # Timeout
     rescue Timeout::Error
       if fatal
-        message = "Cluster founder didn't set #{mark} to #{revision}!"
+        message = \
+          "Cluster founder #{founder_name} didn't set #{mark} " \
+          "to #{revision}! Timed out while waiting for the founder; " \
+          "Please check the chef-client logs for that node to see " \
```
bmwiedemann: how about
```
"Please check /var/log/crowbar/chef-client/#{founder_name}*.log
```

aspiers:
Unforunately the change to the new synchronization mechanism totally broke this PR, but we still really need it. I'll try to do another version unless someone else has the time to do it?  Maybe the SWAT squad, since they stand to gain the most from not having users ask FAQs about why nodes timed out on synchronization?

aspiers:
Obsoleted by #302

PR #206, opened 2017-07-07
haproxy: remove not working check-ssl option (bsc#1045783 bsc#1046093)

Itxaka:
The check-ssl option does not seem to work properly as reported on
the mentioned bugs. Changing to the normal check option and leaving the ssl
check to the ssl-hello-chk option seems to make the checks work properly.

Tested extensively with different barclamps, options and ssl configs, this seems to work as expected for both SSL endpoints and normal ones.

### DIFF
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -52,11 +52,7 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
     <% end -%>
 
     <% content[:servers].each do |server| -%>
-      <% if content[:use_ssl] -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check-ssl verify none inter 2000 rise 2 fall 5
-      <% else -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check inter 2000 rise 2 fall 5
-      <% end -%>
+	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check verify none inter 2000 rise 2 fall 5
     <% end -%>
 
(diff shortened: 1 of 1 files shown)


Itxaka:
[P2: Bug 1045783 - #L3 Keystone SSL error on on control node after applying PTF for BZ#1032756](https://trello.com/c/Q9KXh7yl/329-p2-bug-1045783-l3-keystone-ssl-error-on-on-control-node-after-applying-ptf-for-bz1032756)

Itxaka:
[P2: Bug 1046093 - #L3 Horizon SSL dashboard unreachable](https://trello.com/c/OA7Zwb7J/332-p2-bug-1046093-l3-horizon-ssl-dashboard-unreachable)

cmurphy:
Is it possible that the config needs both 'check' and 'check-ssl' as in example 8 here? https://www.haproxy.com/doc/aloha/7.0/haproxy/healthchecks.html

Itxaka:
@cmurphy just tested in case I missed it on my other tests, but it fails :(

cmurphy:
Is there a bug open for the haproxy package or upstream haproxy?

Itxaka:
Not that I could see that could affect this: http://www.haproxy.org/bugs/bugs-1.6.11.html

I am guessing that it is a config issue. Maybe we dont need check-ssl and instead we need to just use a normal tcp check with ssl-hello-chk as we do here, maybe that configuration was valid on older versions of haproxy, maybe we need to forward the proxy headers, and so on.

So for the moment this works and fixes both bugs, checking and balancing both ssl and non-ssl services properly. I still need to deep dive into the configuration and do more testing to see how we can improve this further, but for the moment that option should be enough:

```
When "option ssl-hello-chk" is set, pure
SSLv3 client hello messages are sent once the connection is established to
the server, and the response is analyzed to find an SSL server hello message.
The server is considered valid only when the response contains this server
hello message.
```

PR #208, opened 2017-07-12
Rework haproxy config for stickiness and balance strategy

matelakat:
This is the re-based version of #179 to see if tests fail

 - We enable stickiness based on SSL sessions
 - We add the ability to do stickiness based on cookies
 - We allow to define the balance strategy on a per-resource basis
 - We stop overriding the default balance strategy (and therefore use roundrobin by default)

### DIFF
--- a/chef/cookbooks/haproxy/providers/loadbalancer.rb
+++ b/chef/cookbooks/haproxy/providers/loadbalancer.rb
@@ -47,6 +47,9 @@ action :create do
   else
     section["mode"] = new_resource.mode
   end
+  unless new_resource.balance.empty?
+    section["balance"] = new_resource.balance
+  end
 
   section["stick"] = new_resource.stick
   section["stick"]["expire"] ||= "30m"
--- a/chef/cookbooks/haproxy/providers/loadbalancer.rb
+++ b/chef/cookbooks/haproxy/providers/loadbalancer.rb
@@ -47,9 +47,7 @@ action :create do
   else
     section["mode"] = new_resource.mode
   end
-  unless new_resource.balance.empty?
-    section["balance"] = new_resource.balance
-  end
+  section["balance"] = new_resource.balance unless new_resource.balance.empty?
 
   section["stick"] = new_resource.stick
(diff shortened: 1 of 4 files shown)


chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -40,11 +40,47 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
...
-    <% if content[:use_ssl] -%>
-       mode tcp
-       balance source
-    <% else -%>
 	mode <%= content[:mode] %>
+    <% unless content[:balance].nil? -%>
+	balance <%= content[:balance] %>
+    <% end -%>
+
+    <% if content[:use_ssl] # http://blog.exceliance.fr/2011/07/04/maintain-affinity-based-on-ssl-session-id/ -%>
```
Itxaka: can we set this url to the proper one? that redirects to `http://www.haproxy.com/blog/maintain-affinity-based-on-ssl-session-id/`

chef/cookbooks/haproxy/providers/loadbalancer.rb
```diff
@@ -47,6 +47,13 @@
   else
     section["mode"] = new_resource.mode
   end
+  unless new_resource.balance.empty?
```
Itxaka: could you fix this?

chef/cookbooks/haproxy/resources/loadbalancer.rb
```diff
@@ -25,6 +25,8 @@
 attribute :address, kind_of: String,  default: "0.0.0.0"
 attribute :port,    kind_of: Integer, default: 0
 attribute :mode,    kind_of: String,  default: "http", equal_to: ["http", "tcp", "health"]
+attribute :balance, kind_of: String,  default: "", equal_to: ["", "roundrobin", "static-rr", "leastconn", "first", "source"]
```
Itxaka: And this one :D

PR #214, opened 2017-07-31
service_object: Use rails logger directly

rsalevsky:
This commit continues the work which was done in
https://github.com/crowbar/crowbar-core/pull/1025. Which means we don't
have to pass the logger anymore.

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -17,7 +17,7 @@
 
 class PacemakerService < ServiceObject
   def initialize(thelogger = nil)
-    super(Rails.logger)
+    super
     @bc_name = "pacemaker"
   end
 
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -17,7 +17,7 @@
 
 class PacemakerService < ServiceObject
   def initialize(thelogger = nil)
```
cmurphy: If you also update the object here https://github.com/crowbar/crowbar-ha/blob/754081238726561a93fbc16a0392c01d965e7f2a/chef/data_bags/crowbar/migrate/pacemaker/100_stonith_in_rails.rb#L14 you could remove the arguments here

rsalevsky:
@cmurphy Fixed it.

crowbar_framework/app/models/pacemaker_service.rb
cmurphy: @rsalevsky do we still need the argument in this method signature?

cmurphy: Some classes inherit from this one and still initialize with the logger so we do still need to keep this.

PR #215, opened 2017-08-01
Add a support to use 'backup' option in server line.

jsuchome:
'backup' means that given server is only used in load
balancing when all other non-backup servers are unavailable.

We might need a setup (e.g. with mariadb cluster) where we want
to prevent existence of multiple servers being accessed at the same time.

### DIFF
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -82,16 +82,20 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
 	stick match req.cook(<%= cookie %>)
       <% end -%>
     <% end -%>
-
+    <% if content[:stick][:on] -%>
+        stick-table type ip size 1
+        stick on <%= content[:stick][:on] %>
+    <% end -%>
     <% content[:options].each do |option| -%>
 	option <%= option %>
     <% end -%>
 
     <% content[:servers].each do |server| -%>
+      <% backup = server[:backup] ? "backup" : "" %>
       <% if content[:use_ssl] -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check check-ssl verify none inter 2000 rise 2 fall 5
+	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check check-ssl verify none inter 2000 rise 2 fall 5 <%= backup %>
       <% else -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check inter 2000 rise 2 fall 5
+	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check inter 2000 rise 2 fall 5 <%= backup %>
       <% end -%>
     <% end -%>
 


jsuchome:
> It seems that this configuration will fallback immediately once a none backup server become available 
> again

Yes, but do we mind? I think the only thing we need is that haproxy directs the traffic to only one server, and that would be still true...

rhafer:
> Yes, but do we mind? 

If all clients need to reconnect once more (which I didn't verify). Yes, I think we do mind.

jsuchome:
@rhafer do you think that `peers` options is needed, or could it be enough with just `stick on dst`... or maybe with `stick-table`, but without `peers`?

jsuchome:
[Confiugure HAProxy for galera](https://trello.com/c/BQJLG0lL/7-confiugure-haproxy-for-galera)

flaviodsr:
I am not an expert, but for me seems like there is no way to avoid failback.

According to the [haproxy docs](https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#3.5), the ``peers`` option is used to propagate the ``stick-table`` entries between different running haproxy instances. In SOC haproxy is controlled by pacemaker so, there is only one haproxy instance running at a time. In that way, the ``stick-table`` entries (which I suppose is required for correctly avoiding the failback) cannot be propagated to other haproxy instance.

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -88,10 +88,11 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
     <% end -%>
 
     <% content[:servers].each do |server| -%>
+      <% backup = server[:backup] ? "backup" : "" %>
```
jhesketh: (nit) In case we want to start plugging other options I would lead this with a space (eg " backup") so that they can be chained together.

Otherwise instead of perhaps providing backup as a boolean in server we could pass through a string of extended_options

jsuchome: Hm; I'm adding that space directly at lines 93/95, so if we have just this "backup" option, I think it's fine this way.

However, if we end up adding more option, I think your idea of extended_options (or server_options) is good. Let's see if there's anything else we need.

jhesketh: Oh yeah, absolutely fine. Purely a nit :-)

rhafer:
> I am not an expert, but for me seems like there is no way to avoid failback.

Neither am I :)

> According to the haproxy docs, the peers option is used to propagate the stick-table entries between different running haproxy instances. In SOC haproxy is controlled by pacemaker so, there is only one haproxy instance running at a time. In that way, the stick-table entries (which I suppose is required for correctly avoiding the failback) cannot be propagated to other haproxy instance.

@flaviodsr The type of failback I was worrying about is not about haproxy itself failing over to another node. But in the case that the currently active  galera node goes down. HAproxy will automatically direct  all the traffic to the next available node. But as soon as the first node comes back up it will redirect  all traffic to the first node again. So effectively we'll have two disruption. And it seems the sticky-table can be used to avoid that.

@jsuchome I am all fine to merge a first version of this without the setting, but we should be aware of the impact. Failover of all database connection can have funny effects and we should only do that if really needed.

jsuchome:
We really need to test both scenarios, node/VIP/haproxy failure (situation handled by pacemaker) and mysql failure (which should be handled by haproxy only)

rhafer:
> We really need to test both scenarios, node/VIP/haproxy failure (situation handled by pacemaker) and mysql failure (which should be handled by haproxy only)

Sure, but AFAIK for haproxy/VIP pacemaker will take care to avoid a failback. (IIRC we set a default for resource-stickiness)

jsuchome:
> @rhafer:
> @flaviodsr The type of failback I was worrying about is not about haproxy itself failing over to another > node. But in the case that the currently active galera node goes down....

So for this case, it still might be good to expand the haproxy option. @flaviodsr - would it make sense then to have just `stick on dest` option and simple `stick-table`, but *without* `peers`? (Because we only care about one instance of haproxy...)

flaviodsr:
@rhafer I see your point, and I totally agree with you!

@jsuchome I think it does make sense, there is no point in using the ``peers`` options if we will have only one haproxy instance running at a time.

PR #216, opened 2017-08-01
Let cluster_vhostname find correct name for database clusters.

jsuchome:
database cluster has different name scheme and we might need it
to configure with HAProxy (haproxy code uses cluster_vhostname).

This may be totally wrong approach, I know...


jsuchome:
Also, this would mean that we could not allow database with services in the same cluster...

jsuchome:
Should not be needed

PR #217, opened 2017-08-07
Reimplement sync marks without node searches

abel-navarro:
This is a continuation on the job started by @vuntz here: https://github.com/crowbar/crowbar-ha/pull/170

The main benefit here is that we don't depend on the chef server
(through searches), which should decrease the load on the admin server
but also make things faster.

There's a slight difference in behavior for the sync marks: only the
founder nodes is waiting for all other nodes on sync; the non-founder
nodes simply wait for the founder because the founder will only set the
mark once all other nodes have already set their mark.

The initial sync mark (about the founder node being setup) is replaced
by a simple attribute on the node, in a similar way that we mark remote
nodes having been setup.

It's worth mentioning that we don't use revisions but only booleans for
the sync marks: revisions are deprecated.

Finally, we convert pre-existing sync marks to this model, which should
make this change non-disruptive.


aspiers:
@abel-navarro Please can you give PRs more descriptive titles?  Otherwise my mailbox fills up with meaningless notifications.  Thanks!

Itxaka:
This kind of works but unfortunately it still times out with a 180s timeout.

After triggering this manually it seems like the founder node does so much stuff than the other nodes that it leads to a timeout always. Looking at the code, It seems that due how we set the sync marks, now the other nodes are much much faster than the founder, as the founder has some costly operations (nova and neutron `not_if` guard take a loong time to execute).

On my local tests, with databse,rabbit,keystone,swift,glance,cinder,neutron,nova,horizon,heat,ceilometer,aohd deployed in an HA env we got the following chef runtimes:

Node1 (Founder):
[2017-08-18T09:42:01+00:00] INFO: Chef Run complete in 251.091648812 seconds

Node2:
[2017-08-18T09:39:15+00:00] INFO: Chef Run complete in 84.705649803 seconds


So Im guessing that node2 is not waiting properly on the sync-mark, as it should be much more closer to the founder.

Itxaka:

Itxaka:
Looks like the nova migrations are taking longer than 120 seconds, which is the timeout set on https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/nova/recipes/database.rb#L30

The founder setting the mark vs the node waiting for the mark happens almost in the same second, leading to a timeout from the node:

```
[2017-08-21T10:07:38+00:00] INFO: CrowbarPacemakerCIBAttribute.set: d52-54-77-77-01-01:crowbar_sync-nova_database:1
[2017-08-21T10:07:38+00:00] INFO: Processing package[nova-api] action install (nova::api line 18)

==> /var/log/crowbar/chef-client/d52-54-77-77-01-02.vn2.cloud.suse.de.log <==
[2017-08-21T10:07:39+00:00] FATAL: Cluster founder didn't set nova_database!

```

Itxaka:
Seems that there is a desync issue, as we are not waiting for all the nodes to reach the same point in the nova database recipe before starting the timeout, so the non-founders could start the timeout before the founder.

PR at https://github.com/crowbar/crowbar-openstack/pull/1181

Itxaka:
Triggered a CI run with the artifact generated on https://github.com/crowbar/crowbar-openstack/pull/1181 to check if that would fix it

Itxaka:
Same stuff on heat: https://github.com/crowbar/crowbar-openstack/pull/1183

This issue affects potentially all barclamps :(

Itxaka:
Looks liket he hidden issue was that we had changed the `node[:pacemaker][:founder]` from a bool to the full name of the founder so there was one method that was returning an empty founder, leading to the sync marks never cleaning up.

Once that was fixed, looks like we were not cleaning up properly the flag that lets the other nodes know that we have finished resetting the sync marks, so I already fixed that. 

This should now pass the tests :D

chef/cookbooks/crowbar-pacemaker/libraries/cibattribute.rb
```diff
@@ -0,0 +1,114 @@
...
+    # This is what we want:
+    # $ crm_attribute --name=crowbar_sync-pacemaker_setup --node=d52-54-77-77-77-2 --query --default=0
+    # Could not map name=d52-54-77-77-77-2 to a UUID
+    # $ crm_attribute --name=crowbar_sync-pacemaker_setup --node=d52-54-77-77-77-02 --query --default=0
+    # scope=nodes  name=crowbar_sync-pacemaker_setup value=0
+    validate_node(node)
+    validate_attribute(attribute)
+    validate_value(default)
+    output = `crm_attribute --name=#{attribute} --node=#{node} --query --default=#{default} 2>&1`
+    if output !~ /^Could not map name=/
```
cmurphy: Isn't this backwards? If output =~ /Could not map name/ that would mean it cannot fetch the attribute, right?

chef/cookbooks/crowbar-pacemaker/libraries/cibattribute.rb
```diff
@@ -0,0 +1,114 @@
...
+      xml = REXML::Document.new(`crm_mon --show-node-attributes --as-xml 2> /dev/null`)
+    rescue REXML::ParseException => e
+      raise "Cannot parse output of crm_mon when listing attributes!"
+    end
+
+    raise "No xml output from crm_mon when listing attributes!" if xml.root.nil?
+
+    all_attributes = {}
+
+    xml.root.elements.each("node_attributes/node[@name=\"#{node}\"]/attribute") do |e|
```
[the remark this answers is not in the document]
cmurphy: (nit) I agree with hound on this one, I would use element instead of e

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -182,11 +182,12 @@ def self.cluster_founder(node)
       node
     else
       begin
-        Node.load(node[:pacemaker][:founder])
+        Chef::Node.load(node[:pacemaker][:founder])
       rescue Net::HTTPServerException => e
         raise "No cluster founder found!" if e.response.code == "404"
         raise e
       end
+    end
```
cmurphy: These two changes seem like syntax issues, would this even have worked without this fixup commit?

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -75,7 +75,7 @@
   # we can't reset sync marks if the pacemaker stack is not setup...
   CrowbarPacemakerSynchronization.reset_marks(node) if node[:pacemaker][:setup]
   # ... but we don't want to block the other nodes forever
-  node.set[:pacemaker][:reset_sync_marks]
+  node.set[:pacemaker][:reset_sync_marks] = false
```
cmurphy: :+1: I was wondering about that

cmurphy:
github wants to make it really really hard to review one commit at a time so I'm not sure my comments are clear.

It seems like the last three commits are fixes for the first three commits - reviewing would have been easier if they were fixedup in the commit that introduced them (and with something this complex it's nice to try to make reviewing as easy as possible)

chef/cookbooks/crowbar-pacemaker/resources/sync_mark.rb
```diff
@@ -55,8 +50,8 @@
 # paramater
 attribute :mark,      kind_of: String,  default: nil
 
-# this is optional in crowbar; the barclamp proposal revision will be used
+# deprecated
 attribute :revision,  kind_of: Integer, default: nil
 
 attribute :fatal,     kind_of: [TrueClass, FalseClass], default: true
-attribute :timeout,   kind_of: Integer, default: 60
+attribute :timeout,   kind_of: Integer, default: 120
```
cmurphy: Isn't the point of this change to improve performance? This sort of indicates we went a step backward to me

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
Itxaka: This should probably never have even be possible to upload to the chef server due to the syntax error.

chef/data_bags/crowbar/migrate/pacemaker/100_founder_attribute.rb
```diff
@@ -0,0 +1,38 @@
+def upgrade(ta, td, a, d)
+  founder_name = d["elements"]["pacemaker-cluster-member"].first
+  d["elements"]["pacemaker-cluster-member"].each do |n|
+    begin
+      node = Chef::Node.load(name)
```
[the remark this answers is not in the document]
Itxaka: YES! The only way we are not hitting this is because  `d["elements"]["pacemaker-cluster-member"]` so we dont run over this code on a clean install.
Great catch!

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -301,62 +301,44 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
-      #    we want to keep)
-      if founder.nil?
-        member_nodes.each do |member_node|
-          if (member_node[:pacemaker][:founder] rescue false) == true
-            founder = member_node
-            break
-          end
-        end
+      #  - the proposal was deactivated (in which case we lost the info on
+      #    which node was the founder, but that's no big issue)
```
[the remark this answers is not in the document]
Itxaka: I dont think its possible to change founders between proposals, unless you deactivate the proposal.

Regarding why is this, I dont know the reasoning for this, as It was in place when I took over this. Let me have a look more deeply into it to see why this could be, but AFAIK the old founder fqdn should still be available on the node[:pacemaker][:founder] so there should be no reason to check if the old founder is still on the cluster and use it. 

Not sure if it makes any sense, as there should not be much differences between one node or other being the founder after barclamp deactivation, other than the founder doing more stuff than the other nodes.

chef/data_bags/crowbar/migrate/pacemaker/100_founder_attribute.rb
```diff
@@ -0,0 +1,38 @@
...
+      next unless node.key?("pacemaker")
+      founder_name = name if node["pacemaker"]["founder"] == name
+      node["pacemaker"].delete("founder")
+      node.save
+    rescue Net::HTTPServerException => e
+      raise e if e.response.code != "404"
```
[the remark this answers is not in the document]
Itxaka: I returns a 404 if the node doesnt exists, which should not be possible I think

chef/cookbooks/crowbar-pacemaker/libraries/cibattribute.rb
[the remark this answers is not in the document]
Itxaka: +1 will fix

chef/cookbooks/crowbar-pacemaker/libraries/cibattribute.rb
Itxaka: Good catch!

Itxaka:
Marking do not merge as I need to address @cmurphy comments. Plus after checking the migration, I have a feeling that its not properly implemented, so I would like to have an extra look in there

chef/cookbooks/crowbar-pacemaker/libraries/cibattribute.rb
Itxaka: ooh, this was fixed in a later commit :D

crowbar_framework/app/models/pacemaker_service.rb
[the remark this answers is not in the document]
Itxaka: I think this is meant to be run only if the proposal has been deactivated (thus the role deleted, and that info lost) so it should not affect most of our deployments like ever. 

If there is a problem where we change founders between proposal runs, then the rest of the code is broken, and the environment is broken, as that is not suppose to happen ever I think.

chef/data_bags/crowbar/migrate/pacemaker/100_founder_attribute.rb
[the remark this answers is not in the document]
Itxaka: fixed!

chef/data_bags/crowbar/migrate/pacemaker/100_founder_attribute.rb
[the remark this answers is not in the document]
Itxaka: Fixed with a comment

Itxaka:
@cmurphy fixed/answered most of your comments and reworked the commits

matelakat:
@cmurphy are you happy with the changes?

chef/cookbooks/corosync/recipes/authkey.rb
```diff
@@ -28,30 +28,30 @@
   return
 end
 
-# Find pre-existing authkey on other node(s)
-query  = "chef_environment:#{node.chef_environment}"
-query += " AND corosync_cluster_name:#{cluster_name}"
-
 # FIXME: move this Crowbar-specific code into the crowbar-pacemaker cookbook
 is_crowbar = !(node[:crowbar].nil?)
```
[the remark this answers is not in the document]
cmurphy: My guess is it means it's managed by crowbar, as opposed to being a standalone chef cookbook.

chef/cookbooks/crowbar-pacemaker/libraries/cibattribute.rb
```diff
@@ -0,0 +1,123 @@
+#
+# Copyright 2016, SUSE
```
cmurphy: nit - 2017

chef/data_bags/crowbar/migrate/pacemaker/101_founder_attribute.rb
```diff
@@ -0,0 +1,38 @@
...
+      node = Chef::Node.load(name)
+      node.set["pacemaker"]["founder"] = (name == founder_name)
+      node.save
+    rescue Net::HTTPServerException => e # node doesn not exists on the chef server
+      raise e if e.response.code != "404"
+    end
+  end
+  a.delete("founder")
+  return a, d
+end
```
cmurphy: Does this need to be in the 200s since this is against master?

chef/data_bags/crowbar/migrate/pacemaker/101_founder_attribute.rb
```diff
@@ -0,0 +1,38 @@
...
+      node = Chef::Node.load(name)
+      next unless node.key?("pacemaker")
+      founder_name = name if node["pacemaker"]["founder"]
+      node["pacemaker"].delete("founder")
+      node.save
+    rescue Net::HTTPServerException => e # node doesn not exists on the chef server
+      raise e if e.response.code != "404"
+    end
+  end
+  # are we a role? If yes, save the founder info there
```
cmurphy: I didn't realize schema migrations could operate on roles, that's kind of confusing. How does this get run?

chef/data_bags/crowbar/migrate/pacemaker/101_founder_attribute.rb
```diff
@@ -0,0 +1,38 @@
+def upgrade(ta, td, a, d)
+  founder_name = d["elements"]["pacemaker-cluster-member"].first
+  d["elements"]["pacemaker-cluster-member"].each do |name|
+    begin
+      node = Chef::Node.load(name)
+      next unless node.key?("pacemaker")
+      founder_name = name if node["pacemaker"]["founder"]
```
cmurphy: Is this right? Shouldn't it be `founder_name = name if node["pacemaker"]["founder"] == name` (looked like it used to be that in the first commit)

cmurphy:
Also - some of the fixup work for syntax errors ended up on the second commit when they are fixing up the first, e.g.:

https://github.com/crowbar/crowbar-ha/pull/217/commits/70e1620299140596006a33be182c5b54a55f6728#diff-43e12b18bb9031c640a53693301b6668R27
https://github.com/crowbar/crowbar-ha/pull/217/commits/70e1620299140596006a33be182c5b54a55f6728#diff-cd3c60551cceb8cc4c7ceb96e2467706R185

I know it seems nitpicky and git makes it a PITA to fix but since this is a fairly complicated change it would be really beneficial to have the ability to revert individual commits if we need to, so different commits shouldn't be fixing each other up. (I wouldn't ask at this point except I think there's already at least one real issue in the migration that needs to be fixed)

chef/data_bags/crowbar/migrate/pacemaker/101_founder_attribute.rb
Itxaka: indeed!

chef/data_bags/crowbar/migrate/pacemaker/101_founder_attribute.rb
Itxaka: yep, good catch!

chef/data_bags/crowbar/migrate/pacemaker/101_founder_attribute.rb
Itxaka: When you do a migration it runs for the barclamp **and** the role (if any) for the proposals attached to a barclamp: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/schema_migration.rb#L60

cmurphy:
I pushed a minor reorganization of some of the syntax fixups but didn't touch anything else.

chef/data_bags/crowbar/migrate/pacemaker/101_founder_attribute.rb
Itxaka: actually no, sorry. 

So the old `node["pacemaker"]["founder"]` was a bool, so this would yield the name of the node which has that key set to true. So this works as expected.

Itxaka:
@cmurphy That may be because the migration is not being run. The template-rabbit.json was poiting to the migration 100 as the latest :/

Itxaka:
@cmurphy I just tested this on a fresh environment with pacemaker+database already deployed with the old syncmarks, patch applied, rabbitmq proposal deployed on cluster, and it worked as expected.

I could see the migration working on the logs:

```
D, [2017-09-05T14:24:31.599291 #1390:0x000000006c2618] DEBUG -- SQL: INSERT/UPDATE transaction: filtering query to not expose potentially sensitive data
D, [2017-09-05T14:24:31.600573 #1390:0x000000006c2618] DEBUG --    (0.5ms)  COMMIT
D, [2017-09-05T14:24:32.015159 #1390:0x000000006c2618] DEBUG -- Saving role: pacemaker-config-proposal_1 - 6
D, [2017-09-05T14:24:32.015542 #1390:0x000000006c2618] DEBUG -- Acquire role:pacemaker-config-proposal_1 lock enter as uid 487
D, [2017-09-05T14:24:32.015696 #1390:0x000000006c2618] DEBUG -- Acquiring role:pacemaker-config-proposal_1 lock with {:shared=>false}
D, [2017-09-05T14:24:32.015739 #1390:0x000000006c2618] DEBUG -- Lock /opt/dell/crowbar_framework/tmp/role:pacemaker-config-proposal_1 attempt 1
D, [2017-09-05T14:24:32.015779 #1390:0x000000006c2618] DEBUG -- Acquire role:pacemaker-config-proposal_1 lock exit: #<File:/opt/dell/crowbar_framework/tmp/role:pacemaker-config-proposal_1>
D, [2017-09-05T14:24:32.077335 #1390:0x000000006c2618] DEBUG -- Release role:pacemaker-config-proposal_1 lock enter: #<File:/opt/dell/crowbar_framework/tmp/role:pacemaker-config-proposal_1>
D, [2017-09-05T14:24:32.077443 #1390:0x000000006c2618] DEBUG -- Release role:pacemaker-config-proposal_1 lock exit
D, [2017-09-05T14:24:32.077503 #1390:0x000000006c2618] DEBUG -- Done saving role: pacemaker-config-proposal_1 - 7
```
So Im guessing that the error on the migration numbering was the issue here, the migration did not run for your run so it failed to find the key.

matelakat:
```
Error (88): 'crowbar database proposal --file=/root/database.default.proposal edit default' failed with exit code: 1
```

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -17,16 +17,9 @@
...
-# this code while the founder node has not its attributes indexed yet on the
-# server side. This only happens the very first time (which is when we don't
-# even have an auth key); on next runs, we know we're good.
-if node[:corosync][:authkey].nil?
-  include_recipe "crowbar-pacemaker::wait_for_founder"
-end
+require "timeout"
 
-node.normal[:corosync][:cluster_name] = CrowbarPacemakerHelper.cluster_name(node)
-node.normal[:corosync][:bind_addr] = Barclamp::Inventory.get_network_by_type(node, "admin").address
```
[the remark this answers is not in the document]
scottwulf: Agreed. Without it bind_addr is left as its hard-coded, default value of 192.168.124.0.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -296,67 +296,49 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
...
-        end
+        old_founder_name = old_role.default_attributes["pacemaker"]["founder"]
+        founder_name = old_founder_name if members.include?(old_founder_name)
 
         # the founder from the old role is not there anymore; let's promote
         # another node to founder, so we get the same authkey
-        if founder.nil?
-          founder = old_nodes.first
+        if founder_name.nil?
+          old_members = attributes["elements"]["pacemaker-cluster-member"]
```
vuntz: `attributes` here is wrong, it should be `old_role.override_attributes[@bc_name]` (since we care about the old role, not the new one).

matelakat:
Let me test this change

crowbar_framework/app/models/pacemaker_service.rb
Itxaka: Isnt this done once we have found out that there is no founder from the `old_role` so we pick a new one? In that case, we **need** to use the `role` to get it, as there is no founder in the `old_role` for whatever reason. Even the comment above says so.

Maybe you meant to post this on line 315? Also, do we store the crowbar pushed attributes into the override or default? Because if its on default then line 315 is already taking care of this.

Itxaka: oh no, I see, we want to pick up a node from the `old_role` for the authkey to be the same. Duuh

chef/cookbooks/corosync/recipes/authkey.rb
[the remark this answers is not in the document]
matelakat: Could we ever have it False in our environment then?

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -17,16 +17,9 @@
...
+node[:corosync][:cluster_name] = CrowbarPacemakerHelper.cluster_name(node)
```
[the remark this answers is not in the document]
Itxaka: Im keeeping this on the current PR, not sure if needed but I rather do a cleanup afterwards with more testing

Itxaka:
@vuntz updated by addressing your comments!

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -66,105 +64,124 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
...
+    elsif !CrowbarPacemakerCIBAttribute.cib_up_for_node?
       if fatal
-        message = "Some cluster nodes didn't set #{mark} to #{revision}!"
+        message = "Node #{node[:hostname]} does not have CIB connection"
         Chef::Log.fatal(message)
         raise message
       else
-        message = "Some cluster nodes didn't set #{mark} to #{revision}! Going on..."
-        Chef::Log.warn(message)
+        Chef::Log.debug("Node does not have CIB connection." \
```
vuntz: Gah, I missed that: s/debug/warn/ here

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -16,14 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
```
vuntz: (and minor style nitpick: keep the empty line after the copyright header)

Itxaka:
@vuntz done!

chef/data_bags/crowbar/migrate/pacemaker/100_founder_attribute.rb
```diff
@@ -0,0 +1,38 @@
+def upgrade(ta, td, a, d)
+  founder_name = d["elements"]["pacemaker-cluster-member"].first
+  d["elements"]["pacemaker-cluster-member"].each do |name|
+    begin
+      node = Chef::Node.load(name)
+      next unless node.key?("pacemaker")
+      founder_name = name if node["pacemaker"]["founder"] == name
```
matelakat: don't we want something like:

`founder_name = name if node["pacemaker"]["founder"]`

instead? During the upgrade `node["pacemaker"]["founder"]` is still a boolean, isn't it?

Itxaka:
> @Itxaka could we amend the commits, so the migration script is only changed on the first commit? It seems strange that both 3 commits are touching that poor file. I was reviewing it commit by commit, and found the same issue on the first commit that was fixed in the second. I still need to read the rest of the commits.

@matelakat I only see the migration script on the first commit, the others do not touch it?

chef/data_bags/crowbar/migrate/pacemaker/100_founder_attribute.rb
Itxaka: Indeed, and the only reason this worked is because we already got the founder name in the first line. Great catch!

Itxaka: ahh, this was fixed afterwards in a commit,.

chef/cookbooks/corosync/recipes/authkey.rb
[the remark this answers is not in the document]
matelakat: I received the answer from vuntz: it was kept there so if we decided to upstream cookbooks then they would remain compatible, but in our environment this will always be true

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -36,17 +30,57 @@
...
-  # phases, and this wait is fatal in case of errors
-  timeout 120
-  fatal true
-end.run_action(:guess)
+unless CrowbarPacemakerHelper.is_cluster_founder?(node)
+  begin
+    # we use a long timeout because the wait / attribute-set are in two different
+    # phases, and this wait is fatal in case of errors
+    Timeout.timeout(120) do
+      Chef::Log.info("Waiting for cluster founder to be setup...")
```
matelakat: `to be set up` I think but we should really ask someone who knows English.

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -36,17 +30,57 @@
...
+      loop do
+        founder = CrowbarPacemakerHelper.cluster_founder(node)
+        # be safe, in case the node doesn't have pacemaker attributes yet
+        # (depends on whether chef already started on it or not)
+        if founder.fetch("pacemaker", {})[:setup]
+          if founder[:pacemaker][:reset_sync_marks]
+            Chef::Log.info("Cluster founder is resetting sync marks, waiting...")
+          else
+            Chef::Log.info("Cluster founder is setup, going on...")
```
matelakat: I think it is two words `set up` here as well

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -36,17 +30,57 @@
...
+            break
+          end
+        else
+          Chef::Log.info("Cluster founder not setup yet, waiting...")
```
matelakat: And here

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -36,17 +30,57 @@
...
+        end
+        sleep(5)
+      end # while true
+    end # Timeout
+  rescue Timeout::Error
+    raise "Cluster founder not setup!"
```
matelakat: and here

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -36,17 +30,57 @@
...
+    CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
+      CrowbarPacemakerSynchronization.migrate_sync_marks_v1(cluster_node)
+    end
+  end
+  node.normal_attrs[:pacemaker].delete(:sync_marks)
+  dirty = true
+end
+
+if CrowbarPacemakerHelper.is_cluster_founder?(node) && node[:pacemaker][:reset_sync_marks]
+  # we can't reset sync marks if the pacemaker stack is not setup...
```
matelakat: and here

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -57,8 +91,13 @@
 # saving the corosync authkey attribute is done in convergence phase for
 # founder (but reading the attribute is done in compile phase for non-founder
 # nodes) -- see the corosync::authkey_generator recipe.
-crowbar_pacemaker_sync_mark "create-pacemaker_setup" do
-  revision node[:pacemaker]["crowbar-revision"]
+ruby_block "mark node as ready for pacemaker" do
+  block do
+    unless node[:pacemaker][:setup]
```
matelakat: I guess we should have called this `ready` instead of `setup`, but it's not time to address these issues

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -66,105 +64,124 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
...
+    # founder waits for the mark to be set on all non-founders, and then sets
+    # its mark; if the mark is already set, we can skip everything
+    if CrowbarPacemakerCIBAttribute.get(node[:hostname], "#{prefix}#{mark}", "0") != "0"
+      Chef::Log.info("Synchronization cluster mark #{mark} already set.")
+      return
+    end
 
     if CrowbarPacemakerHelper.being_upgraded?(node)
       Chef::Log.debug("Node is being upgraded." \
         "Skipping wait loop for all other cluster nodes.")
```
matelakat: We are sending a log message that we are skipping something, but then we don't return.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -66,105 +64,124 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
...
-    cluster_name = CrowbarPacemakerHelper.cluster_name(node)
-
-    node[:pacemaker][:sync_marks] ||= {}
-    node[:pacemaker][:sync_marks][cluster_name] ||= {}
-    if node[:pacemaker][:sync_marks][cluster_name][mark] != revision
-      node[:pacemaker][:sync_marks][cluster_name][mark] = revision
-      node.save
-      Chef::Log.info("Setting founder cluster mark #{mark} to #{revision}.")
+    if CrowbarPacemakerCIBAttribute.get(node[:hostname], "#{prefix}#{mark}", "0") != "0"
+      Chef::Log.info("Synchronization cluster mark #{mark} already set.")
```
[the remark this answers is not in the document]
Itxaka: I gues its just a guard in case that we have set the mark already for some reason? I dont think this code should be hit, unless there is some kind of issue with the marks.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
Itxaka: Good point!

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -66,105 +64,124 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
...
+      end
+    else
+      begin
+        Chef::Log.info("Checking if all other cluster nodes have set #{mark}...")
+
+        Timeout.timeout(timeout) do
+          CrowbarPacemakerHelper.cluster_nodes_names(node).each do |name|
+            next if name == node[:hostname]
+            loop do
+              break if CrowbarPacemakerCIBAttribute.get(name, "#{prefix}#{mark}", "0") != "0"
```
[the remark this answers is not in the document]
Itxaka: Im torn regarding this. One one side, having that `is_set` method makes sense. On the other, having a extra class just for a one liner is terrible and bad practice IMHO, classes are big bloats of code for simple things and they make it more difficult to follow up the code when you need to keep jumping from file to file when reading the code.

In this case `#{prefix}#{mark}` can be added as a var at the function level so we can reuse it across all the calls to avoid repetition, but I would be against the creation of a class for that.

For the 2 extra methods...I dont know. I rather add them to the `CrowbarPacemakerCIBAttribute` as helper methods instead, even if they obscure a bit the code, if you are checking this code you would also have the `CrowbarPacemakerCIBAttribute` file open so it should be easier to read/debug

@matelakat @abel-navarro @vuntz thoughts?

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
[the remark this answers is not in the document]
vuntz: Actually, it should occur in periodic chef runs, since the marks are not reset before they happen. And I think it's relevant when debugging to know that the mark was already set.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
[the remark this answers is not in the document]
matelakat: If you are OK with it, I can read it, it just hurts my eyes. Let's get it in, and then maybe we can see how a post cleanup pr would look like.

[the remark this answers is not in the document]
vuntz: I'm fine with the current way, but I guess we could add two small helpers:

```
def self.set_mark(hostname, mark)
  CrowbarPacemakerCIBAttribute.set(hostname, "#{prefix}#{mark}", "1")
end

def self.is_mark_set?(hostname, mark)
  CrowbarPacemakerCIBAttribute.get(hostname, "#{prefix}#{mark}", "0") != "0"
end
```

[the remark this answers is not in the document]
Itxaka: I would say lets move this into a follow up card so it can be re-tested properly and merge this so we have the base feature already available

Itxaka:
@mate @vuntz updated:
 - Fixed setup -> set up
 - Added the missing return statement when an upgrade is ongoing
 - Used the `attribute` var to contain `#{prefix}#{mark}` to avoid so much repetition

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -66,105 +64,129 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
...
-      node[:pacemaker][:sync_marks][cluster_name][mark] = revision
-      node.save
-      Chef::Log.info("Setting synchronization cluster mark #{mark} to #{revision}.")
-    else
-      Chef::Log.info("Synchronization cluster mark #{mark} already set to #{revision}.")
+    # non-founders simply set the mark and then wait for the founder to set the
+    # mark
+    unless CrowbarPacemakerHelper.is_cluster_founder?(node)
+      Chef::Log.info("Setting synchronization cluster mark #{mark}.")
+      CrowbarPacemakerCIBAttribute.set(node[:hostname], "#{prefix}#{mark}", "1")
```
matelakat: ok, we shall fix this as well, but let it be a separate PR. Get this thing going

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -66,105 +64,129 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
...
       if fatal
-        message = "Some cluster nodes didn't set #{mark} to #{revision}!"
+        message = "Node #{node[:hostname]} does not have CIB connection"
         Chef::Log.fatal(message)
         raise message
       else
-        message = "Some cluster nodes didn't set #{mark} to #{revision}! Going on..."
-        Chef::Log.warn(message)
+        Chef::Log.warn("Node does not have CIB connection." \
+          "Skipping wait loop for all other cluster nodes.")
```
vuntz: Sorry, but... don't we need a `return` here too, to not set the sync mark in case we can read attributes later on? It may very well be that we don't need it (it's starting to be my opinion); we'd need a comment to explain that, in that case.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -66,105 +64,129 @@ def self.wait_for_mark_from_founder(node, mark, revision, fatal = false, timeout
...
+    elsif !CrowbarPacemakerCIBAttribute.cib_up_for_node?
       if fatal
-        message = "Some cluster nodes didn't set #{mark} to #{revision}!"
+        message = "Node #{node[:hostname]} does not have CIB connection"
         Chef::Log.fatal(message)
         raise message
       else
-        message = "Some cluster nodes didn't set #{mark} to #{revision}! Going on..."
-        Chef::Log.warn(message)
+        Chef::Log.warn("Node does not have CIB connection." \
```
vuntz: And I guess we need a space after the `.`, as it's generally good style to have a space between two sentences... This is valid for line 131 too, which is not touched in this PR, so I guess it's just some copy & paste thing.

matelakat:
I triggered a re-build as it seems CI failure was caused by some temporary zypper error.

vuntz:
All green, yay!

@Itxaka were you planning to do the backport?

Itxaka:
Finally got in \o/
@vuntz doing the backport rigth now!

aspiers:

PR #219, opened 2017-08-11
[stable/4.0] Add a support to use 'backup' option in server line.

jsuchome:
'backup' means that given server is only used in load
balancing when all other non-backup servers are unavailable.

'stick on dst' should make sure that destination server is not
left even if it is backup one and non-backup is healthy again.
This should prevent additional disruption once the failing node
is back and running.

We might need a setup (e.g. with mariadb cluster) where we want
to prevent existence of multiple servers being accessed at the same time.

(cherry picked from commit c1cf765bb38932ed5747e749ebf5341846094dd7)

Backport of https://github.com/crowbar/crowbar-ha/pull/215


jsuchome:
[P3 (3) Confiugure HAProxy for galera](https://trello.com/c/BQJLG0lL/7-p3-3-confiugure-haproxy-for-galera)

PR #220, opened 2017-08-17
Allow configuration of more options for server lines.

jsuchome:
fastinter/rise/fall values influence haproxy monitoring behavior.
See e.g. https://www.digitalocean.com/community/tutorials/how-to-use-haproxy-to-set-up-mysql-load-balancing--3

PR #222, opened 2017-08-22
A quick tidy up of haproxy server options

jhesketh:
Following on from PR #220, improve the code readability of the server
options. This change places the logic in a clear position before
spitting it out into the server line.

### DIFF
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -91,13 +91,15 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
     <% end -%>
 
     <% content[:servers].each do |server| -%>
-      <% backup = server[:backup] ? " backup" : "" %>
-      <% fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : "" %>
-      <% if content[:use_ssl] -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check check-ssl verify none inter <%= server[:inter] || 2000 %><%= fastinter %> rise <%= server[:rise] || 2 %> fall <%= server[:fall] || 5 %><%= backup %>
-      <% else -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check inter <%= server[:inter] || 2000 %><%= fastinter %> rise <%= server[:rise] || 2 %> fall <%= server[:fall] || 5 %><%= backup %>
-      <% end -%>
+      <%
+        ssl = content[:use_ssl] ? " check-ssl verify none" : ""
+        inter = " inter #{server[:inter] || 2000}"
+        fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : ""
+        rise = " rise #{server[:rise] || 2}"
+        fall = " fall #{server[:fall] || 5}"
+        backup = server[:backup] ? " backup" : ""
+      %>
+        server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check<%= ssl %><%= inter %><%= fastinter %><%= rise %><%= fall %><%= backup %>
     <% end -%>
 
   <% end -%>


chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -91,13 +91,15 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
...
     <% content[:servers].each do |server| -%>
-      <% backup = server[:backup] ? " backup" : "" %>
-      <% fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : "" %>
-      <% if content[:use_ssl] -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check check-ssl verify none inter <%= server[:inter] || 2000 %><%= fastinter %> rise <%= server[:rise] || 2 %> fall <%= server[:fall] || 5 %><%= backup %>
-      <% else -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check inter <%= server[:inter] || 2000 %><%= fastinter %> rise <%= server[:rise] || 2 %> fall <%= server[:fall] || 5 %><%= backup %>
-      <% end -%>
+      <%
+        ssl = content[:use_ssl] ? " check-ssl verify none"
```
jsuchome: Youre missing `: ""`

jhesketh: thanks :-)

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -91,13 +91,15 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
...
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check check-ssl verify none inter <%= server[:inter] || 2000 %><%= fastinter %> rise <%= server[:rise] || 2 %> fall <%= server[:fall] || 5 %><%= backup %>
-      <% else -%>
-	server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check inter <%= server[:inter] || 2000 %><%= fastinter %> rise <%= server[:rise] || 2 %> fall <%= server[:fall] || 5 %><%= backup %>
-      <% end -%>
+      <%
+        ssl = content[:use_ssl] ? " check-ssl verify none" : ""
+        inter = " inter " + (server[:inter] || 2000)
+        fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : ""
+        rise = " rise " + String(server[:rise] || 2)
+        fall = " fall " + String(server[:fall] || 5)
```
jsuchome: I do not really like this way of type conversion ... I would use `#{...}` , or at `to_s`. But maybe some Ruby expert should decide.

BTW, you're missing it with `inter`.

jhesketh: Right, I'm quite new to ruby so I'm not sure which is preferable or even the differences. Happy to take advice

inter is on line 96?

jsuchome: Yes, I mean line 96.

Itxaka: Agreed, please dont sum strings, rather use string interpolation for that

```
rise = " rise #{server[:rise] || 2}"
```

I also dont really like the space in the vars, I rather have the spaces in the line below instead, which makes more sense visually, but that is personal preference with no other logic behind it :P

jsuchome: > I rather have the spaces in the line below instead

The problem (small, I agree) is with extra spaces in the line in case you do not have any value for an option.

jhesketh: Thanks for the feedback, will re-roll.
Regarding the spaces, jsuchome is right, some of the variables may be unset. We could construct the params as a dict and stitch them together which may be neater.

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -91,13 +91,15 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
...
-      <% end -%>
+      <%
+        ssl = content[:use_ssl] ? " check-ssl verify none" : ""
+        inter = " inter #{server[:inter] || 2000}"
+        fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : ""
+        rise = " rise #{server[:rise] || 2}"
+        fall = " fall #{server[:fall] || 5}"
+        backup = server[:backup] ? " backup" : ""
+      %>
+      server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check<%= ssl %><%= inter %><%= fastinter %><%= rise %><%= fall %><%= backup %>
```
nicolasbock: Please have a look at https://github.com/crowbar/crowbar-openstack/pull/1186 with regards to ERB syntax. @cmurphy started a very nice cleanup effort of our ERB usage and I think this would fit in here very neatly.

nicolasbock: Sorry, forgot to say that otherwise I really like the PR! :smile:

jsuchome: @jhesketh tjhe indenting is still not quite right, it looks like http://pastebin.nue.suse.com/25434/src now

PR #223, opened 2017-08-24
crowbar-pacemaker: Allow overriding start/stop commands for crm_resource

rhafer:
The 1st commit is just a pretty simple fix for `service_is_running?` to make it work with Master/Slave resources.

The 2nd commit enhances the pacemaker "service" provider to allow passing in alternative start/stop commands when used with the "restart_crm_resource" flag. We need this for the galera resource agent. In order to restart mysql on a single node we need to get the resource though a demote/promote cycle once (instead of start/stop). See e.g. https://github.com/jhesketh/crowbar-openstack/commit/ce00239fe10848b591a561a16d28a43d7f68a5a7 for the current work in progress.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/providers/service.rb
+++ b/chef/cookbooks/crowbar-pacemaker/providers/service.rb
@@ -179,15 +179,17 @@ action :restart do
   use_crm_resource = new_resource.supports[:restart_crm_resource]
   no_maintenance_mode = new_resource.supports[:no_crm_maintenance_mode]
   pacemaker_resource = new_resource.supports[:pacemaker_resource_name] || service_name
+  resource_stop_cmd = new_resource.supports[:crm_resource_stop_cmd] || "force-stop"
+  resource_start_cmd = new_resource.supports[:crm_resource_start_cmd] || "force-start"
 
   if service_is_running?(service_name, use_crm_resource, pacemaker_resource)
     set_maintenance_mode unless no_maintenance_mode
 
     if use_crm_resource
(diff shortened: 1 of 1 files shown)


rhafer:
[P11 galera config: Find the best value for SST](https://trello.com/c/27Q5XW7K/28-p11-galera-config-find-the-best-value-for-sst)

chef/cookbooks/crowbar-pacemaker/providers/service.rb
```diff
@@ -221,10 +221,14 @@
 def service_is_running?(name, use_crm_resource, pacemaker_resource)
   if use_crm_resource
     `crm_resource --force-check --resource #{pacemaker_resource}`
+    # For Master/Slave resources "monitor" will return OCF_RUNNING_MASTER (8)
+    # on nodes that are running the resource in Master role currently. We need
+    # to treat that as a successfully result as well.
+    $?.success? || $?.exitstatus == 8
```
nicolasbock: Wow, that is certainly unusual :frowning_face:.

jsuchome: Don't we need to limit this to certain services only?

rhafer: Why? OCF_RUNNING_MASTER is only ever returned for Master/Slave resources.

jhesketh:
This looks good to me.

chef/cookbooks/crowbar-pacemaker/providers/service.rb
jsuchome: OK.

PR #225, opened 2017-08-28
[4.0] crowbar-pacemaker: Allow overriding start/stop commands for crm_resource

jsuchome:
Backport of https://github.com/crowbar/crowbar-ha/pull/223


jsuchome:
[P11 galera config: Find the best value for SST](https://trello.com/c/27Q5XW7K/28-p11-galera-config-find-the-best-value-for-sst)

PR #226, opened 2017-09-01
set default bmc_interface to lanplus

tpatzig:
The stonith agent external/ipmi requires an interface to work.
The model is reading bmc_interface from the node's ipmi attributes,
but there is no such attribute stored/set on the node.
To circumvent, we default to "lanplus", which makes the stonith ipmi
resource work properly.

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -762,7 +762,7 @@ class PacemakerService < ServiceObject
         end
         params["userid"] = cluster_node["ipmi"]["bmc_user"]
         params["passwd"] = cluster_node["ipmi"]["bmc_password"]
-        params["interface"] = cluster_node["ipmi"]["bmc_interface"]
+        params["interface"] = cluster_node["ipmi"]["bmc_interface"] || "lanplus"
 
         stonith_attributes["per_node"]["nodes"][stonith_node_name] ||= {}
         stonith_attributes["per_node"]["nodes"][stonith_node_name]["params"] = params


cmurphy:
lanplus should already be the default: https://github.com/crowbar/crowbar-core/blob/fdd395c9fe111402f35e6e07eda1f31a4b1fe02c/chef/data_bags/crowbar/template-ipmi.json#L9

It is relatively recent, is it possible that a schema migration didn't run? Or that crowbar-ha and crowbar-core are out of sync where you found this problem?

bmwiedemann:
or they are using SOC6 and the template change is not on the crowbar 3.0 branch

cmurphy:
But in that case SOC6 wouldn't try to use the attribute here.

tpatzig:
no we're using SOC7 here. But we do not have commit https://github.com/crowbar/crowbar-core/commit/aa5972d9a48b590d37ca53f9e2fc174ac082b11c
We have latest crowbar-core from SOC7-Updates repo `crowbar-core-4.0+git.1497443258.d629532c8-13.2`, which does not include that commit, yet.

cmurphy:
The commit that introduced it in crowbar-ha shouldn't be in SOC7 either - https://github.com/crowbar/crowbar-ha/commit/c108bef064ebc17c5cca39aa775f5cbb8588ce8c

tpatzig:
But it is in crowbar-ha. Here we have `crowbar-ha-4.0+git.1500189498.a38e2a0-4.9.1`, which is slightly newer (latest commit is from 16th of july) and already inlcudes the `bmc_interface`.
Thats why I was wondering, why `bmc_interface` it is nowhere else used/set.
So it seems we just need the latest crowbar-core; can you push that please.

tpatzig:
as commented, this is already fixed, we just need the updated versions in the official update channel

PR #227, opened 2017-09-01
Increase pacemaker timeout to 120 (bsc#1056473)

bmwiedemann:
to make it longer than the 90s that systemd is waiting
before it sends a SIGKILL to a stuck service.

This can help to avoid needless random fencing of nodes during shutdown/reboot
https://bugzilla.suse.com/show_bug.cgi?id=1056473

### DIFF
--- a/chef/cookbooks/pacemaker/attributes/default.rb
+++ b/chef/cookbooks/pacemaker/attributes/default.rb
@@ -52,7 +52,9 @@ default[:pacemaker][:founder] = nil
 default[:pacemaker][:is_remote] = false
 default[:pacemaker][:crm][:initial_config_file] = "/etc/corosync/crm-initial.conf"
 default[:pacemaker][:crm][:no_quorum_policy] = "ignore"
-default[:pacemaker][:crm][:op_default_timeout] = 60
+# Should be longer than the systemd timeouts (defaults to 90s) so that
+# pacemaker only reacts when systemd is not helping anymore
+default[:pacemaker][:crm][:op_default_timeout] = 120
 default[:pacemaker][:crm][:migration_threshold] = 3
 
 # acceptable CIB syntax version; if lower is detected, we must force its upgrade


jsuchome:
Just seen something similar with haproxy, but @rhafer suggested solution based on systemd service...

bmwiedemann:
it seems some systemd services like cinder-scheduler sometimes just take longer than 60s to stop, though...

jsuchome:
So shouldn't we rather increase the timeouts for the specific services we already know are problematic?

bmwiedemann:
we could **de**crease the haproxy service timeout to be lower than pacemaker's 60s

jsuchome:
> we could decrease the haproxy service timeout to be lower than pacemaker's 60s

Yep, that's one of the solution @rhafer suggested.

I was referring to your cinder-scheduler and this PR - we might change the value of `op_default_timeout` for specific resources only, not globalyl.

rhafer:
> Just seen something similar with haproxy, but @rhafer suggested solution based on systemd service...

That was specifically for the haproxy issue. Since we know that stopping the service will fail (bsc#1056371) regardless how long we set the timeout.

I can't speak much for other resources we manage via pacemaker. But unconditionally increasing all timeouts for all pacemaker operations (at least those that don't specifically override the timeout) sounds like a bit too much. Won't this also have a bad effect on failover times and such?

aspiers:
> Won't this also have a bad effect on failover times and such?

Presumably only for resources which are not specifying an `op monitor` interval, and there probably shouldn't be any of those.  The problem is that Pacemaker doesn't allow setting separate global defaults for start/stop/monitor: http://clusterlabs.org/doc/en-US/Pacemaker/1.0/html/Pacemaker_Explained/s-operation-defaults.html

dirkmueller:
I think this sounds like a good approach of making pacemaker less agressive. when systemd successfully killed a service after 90s then pacemaker should consider that a "succes" and not by itself take action.

PR #228, opened 2017-09-04
haproxy: provide a option to ratelimit frontends

Itxaka:
Provides a new option on the loadbalancer resource to allow for rate
limiting per frontend. A integer values is passed and it signifies the
maximum request per minute from an IP.

Also provides a custom 429 error file for haproxy to server when the
rate limit is hit, so clients receive the proper response

Based on the code from https://github.com/sap-oc/crowbar-openstack/issues/21

### DIFF
--- a/chef/cookbooks/haproxy/attributes/default.rb
+++ b/chef/cookbooks/haproxy/attributes/default.rb
@@ -21,6 +21,7 @@ default[:haproxy][:platform][:package] = "haproxy"
 default[:haproxy][:platform][:user] = "haproxy"
 default[:haproxy][:platform][:group] = "haproxy"
 default[:haproxy][:platform][:config_file] = "/etc/haproxy/haproxy.cfg"
+default[:haproxy][:platform][:error_dir] = "/etc/haproxy/errorfiles"
 
 default[:haproxy][:global][:maxconn] = 4096
 default[:haproxy][:global][:bufsize] = 16384
--- a/chef/cookbooks/haproxy/recipes/setup.rb
+++ b/chef/cookbooks/haproxy/recipes/setup.rb
@@ -24,4 +24,30 @@ template node[:haproxy][:platform][:config_file] do
   owner "root"
   group "root"
   mode 00644
+  variables(
+    lazy {
+      {
+        rate_limit_enabled: node[:haproxy][:sections].keys.each.any? {
+            |type| node[:haproxy][:sections][type].any? {
+              |service, values| !values["rate_limit"].nil? && !values["rate_limit"].to_i.zero?
+          }
+        }
+      }
+    }
+  )
+end
+
+directory node[:haproxy][:platform][:error_dir] do
+  action :create
+  owner "root"
+  group "root"
+  mode 0o644
+end
+
+cookbook_file "#{node[:haproxy][:platform][:error_dir]}/429.http" do
+  source "429.http"
+  action :create
+  owner "root"
+  group "root"
+  mode 0o644
 end
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -101,6 +101,25 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
       %>
         server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check<%= ssl %><%= inter %><%= fastinter %><%= rise %><%= fall %><%= backup %>
     <% end -%>
+	<% unless content[:rate_limit].nil? || content[:rate_limit].to_i.zero? -%>
+	# Rate limiting config options
+	tcp-request inspect-delay 5s
+	acl too_many_reqs_by_user sc0_gpc0_rate() gt <%= content[:rate_limit] %>
+	acl mark_seen sc0_inc_gpc0 gt 0
+	stick-table type string size 100k expire 60s store gpc0_rate(60s)
+	tcp-request content track-sc0 src
+
+	use_backend be_429_slow_down if mark_seen too_many_reqs_by_user
+	# End Rate limiting config options
+	<% end -%>
 
   <% end -%>
+<% end -%>
+
+<% if @rate_limit_enabled -%>
+backend be_429_slow_down
+	mode http
+	timeout tarpit 2s
+	errorfile 500 <%= node[:haproxy][:platform][:error_dir] %>/429.http
+	http-request tarpit
 <% end -%>
--- a/chef/cookbooks/haproxy/files/default/429.http
+++ b/chef/cookbooks/haproxy/files/default/429.http
@@ -1,0 +1,7 @@
+HTTP/1.1 429 Too Many Requests
+Cache-Control: no-cache
+Connection: close
+Content-Type: text/plain
+Retry-After: 60
+
+Too Many Requests (HAP429).\r\n
--- a/chef/cookbooks/haproxy/providers/loadbalancer.rb
+++ b/chef/cookbooks/haproxy/providers/loadbalancer.rb
@@ -63,6 +63,10 @@ action :create do
   end
   section["servers"] = new_resource.servers
 
+  if new_resource.rate_limit && new_resource.rate_limit.to_i != 0
+    section["rate_limit"] = new_resource.rate_limit
+  end
+
   node["haproxy"]["sections"][new_resource.type] ||= {}
   node["haproxy"]["sections"][new_resource.type][new_resource.name] = section
 end
--- a/chef/cookbooks/haproxy/resources/loadbalancer.rb
+++ b/chef/cookbooks/haproxy/resources/loadbalancer.rb
@@ -31,3 +31,4 @@ attribute :use_ssl, kind_of: [TrueClass, FalseClass], default: false
 attribute :stick,   kind_of: Hash,    default: {}
 attribute :options, kind_of: Array,   default: []
 attribute :servers, kind_of: Array,   default: []
+attribute :rate_limit, kind_of: Integer, default: nil


Itxaka:
[Nova OpenStack API rate Limit](https://trello.com/c/GsZ6iHh9/40-nova-openstack-api-rate-limit)

Itxaka:
Blocked by https://github.com/crowbar/crowbar-ha/pull/229

Itxaka:
Maybe a better way of setting the rate_limit would be:

```
haproxy_loadbalancer "keystone-admin" do
  address node[:keystone][:api][:admin_host]
  port node[:keystone][:api][:admin_port]
  use_ssl (node[:keystone][:api][:protocol] == "https")
  servers CrowbarPacemakerHelper.haproxy_servers_for_service(node, "keystone", "keystone-server", "admin_port")
  action :nothing
  rate_limit {"requests": 5, "timeframe": "1s"}
end.run_action(:create)
```

So its more flexible and we can select 5 request per second or 300 per minute and so on?

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -101,6 +101,28 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
...
+	tcp-request content track-sc0 src
+
+	use_backend be_429_slow_down if mark_seen too_many_reqs_by_user
+	# End Rate limiting config options
+	<% end -%>
 
   <% end -%>
 <% end -%>
+
+<% if node[:haproxy][:sections].keys.each.any? {
```
aspiers: This is quite a lot of logic to put in the template.  I think it would be more readable/cleaner if you calculated this in the recipe and assigned it to a clearly named variable which then gets passed into the template.

Itxaka: Yeah, its quite terrible. The problem I have no idea on how to check this properly to make it more generic, and not make it resource dependant.

This is because of how the config is generated, which is not done on the recipes but after all resources have been compiled with the `haproxy_loadbalancer` resource during the compilation phase, then the config template is deployed and it gets the values from the `node[:haproxy]`

I could set a mark in the node, but Im not sure if that would be less visible. What a strange approach in this cookbook :)

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -101,6 +101,28 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
       %>
         server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check<%= ssl %><%= inter %><%= fastinter %><%= rise %><%= fall %><%= backup %>
     <% end -%>
+	<% unless content[:rate_limit].nil? %>
```
matelakat: Do we need to close it with a `-%>` so nothing is rendered of it evaluates to false?

matelakat: Shall we use `rate_limit_enabled` here as well?

Itxaka: Fixed!

Itxaka: In here we use rate_limit because that contains the value for the rate limit, and its dependant on the service that we are setting the frontend for, while on the other case, its just a generic flag under `node[:haproxy]` that indicates that at least one of the resources has that key, thus we need to set the generic backend to redirect the errors to

Itxaka: That was probably the worst explanation ever, I dont even understand what I wrote.

Lets see, there is 2 parts in here, one that is `node[:haproxy][:sections]` which contains the list of sections to write to the config, be them frontend/backends/listen . Basically that is what gets translated into the different endpoints that haproxy manages to loadbalance between servers.
So in that section, for each service that we set, we can set a maximum number of requests per minute by setting the `rate_limit` attribute on the `haproxy_loadbalancer` resource , like in here: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/recipes/ha.rb#L18 , so in that case, adding an extra attribute called `rate_limit 300` would make that haproxy endpoint rate limited to 300 requests per IP per minute.

Now, there is another part on the haproxy config that needs to be added, and that its a generic backend that discards the requests and returns a custom error. As this backend would be shared between all services that had rate limiting enabled, we can make it only appear if at least one of the services is rate limited, and that its done https://github.com/crowbar/crowbar-ha/pull/228/files#diff-95df28e46b705d8b4da406cb9ebb995aR66 .

So we use that to set a generic flag on the node in order to show that part of the config.

Still, doesnt matter much because that doesnt work as expected, nor Im sure that we can make it work easily :/

I could make a trigger on the haproxy template that calculates that value on the fly, but that would trigger after the template has been written, so....

Itxaka:
@aspiers @matelakat updated the ugly code on the template, probably makes it much easier to undertand the difference now on the last template version.

Also @aspiers totally tested, works as expected :)

chef/cookbooks/haproxy/recipes/setup.rb
```diff
@@ -24,4 +24,30 @@
   owner "root"
   group "root"
   mode 00644
+  variables(
+    lazy {
```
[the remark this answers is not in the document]
Itxaka: shut up, this need to be like this, dog.

chef/cookbooks/haproxy/recipes/setup.rb
```diff
@@ -24,4 +24,30 @@
...
+      {
+        rate_limit_enabled: node[:haproxy][:sections].keys.each.any? {
```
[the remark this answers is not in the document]
Itxaka: same

Itxaka:
Not sure what to do with hound, it looks ok rigth now, and makes it easier to understand. If I change it to what hound wants, is gonna look terrible.

chef/cookbooks/haproxy/attributes/default.rb
```diff
@@ -21,6 +21,7 @@
 default[:haproxy][:platform][:user] = "haproxy"
 default[:haproxy][:platform][:group] = "haproxy"
 default[:haproxy][:platform][:config_file] = "/etc/haproxy/haproxy.cfg"
+default[:haproxy][:platform][:error_dir] = "/etc/haproxy/errorfiles/"
```
vuntz: I'd prefer to not have a trailing slash here, because it leads to `cookbook_file "#{node[:haproxy][:platform][:error_dir]}429.http"` which I find unreadable ;-)

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
vuntz: I'm nearly sure this is missing a check, as we need to do the same check as in the recipe: `!values["rate_limit"].nil? && !values["rate_limit"].to_i.zero?`

So we need to check that the rate limit is not 0.

chef/cookbooks/haproxy/attributes/default.rb
Itxaka: indeed, fixed!

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
Itxaka: good one, fixed!

Itxaka:
@vuntz updated!

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -101,6 +101,25 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
       %>
         server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check<%= ssl %><%= inter %><%= fastinter %><%= rise %><%= fall %><%= backup %>
     <% end -%>
+	<% unless content[:rate_limit].nil? || content[:rate_limit].to_i.zero? %>
```
matelakat: Dont'we need to close it with `-%>` ?

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -101,6 +101,25 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
...
+	<% end -%>
 
   <% end -%>
 <% end -%>
+
+<% if @rate_limit_enabled -%>
+backend be_429_slow_down
+	mode http
+	timeout tarpit 2s
+	errorfile 500 /etc/haproxy/errorfiles/429.http
```
matelakat: This (`/etc/haproxy/errorfiles`) should come from a variable?

Itxaka: ah, thats a good point, it should indeed use the provided value

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
Itxaka: No need, but I will update it to be consistent with other templates

Itxaka:
> I am no haproxy expert, but it seems that we are doing quite intensive gymnastics to get a boolean value for the template so we can include the section backend be_429_slow_down - can't we just skip all the gymnastics and include that always? 

Dont agree there. We should only deploy that backend when its gonna be needed, or it will feel out of place and could lead to confusion when other people check the templates and find a strange backend not called from anywhere.

I dont think we are doing "intensive gymnastics" on it, its just a double iteration on a dictionary to check if there are any values present, although I admit that ruby makes it appear more complex than what it is.

Itxaka:
@matelakat updated!
@vuntz re-review pls? Only things changed are the 2 nits that Mate found.

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -101,6 +101,25 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
       %>
         server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check<%= ssl %><%= inter %><%= fastinter %><%= rise %><%= fall %><%= backup %>
     <% end -%>
+	<% unless content[:rate_limit].nil? || content[:rate_limit].to_i.zero? -%>
```
Itxaka: @vuntz actually the check for zero is useless as we only pass `content[:rate_limit]` in the case that we call the loadbalancer resource with that **and** its not zero: https://github.com/crowbar/crowbar-ha/pull/228/files#diff-95df28e46b705d8b4da406cb9ebb995aR67

vuntz: @Itxaka I'm happy if it's dropped, as long as it's also dropped in the other part of the code where it's checked :-)

PR #230, opened 2017-09-05
haproxy: Make sure that systemd kills haproxy service on restart

jsuchome:
It's possible that haproxy is unable to close open connections during
restart action so we make sure systemd kills it before we hit pacemaker
service timeout that would ultimately fence the haproxy node (bsc#1056371).

### DIFF
--- a/chef/cookbooks/haproxy/recipes/setup.rb
+++ b/chef/cookbooks/haproxy/recipes/setup.rb
@@ -25,3 +25,25 @@ template node[:haproxy][:platform][:config_file] do
   group "root"
   mode 00644
 end
+
+# Make sure that systemd kills haproxy services when restart action is requested.
+# It's possible haproxy is unable to close open connections during restart
+# so we make sure systemd kills it before we hit pacemaker service
+# timeout that would ultimately fence the haproxy node (bsc#1056371).
+
+systemd_override_dir = \
+  "/etc/systemd/system/#{node[:haproxy][:platform][:package]}.service.d"
+
+directory systemd_override_dir do
+  owner "root"
+  group "root"
+  mode "0755"
+  action :create
+end
+
+cookbook_file "#{systemd_override_dir}/10-timeout.conf" do
+  source "haproxy.service.override.conf"
+  owner "root"
+  group "root"
+  mode "0644"
+end
--- a/chef/cookbooks/haproxy/files/default/haproxy.service.override.conf
+++ b/chef/cookbooks/haproxy/files/default/haproxy.service.override.conf
@@ -1,0 +1,2 @@
+[Service]
+TimeoutStopSec=30


jsuchome:
@nicolasbock I guess it might but I do not know what else could we do... the problem (described https://bugzilla.suse.com/show_bug.cgi?id=1056371#c8 ) is that we know we want to restart it, but haproxy is not able to find out what should be closed.

It's indeed bug in haproxy we're working around here....

toabctl:
@jsuchome I have not read the bug report but isn't that something that should be added to the .service file in the haproxy package?

jsuchome:
@toabctl that's the other option, already requested by @s-t-e-v-e-n-k . Looks like there's a greater desire for such alrternative...

toabctl:
> Looks like there's a greater desire for such alrternative...

It will fix the problem for **all** users of haproxy, not just for crowbar users.

jsuchome:
TBH, I have no idea where does the haproxy package coming from to our projects

toabctl:
> TBH, I have no idea where does the haproxy package coming from to our projects

First thing would be for ```openSUSE:Factory``` and HAProxy is developed in ```server:http``` (see ```osc maintainer haproxy```).
And then for SLES it comes from ```SUSE:SLE-12-SP2:Update```

chef/cookbooks/haproxy/recipes/setup.rb
```diff
@@ -25,3 +25,25 @@
...
+  "/etc/systemd/system/#{node[:haproxy][:platform][:package]}.service.d"
+
+directory systemd_override_dir do
+  owner "root"
+  group "root"
+  mode "0755"
+  action :create
+end
+
+template "#{systemd_override_dir}/10-timeout.conf" do
```
rhafer: As this is really a static file being created, you'd normally use the `cookbook_file` resource here instead of template: https://docs.chef.io/resource_cookbook_file.html

And put the source file to `chef/cookbooks/haproxy/files/default/...`

jsuchome: Yeah, I was thinking about this... before I found https://www.frankmitchell.org/2013/03/chef-templates/
No, it looks like we're not gonna change the content, of that file. But still.. :-)

rhafer:
> It will fix the problem for all users of haproxy, not just for crowbar users.

@toabctl There is not problem in haproxy (well at least not with that timeout). It is really specific to our pacemaker-based setup. The default timeout is 90s, we're lowering that to 30s so that the kill happens before pacemaker runs into it's timeout (which we have configured to 60s).

chef/cookbooks/haproxy/recipes/setup.rb
jsuchome: But OK, I've changed it

PR #231, opened 2017-09-05
haproxy: Fix VIP creation for haproxy

rhafer:
Up to now the haproxy resource couldn't cleanly cope with with new
virtual IPs being added for haproxy after the initial deployment.
Since we are using "pacemaker_transaction" for creating new resources
and "pacemaker_primitive" for updating existing ones there was an
ordering issue when a new VIP was added when the haproxy group already
existed. The :update for the group was not executed since VIP was not
yet created.

With MariaDB support now available this is something that can happen
quite regulary. MariaDB only needs a VIP in the "admin" network, if
later "Keystone" is deployed to the same cluster it needs an additional
VIP for the "public" network.

We're trying to fix this shortcomming now by creating the VIPs outside
of the transaction (but with target-role="Stopped") and explicitly
starting them after the transaction ran.

Related-Bug: bsc#1056371

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
@@ -56,6 +56,19 @@ if node[:pacemaker][:haproxy][:clusters].key?(cluster_name) && node[:pacemaker][
   cluster_vhostname = CrowbarPacemakerHelper.cluster_vhostname(node)
   service_name = "haproxy"
 
+  # Make sure new pacemaker resources get created before before
+  # the group ("g-haproxy") is updated. Otherwise we run into ordering
+  # issue when subsequent chef-client runs need to create additional
+  # VIP resources and add them to the group.
+  pacemaker_transaction "haproxy service" do
+    # Use lazy evaluation here, as transaction_objects is populated
+    # further down in this recipe.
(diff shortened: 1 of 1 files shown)


rhafer:
With this the VIP is again started before the group is updated, which causes pacemaker to shuffle around resources, when the group is updated :/

rhafer:
Adding the do not merge label to add a wait loop which checks for the VIPs/haproxy being started.

rhafer:
> Adding the do not merge label to add a wait loop which checks for the VIPs/haproxy being started.

Adding a wait loop is hardly possible and very racey. But https://github.com/crowbar/crowbar-ha/pull/234 does the trick in much simpler way.

chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
```diff
@@ -56,6 +56,19 @@
   cluster_vhostname = CrowbarPacemakerHelper.cluster_vhostname(node)
   service_name = "haproxy"
 
+  # Make sure new pacemaker resources get created before before
+  # the group ("g-haproxy") is updated. Otherwise we run into ordering
+  # issue when sub-sequent chef-client runs need to create additional
```
s-t-e-v-e-n-k: Nit: subsequent is not a word that requires hyphenation.

PR #236, opened 2017-09-07
[4.0] pacemaker: Use --wait with crm configure command

rhafer:
Sometimes reconfiguring existing resources (e.g. groups) in the cluster
will trigger some resources of the cluster being moved and/or restarted.
Use --wait to block until the transition has finished and the cluster
has settled. This helps to avoid race issues e.g. when pacemaker needs
to restart resources after new resource is added to an existing group.

(cherry picked from commit 908a7743a6dffd627423508d7af9d7d3d5c47a13)


rhafer:
Backport of #234

rhafer:
rebased

rhafer:
rebased again

PR #237, opened 2017-09-07
[4.0] haproxy: Fix VIP creation for haproxy

rhafer:
Up to now the haproxy resource couldn't cleanly cope with with new
virtual IPs being added for haproxy after the initial deployment. Since
the "pacemaker_transaction" for creating new resources happened after
the "pacemaker_primitive" for updating existing ones there was an
ordering issue when a new VIP was added when the haproxy group already
existed.  The :update for the group was not executed since VIP was not
yet created.

With MariaDB support now available this is something that can happen
quite regulary. MariaDB only needs a VIP in the "admin" network, if
later "Keystone" is deployed to the same cluster it needs an additional
VIP for the "public" network.

This commit tries to address the issue by reodering the transaction and
the primitive/group resources. And by using lazy evaluation.

Related-Bug: bsc#1056371
(cherry picked from commit 58093f6f510c1e5697cb6b80cefc5a3a4e1bf55d)


rhafer:
backport of #231

PR #238, opened 2017-09-10
pacemaker: Add option to not manage stateless active/active services

vuntz:
Pacemaker is not strictly required for stateless active/active services,
since orchestration, fencing, and so on is not necessarily useful in
that case.

Enabling this option does simplify the setup in the OpenStack case quite
significantly, but comes with a cost, like pacemaker not restarting
services that have failed.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
@@ -17,6 +17,8 @@
 # limitations under the License.
 #
 
+return unless node[:pacemaker][:clone_stateless_services]
+
 # Define pacemaker primitive for apache service
 
 # This is required for the OCF resource agent
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -47,14 +47,15 @@
       "haproxy": {
         "admin_name": "",
         "public_name": ""
-      }
+      },
+      "clone_stateless_services": true
     }
   },
   "deployment": {
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 100,
+      "schema-revision": 200,
       "element_states": {
         "pacemaker-cluster-member"    : [ "readying", "ready", "applying" ],
         "hawk-server"                 : [ "readying", "ready", "applying" ],
--- a/crowbar_framework/config/locales/pacemaker/en.yml
+++ b/crowbar_framework/config/locales/pacemaker/en.yml
@@ -64,6 +64,15 @@ en:
             down or rebooted; you will have to manually start it after having
             fixed the issue. "Automatic" means that this setting will be set to
             "true" for two-nodes cluster, and to "false" otherwise.'
+        clone_stateless_services: 'Manage stateless active/active services with Pacemaker'
+        clone_stateless_services_hint:
+          'Stateless active/active services can be managed by Pacemaker, or can
+          run as standard services. If they are managed by Pacemaker, then
+          Pacemaker will automatically restart them on failures and will setup
+          cluster-wide constraints to ensure correct ordering when starting
+          services. If they run as standard services, then the complexity of
+          the configuration is reduced and this limits the causes for fencing
+          in case of failures.'
         crm:
           no_quorum_policy: 'Policy when cluster does not have quorum'
           no_quorum_policy_hint_html: 'Refer to the <a href="http://clusterlabs.org/doc/en-US/Pacemaker/1.1-crmsh/html/Pacemaker_Explained/_available_cluster_options.html">pacemaker documentation</a> for a description of each value.'
--- a/chef/data_bags/crowbar/migrate/pacemaker/200_clone_stateless_services.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/200_clone_stateless_services.rb
@@ -1,0 +1,10 @@
+def upgrade(ta, td, a, d)
+  # stay compatible with previous behavior
+  a["clone_stateless_services"] = true unless a.key?("clone_stateless_services")
+  return a, d
+end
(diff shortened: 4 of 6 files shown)


vuntz:
We may want an option to change the systemd services to use `Restart=on-failure`, but this would be a followup pull request. And actually, this would probably not be HA-specific, I guess.

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -47,14 +47,15 @@
       "haproxy": {
         "admin_name": "",
         "public_name": ""
-      }
+      },
+      "no_stateless_clone": true
```
vuntz: One open question: what should be the default? I set it to true just to have it used in the CI, but no strong opinion.

Itxaka: I would say false? If we lose automatic restarting and constrains, then it seems better to have it as default off, and if someone wants to enable then they know the downsides to it.

vuntz: @Itxaka the automatic restart can be done with systemd; it's a followup work item to this. So it's actually not that bad.

vuntz:
I updated the patch: the migration now sets this setting to false, to not change the behavior in an existing cluster.

chef/data_bags/crowbar/template-pacemaker.json
aspiers: > So it's actually not that bad.

Ugh, "it's not that bad" is not a compelling argument for making things obviously worse.  You're assuming (or committing to ensure) that the follow-up work item will happen soon.  Why not change the default as part of the follow-up work, when it's safe?

Also, the switch to systemd (which BTW is what RH has already done) rules out proper monitoring via OCF RAs (which we should have been using but never did IIRC).  The correct way to fill this gap is via HAProxy's `httpchk`.  This can get complicated by SSL termination.  @dirkmueller and I discussed this and related topics (`mod_wsgi` vs. `uwsgi` etc.) with the OpenStack Ansible folks at the PTG, and need to share our findings internally.

crowbar_framework/config/locales/pacemaker/en.yml
```diff
@@ -64,6 +64,13 @@ en:
             down or rebooted; you will have to manually start it after having
             fixed the issue. "Automatic" means that this setting will be set to
             "true" for two-nodes cluster, and to "false" otherwise.'
+        no_stateless_clone: 'Do not setup pacemaker resources for stateless active/active services'
+        no_stateless_clone_hint:
+          'Stateless active/active services can be run outside of pacemaker, as
+          standard services. This will reduce the impact of pacemaker on the
+          nodes, but this also means that pacemaker will not automatically
```
aspiers: "reduce the impact" is pretty vague.  The sentence seems to be implying that this is some kind of advantage, but it's not clear what it is; can you make it more explicit?

vuntz: Better now?

vuntz:
> Why not change the default as part of the follow-up work, when it's safe?

Works for me, I'll change it; just want to get some CI testing with the setting on for now.

vuntz:
>> Why not change the default as part of the follow-up work, when it's safe?
> Works for me, I'll change it; just want to get some CI testing with the setting on for now.

Changed now, I triggered a run with true as default in https://github.com/crowbar/crowbar-openstack/pull/1256

chef/cookbooks/crowbar-pacemaker/recipes/apache.rb
```diff
@@ -17,6 +17,8 @@
 # limitations under the License.
 #
 
+return if node[:pacemaker][:no_stateless_clone]
```
abel-navarro: It's hard to guess from the `no_stateless_clone` name that this is related to active/active services. Active/active services do not imply stateless services. Would it make sense to rename it to `no_activeactive_clone`?

vuntz: @abel-navarro well, unfortunately, active/active can be stateful too, so you lose some information by not specifying "stateless". However, the "clone" part of the name does imply "active/active". Hence the current name.

aspiers: I'd prefer `no_stateless_clones` (plural), or am I missing something?  Also, any particular reason why you didn't call it `stateless_clones` and have it default to true?  Having a negative option slightly decreases readability.

vuntz: Renamed to `clone_stateless_services` now.

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -47,14 +47,15 @@
       "haproxy": {
         "admin_name": "",
         "public_name": ""
-      }
+      },
+      "clone_stateless_services": false
```
aspiers: Default should now be true :-)

crowbar_framework/config/locales/pacemaker/en.yml
```diff
@@ -64,6 +64,14 @@ en:
             down or rebooted; you will have to manually start it after having
             fixed the issue. "Automatic" means that this setting will be set to
             "true" for two-nodes cluster, and to "false" otherwise.'
+        clone_stateless_services: 'Setup pacemaker resources for stateless active/active services'
+        clone_stateless_services_hint:
```
aspiers: Hint needs updating, since standard assumption is that it explains what selecting the option does, not what deselecting it does.

vuntz: @aspiers next try :-) Is the new hint better?

aspiers: Yes thanks!

aspiers:
See #300 which changes the default value of this new option.

PR #239, opened 2017-09-14
Add support for multiple Corosync rings

scottwulf:
Customer has requested the ability to configure an additional corosync
ring to "fully leverage the second bond in their dual bond environment".
The pacemaker cluster nodes will have visibility of one another over
both bonds and, when a failure occurs on the first bond, a cluster
failure will be averted due to the existence of the second ring over
the second bond.

In order to accomplish this, the flat list of single ring attributes
and web UI form fields is replaced by an array and _**n**_ number of web UI
form fields to support _**n**_ number of rings. The UI and validation code
restrict the maximum number of rings to 2. If there is ever a need for
for more rings, simply search for '2', 'two', and 'second' and make
the appropriate changes.

### DIFF
--- a/chef/cookbooks/corosync/recipes/config.rb
+++ b/chef/cookbooks/corosync/recipes/config.rb
@@ -34,9 +34,34 @@ unless %(udp udpu).include?(node[:corosync][:transport])
   raise "Invalid transport #{node[:corosync][:transport]}!"
 end
 
-if node[:corosync][:transport] == "udpu" && (node[:corosync][:members].nil? || node[:corosync][:members].empty?)
-  raise "Members have to be defined when using \"udpu\" transport!"
+rings = node[:corosync][:rings]
+
+member_count = nil
+rings.each.with_index do |ring, ring_index|
+  # check for empty
+  if ring[:members].nil? || ring[:members].empty?
+    raise "One or more members are required. The member list of "\
+      "ring #{ring_index} is empty."
+  end
+
+  # check for matching member count across rings
+  if member_count.nil?
+    member_count = ring[:members].length
+  elsif ring[:members].length != member_count
+    raise "All rings must have the same number of members. Ring #{ring_index} "\
+      "has #{ring[:members].length} members but all preceding rings have #{member_count}."
+  end
 end
+
+# gather member addresses together by node for corosync.conf.v2.erb by
+# building and transposing a two dimensional member array that is sorted
+# by ring 0 members to avoid unnecessary configuration reloads
+members_v2 = []
+sort_by_ring0 = node[:corosync][:rings][0][:members]
+node[:corosync][:rings].each do |ring|
+  members_v2.push(ring[:members].sort_by.with_index { |_, i| sort_by_ring0[i] })
+end
+members_v2 = members_v2.transpose
 
 template "/etc/corosync/corosync.conf" do
   if node[:platform] == "suse" && node[:platform_version].to_f < 12.0
@@ -49,10 +74,8 @@ template "/etc/corosync/corosync.conf" do
   mode 0600
   variables(
     cluster_name: node[:corosync][:cluster_name],
-    bind_addr: node[:corosync][:bind_addr],
-    mcast_addr: node[:corosync][:mcast_addr],
-    mcast_port: node[:corosync][:mcast_port],
-    members: node[:corosync][:members],
+    rings: node[:corosync][:rings],
+    members_v2: members_v2,
     transport: node[:corosync][:transport]
   )
 
--- a/chef/cookbooks/corosync/recipes/firewall.rb
+++ b/chef/cookbooks/corosync/recipes/firewall.rb
@@ -17,6 +17,15 @@
 # limitations under the License.
 #
 
+mcast_ports = node[:corosync][:rings].map { |n| n["mcast_port"] }
+if node[:corosync][:transport] == "udp"
+  # Corosync uses the port you specify for UDP messaging, and also the
+  # immediately preceding port. Thus if you specify 5405, Corosync
+  # sends messages from UDP port 5404 to UDP port 5405.
+  mcast_ports += mcast_ports.map { |n| n - 1 }
+end
+mcast_ports = mcast_ports.uniq.sort
+
 case node[:platform_family]
 when "suse"
   template "/etc/sysconfig/SuSEfirewall2.d/services/cluster" do
@@ -24,7 +33,7 @@ when "suse"
     mode "0640"
     owner "root"
     variables(
-      mcast_port: node[:corosync][:mcast_port]
+      mcast_ports: mcast_ports
     )
 
     # FIXME: where do I get the name for this from?
--- a/chef/cookbooks/corosync/templates/default/corosync.conf.erb
+++ b/chef/cookbooks/corosync/templates/default/corosync.conf.erb
@@ -45,21 +45,23 @@ totem {
         # This specifies the mode of redundant ring, which may be none, active, or passive.
         rrp_mode: none
 
+<% @rings.each.with_index do |ring, ring_index| -%>
         interface {
-                ringnumber:     0
-                bindnetaddr:    <%= @bind_addr  %>
+                ringnumber:     <%= ring_index %>
+                bindnetaddr:    <%= ring[:bind_addr]  %>
 <% if @transport == 'udp' -%>
-                mcastaddr:      <%= @mcast_addr %>
+                mcastaddr:      <%= ring[:mcast_addr] %>
 <% elsif @transport == 'udpu' -%>
-  <% @members.sort.each do |memberaddr| -%>
+  <% ring[:members].each do |memberaddr| -%>
                 member {
                         memberaddr:     <%= memberaddr %>
                 }
   <% end -%>
 <% end -%>
-                mcastport:      <%= @mcast_port %>
+                mcastport:      <%= ring[:mcast_port] %>
                 ttl:            1
         }
+<% end -%>
 
         transport: <%= @transport %>
 }
--- a/chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
+++ b/chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
@@ -44,15 +44,19 @@ totem {
 	# interface: define at least one interface to communicate
 	# over. If you define more than one interface stanza, you must
 	# also set rrp_mode.
+<% if @rings.length > 1 -%>
+	rrp_mode: passive
+<% end -%>
+<% @rings.each.with_index do |ring, ring_index| -%>
 	interface {
 		# Rings must be consecutively numbered, starting at 0.
-		ringnumber: 0
+		ringnumber: <%= ring_index %>
 		# This is normally the *network* address of the
 		# interface to bind to. This ensures that you can use
 		# identical instances of this configuration file
 		# across all your cluster nodes, without having to
 		# modify this option.
-		bindnetaddr: <%= @bind_addr  %>
+		bindnetaddr: <%= ring[:bind_addr]  %>
 		# However, if you have multiple physical network
 		# interfaces configured for the same subnet, then the
 		# network address alone is not sufficient to identify
@@ -67,19 +71,20 @@ totem {
 		# addresses across multiple Corosync clusters sharing
 		# the same network.
 <% if @transport == 'udp' -%>
-		mcastaddr: <%= @mcast_addr %>
+		mcastaddr: <%= ring[:mcast_addr] %>
 <% end -%>
 		# Corosync uses the port you specify here for UDP
 		# messaging, and also the immediately preceding
 		# port. Thus if you set this to 5405, Corosync sends
 		# messages over UDP ports 5405 and 5404.
-		mcastport: <%= @mcast_port %>
+		mcastport: <%= ring[:mcast_port] %>
 		# Time-to-live for cluster communication packets. The
 		# number of hops (routers) that this ring will allow
 		# itself to pass. Note that multicast routing must be
 		# specifically enabled on most network routers.
 		ttl: 1
 	}
+<% end -%>
 
 	transport: <%= @transport %>
 }
@@ -110,15 +115,15 @@ logging {
 	}
 }
 
-<% if @transport == 'udpu' -%>
 nodelist {
-  <% @members.sort.each do |memberaddr| -%>
+  <% @members_v2.each do |ring_addrs| -%>
 	node {
-		ring0_addr: <%= memberaddr %>
+    <% ring_addrs.each.with_index do |ring_addr, ring_index| -%>
+		ring<%= ring_index %>_addr: <%= ring_addr %>
+    <% end -%>
 	}
   <% end -%>
 }
-<% end -%>
 
 quorum {
 	# Enable and configure quorum subsystem (default: off)
--- a/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
@@ -20,7 +20,12 @@
 require "timeout"
 
 node.normal[:corosync][:cluster_name] = CrowbarPacemakerHelper.cluster_name(node)
-node.normal[:corosync][:bind_addr] = Barclamp::Inventory.get_network_by_type(node, "admin").address
+
+# set bindnetaddr per node, use host address rather than network subnet
+# in case multiple interfaces are configured on the same subnet
+node[:corosync][:rings].each_with_index do |ring, index|
+  ring[:bind_addr] = Barclamp::Inventory.get_network_by_type(node, ring[:network]).address
+end
 
 include_recipe "crowbar-pacemaker::quorum_policy"
 include_recipe "crowbar-pacemaker::stonith"
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -5,10 +5,15 @@
     "pacemaker": {
       "corosync": {
         "transport": "udpu",
-        "mcast_addr": "239.255.0.1",
-        "mcast_port": 5405,
         "password": "crowbar",
-        "require_clean_for_autostart_wrapper": "auto"
+        "require_clean_for_autostart_wrapper": "auto",
+        "rings": [
+          {
+            "network": "admin",
+            "mcast_addr": "",
+            "mcast_port": 5405
+          }
+        ]
       },
       "crm": {
         "no_quorum_policy": "stop"
@@ -55,7 +60,7 @@
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 200,
+      "schema-revision": 201,
       "element_states": {
         "pacemaker-cluster-member"    : [ "readying", "ready", "applying" ],
         "hawk-server"                 : [ "readying", "ready", "applying" ],
--- a/chef/data_bags/crowbar/template-pacemaker.schema
+++ b/chef/data_bags/crowbar/template-pacemaker.schema
@@ -17,10 +17,21 @@
               "required": true,
               "mapping": {
                 "transport": { "type": "str", "required": true },
-                "mcast_addr": { "type": "str", "required": true },
-                "mcast_port": { "type": "int", "required": true },
                 "password": { "type": "str", "required": true },
-                "require_clean_for_autostart_wrapper": { "type": "str", "required": true, "pattern": "/^true$|^false$|^auto$/" }
+                "require_clean_for_autostart_wrapper": { "type": "str", "required": true, "pattern": "/^true$|^false$|^auto$/" },
+                "rings": {
+                  "type": "seq",
+                  "required": true,
+                  "sequence": [ {
+                    "type": "map",
+                    "required": true,
+                    "mapping": {
+                      "network": { "type": "str", "required": true },
+                      "mcast_addr": { "type": "str", "required": true },
+                      "mcast_port": { "type": "int", "required": true }
+                    }
+                  } ]
+                }
               }
             },
             "crm": {
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -56,37 +56,47 @@ class PacemakerService < ServiceObject
     end
   end
 
+  def build_used_mcast_addrs(proposal_id, role_name)
+    used_mcast_addrs = {}
+
+    # iterate proposals, skip current proposal by ID
+    proposals_raw.each do |p|
+      next if p["id"] == proposal_id
+      p["attributes"][@bc_name]["corosync"]["rings"].each do |ring|
+        used_mcast_addrs[ring["mcast_addr"]] = true
+      end
+    end
+
+    # iterate roles, skip current role by name
+    RoleObject.find_roles_by_name("pacemaker-config-*").each do |r|
+      next if r.name == role_name
+      r.default_attributes["pacemaker"]["corosync"]["rings"].each do |ring|
+        used_mcast_addrs[ring["mcast_addr"]] = true
+      end
+    end
+
+    used_mcast_addrs
+  end
+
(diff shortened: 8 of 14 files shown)


scottwulf:
[Double Corosync ring](https://trello.com/c/pC8UY2Si/31-double-corosync-ring)

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -651,6 +656,13 @@ def validate_proposal_after_save proposal
       )
     end
 
+    ring0_network = proposal["attributes"][@bc_name]["corosync"]["ring0_network"]
+    unless ring0_network == "admin"
```
vuntz: Any reason to enforce this?

chef/data_bags/crowbar/template-pacemaker.schema
```diff
@@ -17,8 +17,9 @@
               "required": true,
               "mapping": {
                 "transport": { "type": "str", "required": true },
-                "mcast_addr": { "type": "str", "required": true },
-                "mcast_port": { "type": "int", "required": true },
+                "ring0_network": { "type": "str", "required": true },
+                "ring0_mcast_addr": { "type": "str", "required": true },
+                "ring0_mcast_port": { "type": "int", "required": true },
```
vuntz: Instead of having `ring0_*` attributes (and then `ring1_*` later on), why not just have an array?

crowbar_framework/app/models/pacemaker_service.rb
scottwulf: I am exposing this value in the UI to be consistent with the UI of future rings. It can be changed to editable. Two questions: (1) Should it be editable? (2) Should any two rings be allowed on the same network?

chef/data_bags/crowbar/template-pacemaker.schema
scottwulf: I was saving that for a later commit where the schema contains an array of rings.

crowbar_framework/app/models/pacemaker_service.rb
vuntz: Two rings should not be sharing the same network, so definitely something to validate. (Also, all nodes in the cluster should have an IP in the ring networks -- should be validated too)

And unless there's a good reason to enforce the admin network, I'd make it editable.

scottwulf: Yes all of that validation will be covered. I do not know who makes the editable decision but just in case I have asked Mate the same question.

chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
```diff
@@ -112,9 +117,11 @@ logging {
 
 <% if @transport == 'udpu' -%>
 nodelist {
-  <% @members.sort.each do |memberaddr| -%>
+  <% @rings[0][:members].each.with_index do |memberaddr, member_index| -%>
```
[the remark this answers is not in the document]
Itxaka: Arrays have a `first` and `last` method that provides the first and last member :)

aspiers:
Please can you open up the Trello board to everyone in the SUSE organisation, or at very least to me? I'm not allowed to view it currently.

I don't have time to review the code fully right now, but I wanted to nitpick the commit message:

> For increased availability, people want the ability to configure multiple Corosync rings so that if one network fails nodes do not become fenced.

To me that implies that multiple rings is the only way to obtain redundancy of the network layer, but that's not true.  We already support network bonding which provides the same level of redundancy.  IIRC the main motivation for supporting multiple rings was to accommodate customers who do not have enough NICs to bond, in which case multiple rings is a "poor man's" workaround.  Just wanted to make sure we are all on the same page about the motivation for this change before it gets merged.

Also please note that anything more than 2 rings is definitely not supported.  This was confirmed at the Clusterlabs Summit last week.  I only had time to scan the code quickly, but I couldn't see anything preventing 3 or more rings - if that's really the case then please change it to ensure it does not allow more than 2 rings.  Thanks!

matelakat:
> or at very least to me?

done

matelakat:
> To me that implies that multiple rings is the only way to obtain redundancy of the network layer, but that's not true. We already support network bonding which provides the same level of redundancy. IIRC the main motivation for supporting multiple rings was to accommodate customers who do not have enough NICs to bond, in which case multiple rings is a "poor man's" workaround. Just wanted to make sure we are all on the same page about the motivation for this change before it gets merged.

Actually the original request came from a setup where we'll have two bonds

aspiers:
Ah, so bonded networks but no redundancy on the switches or something like that?

scottwulf:
Changes in latest commit:
* change ring0 to editable
* sort member addresses to avoid triggering corosync reload
* include schema migration script
To do:
* add a warning post-apply that networks should not be changed
* maybe: combine ring0 and ring1 into an array sooner than config.rb
* introduce ring1

chef/cookbooks/corosync/recipes/config.rb
```diff
@@ -34,10 +34,24 @@
...
+rings = [
+  {
+    bind_addr: node[:corosync][:ring0_bind_addr],
+    mcast_addr: node[:corosync][:ring0_mcast_addr],
+    mcast_port: node[:corosync][:ring0_mcast_port],
+    members: node[:corosync][:ring0_members].sort
+  }
+]
+members_v2 = [
+  node[:corosync][:ring0_members].sort
```
matelakat: Maybe we should use `rings[]['members']` here, so we don't need to explicitly enlist all members and re-do the sort? This way when we add the second ring, we just add another dictionary to rings and job done.

chef/cookbooks/corosync/templates/default/corosync.conf.erb
```diff
@@ -45,21 +45,23 @@ totem {
         # This specifies the mode of redundant ring, which may be none, active, or passive.
         rrp_mode: none
 
+<% @rings.each.with_index do |ring, ring_index| -%>
         interface {
-                ringnumber:     0
-                bindnetaddr:    <%= @bind_addr  %>
+                ringnumber:     <%= ring_index %>
```
matelakat: Where do we use `corosync.conf.erb` Maybe we should get rid of it as a whole? Maybe it no longer makes sense in a SOC7/8 environment? Need to check the code in what situations is it used.

matelakat: Yeah, that can go I think:
```
template "/etc/corosync/corosync.conf" do
  if node[:platform] == "suse" && node[:platform_version].to_f < 12.0
    source "corosync.conf.erb"
  else
    source "corosync.conf.v2.erb"
  end
```

matelakat: @aspiers Any other reasons to keep the `corosync.conf.erb` ?

chef/cookbooks/corosync/recipes/config.rb
```diff
@@ -34,10 +34,24 @@
...
 
+rings = [
+  {
+    bind_addr: node[:corosync][:ring0_bind_addr],
+    mcast_addr: node[:corosync][:ring0_mcast_addr],
+    mcast_port: node[:corosync][:ring0_mcast_port],
+    members: node[:corosync][:ring0_members].sort
+  }
+]
+members_v2 = [
```
matelakat: this `members_v2` name is quite cryptic for me. If I understand it correctly this will be a list of lists, and the inner list will be representing the addresses on the rings. What about `list_of_node_addresses` ?

```ruby
nodelist {
  <% @list_of_node_addresses do |node_addresses| -%>
	node {
    <% node_addressses.each.with_index do |ring_addr, ring_index| -%>
		ring<%= ring_index %>_addr: <%= ring_addr %>
    <% end -%>
	}
  <% end -%>
}
```

chef/cookbooks/corosync/recipes/config.rb
matelakat: looking at a bit further it turned out that `rings[]['members']` is only used by the legacy config, so if we get rid of that, we'll need to enlist the members here. I guess when you transition into storing the data in an array, it will be easier to iterate over that.

chef/data_bags/crowbar/migrate/pacemaker/033_fully_qualify_ring0.rb
```diff
@@ -0,0 +1,17 @@
+def upgrade(ta, td, a, d)
+  a["corosync"]["ring0_network"] = "admin"
+  a["corosync"]["ring0_bind_addr"] = a["corosync"]["bind_addr"]
+  a["corosync"]["ring0_mcast_addr"] = a["corosync"]["mcast_addr"]
+  a["corosync"]["ring0_mcast_port"] = a["corosync"]["mcast_port"]
+  a["corosync"]["ring0_members"] = a["corosync"]["members"]
```
matelakat: you want to delete the non-used items as well, like `a["corosync"]["bind_addr"]`

chef/cookbooks/corosync/recipes/config.rb
matelakat: We already discussed with @scottwulf that `members_v2` is a data structure for configuration file v2.

chef/cookbooks/corosync/recipes/config.rb
matelakat: Until we don't get an answer if anyone is using v1, we keep that. so I'm ok with it.

chef/cookbooks/corosync/templates/default/corosync.conf.erb
aspiers: I can't think of one; I guess it was only done like this by @vuntz during a transition period when we wanted to support both SLES 11 and 12 in the same codebase.  @vuntz is there any other reason?

vuntz:
> I can't think of one; I guess it was only done like this by @vuntz during a transition period when we wanted to support both SLES 11 and 12 in the same codebase. @vuntz is there any other reason?

No, that's the only reason.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -651,6 +656,27 @@ def validate_proposal_after_save proposal
...
+        ring_network = nil
+      end
+      unless ring_network
+        validation_error I18n.t(
+          "barclamp.#{bc_name}.validation.ring_network_notfound",
+          ring_network: ring0_network,
+          ring_ordinal: "first"
+        )
+      end
+    end
```
vuntz: You also need some validation that each node in the cluster has an address in the network.

vuntz: Oh, and btw, I'm not really sure that loading the databag is the best way to validate that the network exists. I think I'd prefer if we'd look at the network applied proposal, if the network is there.

crowbar_framework/config/locales/pacemaker/en.yml
```diff
@@ -64,6 +64,7 @@ en:
             down or rebooted; you will have to manually start it after having
             fixed the issue. "Automatic" means that this setting will be set to
             "true" for two-nodes cluster, and to "false" otherwise.'
+          ring0_network: 'Network to use for the first corosync ring (required)'
```
vuntz: No need to pure `(required)`, imho. The validation will make it clear, and it's pre-filled.

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -5,8 +5,9 @@
     "pacemaker": {
       "corosync": {
         "transport": "udpu",
-        "mcast_addr": "239.255.0.1",
-        "mcast_port": 5405,
+        "ring0_network": "admin",
+        "ring0_mcast_addr": "",
```
vuntz: Any reason to change the default here?

scottwulf:
Changes in latest commit:
* change ring0 to editable
* sort member addresses to avoid triggering corosync reload
* include schema migration script
* combine ring0 (to ringN) into an array for corosync as soon as possible

To do:
* add a warning post-apply that networks should not be changed
* introduce ring1
* optional: change network exists code

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -5,10 +5,15 @@
...
         "transport": "udpu",
-        "mcast_addr": "239.255.0.1",
-        "mcast_port": 5405,
         "password": "crowbar",
-        "require_clean_for_autostart_wrapper": "auto"
+        "require_clean_for_autostart_wrapper": "auto",
+        "rings": [
+          {
+            "network": "admin",
+            "mcast_addr": "",
```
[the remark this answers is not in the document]
scottwulf: Many reasons for removal:
* Unlike before, a multicast address is reserved only when necessary (i.e. transport is UDP).
* Code has been changed to behave based on current address value. If a value is supplied, the code assumes it is user provided and simply validates it. If a value is not supplied, the code calculates the next available address.
* Validation code has been added to ensure unique multicast addresses and these addresses are not editable in the crowbar UI (only in raw mode). If the user does not edit the raw data, validation will raise an error on proposals 2 through n.

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -55,7 +60,7 @@
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 200,
+      "schema-revision": 210,
```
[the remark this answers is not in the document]
scottwulf: I looked through the revision history and +10 was the most common increment. The value does not matter to me.

chef/cookbooks/corosync/recipes/config.rb
```diff
@@ -34,9 +34,31 @@
...
+    if ring[:members].nil? || ring[:members].empty?
+      raise "Members have to be defined when using \"udpu\" transport!"
+    end
+  end
+end
+
+# set bindnetaddr per node, use host address rather than network subnet
+# in case multiple interfaces are configured on the same subnet
+rings.each do |ring|
+  ring[:bind_addr] = Barclamp::Inventory.get_network_by_type(node, ring["network"]).address
```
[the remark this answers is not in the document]
scottwulf: I cannot. The bind address(es) cannot be determined until after the number of rings and their networks has been determined. I will use other means of obtaining the address.

chef/cookbooks/corosync/recipes/firewall.rb
```diff
@@ -17,14 +17,20 @@
 # limitations under the License.
 #
 
+mcast_ports = node[:corosync][:rings].map { |n| n["mcast_port"].to_i }
+if node[:corosync][:transport] == "udp"
+  mcast_ports += mcast_ports.map { |n| n - 1 }
```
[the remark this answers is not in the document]
scottwulf: Copied directly from the corosync.conf (and man page) below. For additional references, search online for corosync 5404.

Corosync uses the port you specify here for UDP messaging, and also the immediately preceding port. Thus if you set this to 5405, Corosync sends messages from UDP port 5404 to UDP port 5405.

chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
```diff
@@ -110,15 +115,15 @@ logging {
 	}
 }
 
-<% if @transport == 'udpu' -%>
```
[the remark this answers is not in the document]
scottwulf: UDP was completely broken prior to this pull request. The corosync service failed to start. The nodelist is required regardless of transport else one receives the error "configuration error: nodelist or quorum.expected_votes must be configured".

chef/cookbooks/corosync/recipes/config.rb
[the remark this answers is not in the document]
scottwulf: I searched cookbooks on my admin node and found 87 instances of Barclamp::Inventory and 0 of anything else. Does a better method exist?

chef/data_bags/crowbar/template-pacemaker.json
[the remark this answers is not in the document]
vuntz: Actually we used to do +10 (and then +100) when moving to a new major version. +1 is the standard thing.

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -20,7 +20,6 @@
 require "timeout"
 
 node.normal[:corosync][:cluster_name] = CrowbarPacemakerHelper.cluster_name(node)
-node.normal[:corosync][:bind_addr] = Barclamp::Inventory.get_network_by_type(node, "admin").address
```
vuntz:
```ruby
node[:corosync][:rings].each_with_index do |ring, index|
  node.normal[:corosync][:rings][index][:bind_addr] = Barclamp::Inventory.get_network_by_type(node, ring[:network]).address
done
```

chef/data_bags/crowbar/migrate/pacemaker/210_add_ring_array.rb
```diff
@@ -0,0 +1,35 @@
+def upgrade(ta, td, a, d)
```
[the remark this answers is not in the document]
vuntz: So the way I would approach this is as follows:

```ruby
def upgrade(ta, td, a, d)
[...]
  # Are we a proposal? If yes, then look for the applied role if it exists and modify it
  # (the pacemaker barclamp adds an required_post_chef_calls attribute to the
  # deployment part of the role, so we can use that to know if this is the proposal or not)
  # Note that we don't do this step when migrating the role because this would imply
  # loading/saving the role, while the migration scripts will save the role later on
  # with some data that was already loaded before, therefore outdated, and that
  # would break the migration
  unless d.key?("required_post_chef_calls")
    begin
      role = Chef::Role.load(d["config"]["environment"])
      role[:corosync][:rings] = [ring]
      role.save
    rescue Net::HTTPServerException => e # node does not exists on the chef server
      raise e if e.response.code != "404"
    end
  end

  return a, d
end
```

This requires testing, of course...

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -341,14 +411,31 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+      # allocate member addresses
+      ring["members"] = allocate_member_addresses(member_nodes, ring["network"])
+
+      next unless role.default_attributes["pacemaker"]["corosync"]["transport"] == "udp"
+
+      # build a hash of used mcast_addrs
+      used_mcast_addrs = build_used_mcast_addrs(proposal.id, role.name) if used_mcast_addrs.nil?
+
+      # validate specified mcast_addr if specified else allocate one
+      ring["mcast_addr"] = allocate_mcast_addr(used_mcast_addrs, ring["mcast_addr"])
```
vuntz: I don't get that part: the multicast address is part of the propoal, so we should not do anything here. We should only validate (in `validate_proposal_after_save`) that the multicast address is not really used elsewhere.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -60,32 +60,6 @@ def create_proposal
...
-          base["attributes"][@bc_name]["corosync"]["mcast_addr"] = mcast_addr
-          free_mcast_addr_found = true
-          break
-        end
-      end
-      break if free_mcast_addr_found
-    end
-
-    raise "Cannot find an available multicast address!" unless free_mcast_addr_found
-
```
vuntz: I would not have removed that, this is useful to populate the proposal with a valid multicast address for the first ring when the proposal is created.

chef/cookbooks/corosync/recipes/firewall.rb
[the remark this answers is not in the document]
Itxaka: oh wow. Can we get a comment about that in the code? This looks like something we forget about and bites us down the road.

chef/cookbooks/corosync/recipes/config.rb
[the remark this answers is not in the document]
aspiers: @vuntz is right: any code tying Crowbar and Pacemaker together belongs in the `crowbar-pacemaker` cookbook, not this one.

> The bind address(es) cannot be determined until after the number of rings and their networks has been determined. I will use other means of obtaining the address.

The address(es) can only come from Crowbar; nothing else in the code knows enough about the networks and IP ranges available.  AFAICS that means the addresses / networks / number of rings need to be determined in the `crowbar-pacemaker`, and then stored in a way (probably Chef attributes) which can be accessed by this `corosync` cookbook.

`Barclamp::Inventory` is the right library to use.

chef/cookbooks/corosync/templates/default/corosync.conf.erb
aspiers: @vuntz ping?

chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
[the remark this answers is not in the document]
aspiers: Oh dear ;-/  I guess we need to include this in our CI then.  Rather than adding a whole new dimension to the HA test matrix, maybe it's sufficient to build a separate PR to the automation repo which makes the HA CI jobs randomly choose unicast or multicast?

chef/cookbooks/corosync/templates/default/corosync.conf.erb
vuntz: Sorry, forgot to reply. We can kill that, yes. But at this point, I'd prefer if it's done after this pull request as this one is needed really soon.

Itxaka:
My last comment does not apply anymore. Seems like applying the diff manually and recompiling the assets somehow broke it, while deploying the package with UPDATEREPOS makes it work as expected.

Itxaka:
Tested and it works as expected, pretty cool @scottwulf !

```
root@d52-54-77-77-01-01:~ # corosync-cfgtool -s
Printing ring status.
Local node ID 1084812625
RING ID 0
	id	= 192.168.237.81
	status	= ring 0 active with no faults
RING ID 1
	id	= 10.165.200.2
	status	= ring 1 active with no faults
```

The only thing I can see its that my `bindaddr` for all rings appears empty when using udpu, its ok for the config file to have that attribute empty or should we do not write it if we are using updu?

example:
```
	interface {
		# Rings must be consecutively numbered, starting at 0.
		ringnumber: 1
		# This is normally the *network* address of the
		# interface to bind to. This ensures that you can use
		# identical instances of this configuration file
		# across all your cluster nodes, without having to
		# modify this option.
		bindnetaddr: 
		# However, if you have multiple physical network
		# interfaces configured for the same subnet, then the
		# network address alone is not sufficient to identify
		# the interface Corosync should bind to. In that case,
		# configure the *host* address of the interface
		# instead:
		# bindnetaddr: 192.168.1.1
		# When selecting a multicast address, consider RFC
		# 2365 (which, among other things, specifies that
		# 239.255.x.x addresses are left to the discretion of
		# the network administrator). Do not reuse multicast
		# addresses across multiple Corosync clusters sharing
		# the same network.
		# Corosync uses the port you specify here for UDP
		# messaging, and also the immediately preceding
		# port. Thus if you set this to 5405, Corosync sends
		# messages over UDP ports 5405 and 5404.
		mcastport: 5405
		# Time-to-live for cluster communication packets. The
		# number of hops (routers) that this ring will allow
		# itself to pass. Note that multicast routing must be
		# specifically enabled on most network routers.
		ttl: 1
	}

	transport: udpu
```

Itxaka:
btw, the small touch of providing the multicast available addresses on the error hint is really good!

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -20,7 +20,13 @@
...
+
+# set bindnetaddr per node, use host address rather than network subnet
+# in case multiple interfaces are configured on the same subnet
+node[:corosync][:rings].each_with_index do |ring, index|
+  node.normal[:corosync][:rings][index][:bind_addr] =
```
Itxaka: umm, this does not seem to work, it ends up empty on my env

Itxaka: So if instead of doing this you do:
```
node[:corosync][:rings].each do |ring|
  ring[:bind_addr] = Barclamp::Inventory.get_network_by_type(node, ring[:network]).address
end
```

it seems to work. chef ¬_¬

PR #241, opened 2017-09-14
pacemaker: Add missing  operations to the parser

rhafer:
The parser for the "op" part of a resource definition was missing the
"promote" and "demote" operations. This was never noticed up to now as
we didn't use them until we introduced galera support in the database
barclamp. This bug caused calls to "crm config" in every chef-client run
for resources that contain custom settings for the "promote" or "demote"
operation.

### DIFF
--- a/chef/cookbooks/pacemaker/libraries/pacemaker/resource/primitive.rb
+++ b/chef/cookbooks/pacemaker/libraries/pacemaker/resource/primitive.rb
@@ -35,7 +35,7 @@ class Pacemaker::Resource::Primitive < Pacemaker::Resource
     end
 
     self.op = {}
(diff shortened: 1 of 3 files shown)


jsuchome:
> This bug caused calls to "crm config" in every chef-client 

Did it cause any errors? Performance impact?

rhafer:
> Did it cause any errors? Performance impact?

Can't say for sure. Probably not. But it caused misleading message in the chef log about the cluster config being updated. So, to a large degree this is a cosmetic change.

rhafer:
Rebased to force a package rebuild for gating after https://github.com/crowbar/crowbar-openstack/pull/1233 and https://github.com/crowbar/crowbar-ha/pull/228 where merged

PR #247, opened 2017-09-19
[4.0] Reimplement sync marks without node searches

Itxaka:
This is a continuation on the job started by @vuntz here: #170

The main benefit here is that we don't depend on the chef server
(through searches), which should decrease the load on the admin server
but also make things faster.

There's a slight difference in behavior for the sync marks: only the
founder nodes is waiting for all other nodes on sync; the non-founder
nodes simply wait for the founder because the founder will only set the
mark once all other nodes have already set their mark.

The initial sync mark (about the founder node being setup) is replaced
by a simple attribute on the node, in a similar way that we mark remote
nodes having been setup.

It's worth mentioning that we don't use revisions but only booleans for
the sync marks: revisions are deprecated.

Finally, we convert pre-existing sync marks to this model, which should
make this change non-disruptive.

Backport of #217


Itxaka:

abel-navarro:
backport of https://github.com/crowbar/crowbar-ha/pull/217

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -54,7 +54,7 @@
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 100,
+      "schema-revision": 200,
```
vuntz: This needs to be 101 in the backport. And btw, the migration in master probably needs to be reworked because I don't think it will work if migration already happened before (in stable/4.0).

Itxaka:
@vuntz I am currently testing on a clean env, will move to an existing env afterwards.

chef/data_bags/crowbar/template-pacemaker.json
Itxaka: ah, good catch

aspiers:
@Itxaka If this a direct backport of #217, please can you include that info in the PR description so that it's easily visible, as per line 33 of https://etherpad.nue.suse.com/p/cloud-PR-porting?  Thanks!

Itxaka:
@vuntz @aspiers just manually deployed a cloud7 env, deployed barclamps on a cluster up to neutron, then applied the updated rpm generated by the testbuild.

 - migration run successfully
 - role had the proper founder
 - node attribute pacemaker.founder was empty
 - applying the barclamps again resulted into a successful deploy. The first run properly set up the cluster.

Itxaka:
Only nit that I found is that the  logs are getting the output of the crm attribute queries for some reason. Maybe using the `Regexp.last_match(1)` was not such a great idea and that is polluting the logs :/

Will try a follow up pr to master+4.0 to try to fix that output

PR #248, opened 2017-09-20
[4.0] pacemaker: Add option to stop managing stateless active/active services

vuntz:
Pacemaker is not strictly required for stateless active/active services,
since orchestration, fencing, and so on is not necessarily useful in
that case.

Disabling this option does simplify the setup in the OpenStack case quite
significantly, but comes with a cost, like pacemaker not restarting
services that have failed.

The default is to keep the same behavior as in the past. This may change
when we implement Restart= for systemd services and if it found that
this approach has no negative impact.

(cherry picked from commit ada1d41565c6a08a3d499f141fae7cb5c45fd052)

Backport of https://github.com/crowbar/crowbar-ha/pull/238

### DIFF
--- a/crowbar_framework/config/locales/pacemaker/en.yml
+++ b/crowbar_framework/config/locales/pacemaker/en.yml
@@ -64,6 +64,15 @@ en:
             down or rebooted; you will have to manually start it after having
             fixed the issue. "Automatic" means that this setting will be set to
             "true" for two-nodes cluster, and to "false" otherwise.'
+        clone_stateless_services: 'Manage stateless active/active services with Pacemaker'
+        clone_stateless_services_hint:
+          'Stateless active/active services can be managed by Pacemaker, or can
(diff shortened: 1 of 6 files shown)


skazi0:
@vuntz any idea why the gating failed? it complains about rabbitmq not running.

crowbar_framework/config/locales/pacemaker/en.yml
```diff
@@ -64,6 +64,15 @@ en:
             down or rebooted; you will have to manually start it after having
             fixed the issue. "Automatic" means that this setting will be set to
             "true" for two-nodes cluster, and to "false" otherwise.'
+        clone_stateless_services: 'Manage stateless active/active services with Pacemaker'
+        clone_stateless_services_hint:
```
jsuchome: `clone_stateless_services_hint_html` is required by `_edit_attributes.html.haml`

PR #254, opened 2017-09-21
pacemaker: drop migration for cluster founder

Itxaka:
AS this migration has already been backported to 4.0, there is no need
to have it on master as any upgrades to 5.0 will have that migration
already and new installs dont need it.


Itxaka:

vuntz:
> AS this migration has already been backported to 4.0, there is no need

s/AS/As/

PR #256, opened 2017-10-03
Allow haproxy pacemaker group to be extended

jhesketh:


jsuchome:
This might not be needed after all

jhesketh:
Yep, abandoning it for now.

PR #260, opened 2017-10-18
[4.0] crowbar-framework: fix is_pacemaker flag for RestartManager

Itxaka:
Due to the ordering in here we were only setting the
is_pacemaker_service flag for the RestartManager in case the service was
set with use_crm_resource. That caused the real service to restart
always as we were proxying the call to the normal restart and also
caused the naming of the restart request to be mangled. As a side
effect, no matter what was happening, we were setting the node in
maintenance mode.

Instead, we move the check for allowing the restart to the first line so
we catch the restart the first thing, we save the restart with the
proper name and we avoid setting the node into maintenance mode if we
are not restarting.

(cherry picked from commit 19fe3ccb6a076906f711b0d8f0db278db13bd09f)

Backport-of: https://github.com/crowbar/crowbar-ha/pull/259/

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/providers/service.rb
+++ b/chef/cookbooks/crowbar-pacemaker/providers/service.rb
@@ -187,13 +187,12 @@ action :restart do
   )
 
   if service_is_running?(service_name, use_crm_resource, pacemaker_resource)
-    set_maintenance_mode unless no_maintenance_mode
-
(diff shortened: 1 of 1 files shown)


Itxaka:
[Chef restarting services - Issues](https://trello.com/c/3mpxiYym/175-chef-restarting-services-issues)

PR #262, opened 2017-10-20
Added support for op defaults

AbelNavarro:
**pacemaker: added support for op defaults**

on-fail" action.

other resourced.

be extended later on.

This change by itself does not fullfill the requirement. It needs the
'pacemaker-primitive' to read the value we store here.

**pacemaker: primitive to read global op defaults**

The pacemaker_primitive object has been extended in order to add support
for global 'op monitor on-fail' defaults. In this commit we consider the
barclamp schema already supports an option to configure the defaults for
the 'op monitor on-fail' action.

This code required the 'op monitor on-fail' attribute to be configured
as a Chef::Node::Attribute class. The main reason is that this class
keeps the original values configured in the barclamp as well as the new
ones, so it is possible to roll-back changes.

Some barclamps set this attribute as a Hash object. In that case there's
no possible rollback. The behavior in this case will be to not modify
it, in order not to break anything. The user can still manually set the
values in that particular barclamp to suit their needs, but not as a
global option.

### DIFF
--- a/chef/cookbooks/pacemaker/providers/primitive.rb
+++ b/chef/cookbooks/pacemaker/providers/primitive.rb
@@ -69,7 +69,61 @@ def create_resource(name)
   standard_create_resource
 end
 
+def op_defaults(node)
+  return nil unless node[:pacemaker][:founder]
+  return nil unless !node[:pacemaker].nil? && !node[:pacemaker][:op_defaults].nil?
+
+  node[:pacemaker][:op_defaults].to_hash
+end
+
 def update_resource(name)
+  ops = new_resource.op
+  op_defaults = op_defaults(node)
+
+  # If op_defaults is defined and not nil, set the value
+  # otherwise remove it
+  if !op_defaults.nil? &&
+      op_defaults.key?("monitor") &&
+      op_defaults["monitor"].key?("on-fail") &&
+      !op_defaults["monitor"]["on-fail"].nil? &&
+      !op_defaults["monitor"]["on-fail"].empty?
+
+    if ops.is_a? Hash
+      # Ops was defined as a Hash: we will skip the on-fail value
+      # should it be declared. In a Hash there's no way to know the original
+      # value, so it's either overwrite or skip.
+
+    elsif ops.key?("monitor")
+      # Ops was defined as a Chef::Node::Attribute. Store the on-fail
+      # default value.
+      ops["monitor"] = op_defaults["monitor"]
+    end
+
+  else
+    if ops.is_a? Hash
+      # Ops was defined as a Hash: we will skip the on-fail value
+      # should it be declared. In a Hash there's no way to know the original
+      # value, so it's either overwrite or skip.
+
+    else
+      # Ops was defined as a Chef::Node::Attribute. Store the on-fail
+      # default value.
+
+      # The 'current_normal' nesting level contains the values we have set
+      # in the previous section. The 'current_default' nesting level contains
+      # those values set in the recipe's attributes and are unchanged by
+      # this code
+      if ops.current_normal.key?("monitor")
+        monitor = ops.current_normal.fetch("monitor")
+        if monitor.key?("on-fail")
+          monitor.delete("on-fail")
+          ops["monitor"] = monitor
+        end
+      end
+    end
+
+  end
+
   current_agent = @current_resource.agent
   unless current_agent.include? ":"
     current_agent = "ocf:heartbeat:" + current_agent
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -54,7 +54,12 @@
         "admin_name": "",
         "public_name": ""
       },
-      "clone_stateless_services": true
+      "clone_stateless_services": true,
+      "op_defaults": {
+        "monitor": {
+          "on-fail": ""
+        }
+      }
     }
   },
   "deployment": {
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -398,4 +398,15 @@ module CrowbarPacemakerHelper
 
     existing_resources
   end
+
+  def self.op_defaults(node)
+    return nil unless is_cluster_founder?(node)
+
+    if node[:pacemaker].nil? || node[:pacemaker][:op_defaults].nil?
+      return nil
+    end
+
+    node[:pacemaker][:op_defaults].to_hash
+  end
+
 end
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -398,15 +398,4 @@ module CrowbarPacemakerHelper
 
     existing_resources
   end
-
-  def self.op_defaults(node)
-    return nil unless is_cluster_founder?(node)
-
-    if node[:pacemaker].nil? || node[:pacemaker][:op_defaults].nil?
-      return nil
-    end
-
-    node[:pacemaker][:op_defaults].to_hash
-  end
-
 end
--- a/chef/cookbooks/pacemaker/spec/providers/primitive_spec.rb
+++ b/chef/cookbooks/pacemaker/spec/providers/primitive_spec.rb
@@ -2,6 +2,8 @@ require "spec_helper"
 
 require_relative "../helpers/runnable_resource"
 require_relative "../fixtures/keystone_primitive"
+require_relative "../../../crowbar-pacemaker/libraries/helpers.rb"
+
 
 describe "Chef::Provider::PacemakerPrimitive" do
   # for use inside examples:
--- a/chef/data_bags/crowbar/migrate/pacemaker/201_op_defaults.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/201_op_defaults.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["op_defaults"] = ta["op_defaults"] unless a.key? "op_defaults"
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("op_defaults") unless ta.key? "op_defaults"
+  return a, d
+end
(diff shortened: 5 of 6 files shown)


abel-navarro:
Dependency: https://github.com/crowbar/crowbar-openstack/pull/1387

abel-navarro:

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -48,14 +48,19 @@
         "admin_name": "",
         "public_name": ""
       },
-      "clone_stateless_services": true
+      "clone_stateless_services": true,
+      "op_defaults": {
+        "monitor": {
+          "on-fail": "block"
+	}
```
aspiers: Broken indentation here.

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -48,14 +48,19 @@
         "admin_name": "",
         "public_name": ""
       },
-      "clone_stateless_services": true
+      "clone_stateless_services": true,
+      "op_defaults": {
+        "monitor": {
+          "on-fail": "block"
```
aspiers: Shouldn't the default be `fence`?

abel-navarro: Ah, good point. And that made me think: without an explicit value `restart` is the default without STONITH and `fence` with STONITH. With this approach we are setting the same value for both cases. I'm thinking this should be nil in order to be backwards compatible.

chef/data_bags/crowbar/template-pacemaker.json
abel-navarro: Will fix.

abel-navarro:
needs a final check and cleaning up travis issues.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -70,6 +70,56 @@ def create_resource(name)
 end
 
 def update_resource(name)
+  ops = new_resource.op
+  op_defaults = CrowbarPacemakerHelper.op_defaults(node)
```
vuntz: We should not call something from the crowbar-pacemaker cookbook here.

Can we simply move this helper method to the pacemaker cookbook?

abel-navarro: agree, I think I can move it to the pacemaker cookbook.

abel-navarro:
@vuntz the problem when changing the `op monitor` value when it's stored as a Hash is that you cannot revert it back. You can do that when it is stored as a `Chef::Node::Attribute` (fron now on just CNA).

In a CNA you have [4 levels](https://docs.chef.io/attributes.html) in which to store information (from higher priority to lower]:

* automatic
* override
* normal
* default

The value defined in the barclamp gets stored in the `default` level. I.e. in [Neutron](https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/attributes/default.rb#L237) we set it to "10s", so in the CNA this would look like:

* automatic = {}
* override = {}
* normal = {}
* default = { monitor => { "interval" => "10s" }}

When I do `ops["monitor"] = op_defaults["monitor"]` this looks like internally:

* automatic = {}
* override = {}
* normal = { monitor => { "on-fail" => "fence" }}
* default = { monitor => { "interval" => "10s" }}

But CNA merges all those levels and flattens them out to something like:

`{ monitor => { "interval" => "10s", "interval" => "10s" }}`

Should the user not want a global default for `on-fail` just removing it works. Even if the `default` level has an `on-fail` definition as well as the `normal` level, the normal level takes precedence, and when removed goes back to default: the original barclamp value.

When values are stored as a Hash (I'd have to check again which barclamps do so) we do not have a way back to the original barclamp value. The moment we write that `on-fail` it gets into pacemaker CIB and the default value gets forgotten. Skipping those barclamps that store that value as a Hash is the only way to ensure there's a way back.

vuntz:
@abel-navarro the part that I don't get is that the default is actually in the code: next time the code will run, then we will have the same code, except that `default_op` won't be used, resulting in a different monitor operation -- and the LWRP will just update the resource in pacemaker accordingly.

abel-navarro:
@vuntz I understand your concerns. According to experimentation, executing `chef-client` several times in a row without any changes provide consistent results using op defaults.

The `:update` method in `pacemaker_primitive` always gets called when we use `pacemaker_primitive`. In that method we always read op defaults and apply them.

The first time we set the op default, it will notice the Pacemaker CIB information is different and will apply the changes. The subsequent times op defaults will match Pacemaker CIB information and no changes will be applied.

vuntz:
@abel-navarro I still don't understand the need for saving to normal, then checking back the default, and also checking if we have a hash or not.

Would it be possible to try just the very simple approach where you don't have all of this and simply use the new attribute you define if nothing is specified in the op passed to the resource?

abel-navarro:
@vuntz once you write a value to `new_resource.op["monitor"]` it stays there. The next time chef-client runs, `new_resource.op["monitor"]` will have that value you previously wrote so you cannot revert it.

@aspiers I think you were proposing something like what @vuntz said by merging on the existing hash the new default. I think the problem is still the same: `new_resource` gets modified. I already tried something similar in a past PR with bad results: https://github.com/crowbar/crowbar-openstack/pull/1387/files#diff-be2c96f68eefbe5a37c1685909b24421R31

aspiers:
@abel-navarro IIUC the fundamental problem is that you are treating a `Chef::Node::Attribute` as a writable data structure.  This is wrong; it is only there as a side-effect of passing data into the LWRP from attributes in other cookbooks.  The data flow should be strictly one-way, from the caller of the LWRP into the LWRP implementation.  If you need to modify the data within the LWRP (which you do in this case, because you want to merge the defaults), then it is your job to create a fresh data structure which is a simple nested `Hash`, or whatever other native Ruby data structure, but it should NOT have any `C::N::A` objects inside it.  So like I said yesterday, either build the data structure from scratch, or convert the inputs to native Hashes via `to_hash` (which is essentially a deep clone) and then do the merge on those new structures.

abel-navarro:
@aspiers the flattening out approach you suggest would deal with the `Hash` vs `C::N::A`  separation I implemented. But doesn't solve the biggest problem I have: once you modify the resource there's no way back.

I followed your suggestion and duplicated the resource object. I changed the desired `op` values in the duplicate so `new_resource` remained unchanged. But as soon as the resource is updated, the next chef-run will have `new_resource` with the updated values. This is because `Pacemaker::Resource::Primitive` inherits from `Pacemaker::Resource`  which inherits from `Pacemaker::CIBObject`. That means, whatever we write into the primitive, gets written in the CIB thus becomes part of the resource.

To put it in other words:

We want:
```
op = resource.op + merge(op_defaults)
```

And when we remove defaults we want:
```
op = resource.op + merge(nil)
```

The problem is `resource.op` already has `op_defaults` and there's no way to remove it.

This is the code sample that tries to work with the dup approach: 
https://github.com/abel-navarro/crowbar-ha/commit/2cb4bfec01af658bd85a18d9653f65ea57075c84

aspiers:
@vuntz I talked to @abel-navarro and I think we've cleared up the confusion now.

abel-navarro:
I tried to implement the proposed changes, but I keep hitting the same problem: once you update the pacemaker primitive, the new value becomes the used one.

I found that when the systems uses a Hash to store the values (i.e. apache2, postgresql) those are mainly the definition ones, so it is easy to implement the proposed approach: just merge the default if none was present. When `ops` are stored as a `Chef::Node::Attribute` there's no other way than differentiate that case a manually clear the stored values. I'm working on that approach [here](https://github.com/abel-navarro/crowbar-ha/tree/pacemaker-onfail-revisited).

In case you want to take a look at the try to get that proposed changes working, you can find them [here](https://github.com/abel-navarro/crowbar-ha/tree/pacemaker-onfail-neweyes)

vuntz:
Why not simply do something like:

```ruby
  op_defaults = op_defaults(node)

  if !op_defaults.nil? &&
      op_defaults.key?("monitor") &&
      op_defaults["monitor"].key?("on-fail") &&
      !op_defaults["monitor"]["on-fail"].nil? &&
      !op_defaults["monitor"]["on-fail"].empty?

    ops = new_resource.op.to_hash # actually, this may be wrong, we may need something that is is recursive -- you can write a small helper for this
    ops["monitor"] ||= {}
    ops["monitor"]["on-fail'] ||= op_defaults["monitor"]["on-fail"]
    new_resource.op = ops
  end
```

aspiers:
> once you update the pacemaker primitive, the new value becomes the used one.

I'm not sure what exactly this means, or why it would be a problem.  But I'm pretty sure [this way](https://github.com/crowbar/crowbar-ha/compare/master...abel-navarro:pacemaker-onfail-revisited) is much more complicated than it needs to be :)  What I've been trying to say is that the input data should be treated as immutable, in which case there's no need to check whether it's a `Hash` or a `C::N::A` - all you need to do is read the values from the data structure.  [Your other way](https://github.com/crowbar/crowbar-ha/compare/master...abel-navarro:pacemaker-onfail-neweyes) looks more sensible at a glance (and closer to what @vuntz suggests).  But at this point I really think it makes sense to pair program on this - it probably could have saved you several days worth of experimentation!

BTW @vuntz from memory I *think* `to_hash` works recursively on `Mash` objects.  But ICBW.

abel-navarro:
@aspiers I did a pair programming session with @vuntz last day before winter holidays. He discovered `new_resource` was not _clean_, meaning it did not only contain desired new state but also old state. The problem was traced to that old state being present in the node's chef information.

@vuntz suggested my old changes did (wrongly) store some information in the chef's node database. I have tried with a clean environment and there's some piece of code that stores that information, but I haven't found which, since it looks like a simple read triggers that.

I'd welcome a new pair session for this because the reason that node information written it is still unknown to me.

aspiers:
Let's try to make time for this tomorrow.

AbelNavarro:
Today I got invaluable help from @aspiers to solve the block I had here. @vuntz proposal was fundamentally right, but it failed in assigning `ops` to `new_resource` (`no such method op`).  @aspiers found that the proper way was `new_resource.op(ops)`.

My approach was flawed because I was duplicating `new_resource`, but the inner `op` object wasn't. Modifying that original `op` object meant the data ended up in Chef's node information. You could see the attribute written with `knife node show` when it was only intended to live inside the pacemaker primitive resource.

I have taken an alternate approach I had and applied the latest findings there in order to have a clean PR. You can find it here: https://github.com/crowbar/crowbar-ha/pull/297. Subsequent work should happen there, so I'm closing this PR.

PR #263, opened 2017-10-20
[4.0] Add support for multiple Corosync rings

scottwulf:
Customer has requested the ability to configure an additional corosync
ring to "fully leverage the second bond in their dual bond environment".
The pacemaker cluster nodes will have visibility of one another over
both bonds and, when a failure occurs on the first bond, a cluster
failure will be averted due to the existence of the second ring over
the second bond.

In order to accomplish this, the flat list of single ring attributes
and web UI form fields is replaced by an array and _**n**_ number of web UI
form fields to support _**n**_ number of rings. The UI and validation code
restrict the maximum number of rings to 2. If there is ever a need for
for more rings, simply search for '2', 'two', and 'second' and make
the appropriate changes.

Backport-of: https://github.com/crowbar/crowbar-ha/pull/239

### DIFF
--- a/chef/cookbooks/corosync/attributes/default.rb
+++ b/chef/cookbooks/corosync/attributes/default.rb
@@ -15,11 +15,14 @@
 
 default[:corosync][:cluster_name] = "hacluster"
 
-default[:corosync][:bind_addr]   = "192.168.124.0"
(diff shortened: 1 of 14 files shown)


scottwulf:
[Double Corosync ring](https://trello.com/c/pC8UY2Si/31-double-corosync-ring)

PR #266, opened 2017-10-26
Fix for pacemaker proposal migration failure

scottwulf:
This fix resolves the issue where the migration of pacemaker
proposals fail with the following error:

rake aborted!
Failed to migrate proposal data for pacemaker: error while executing migration script /opt/dell/chef/data_bags/crowbar/migrate/pacemaker/102_add_ring_array.rb:
undefined method `[]' for #<Chef::Role:0x00000008b3d618> 
/opt/dell/crowbar_framework/lib/schema_migration.rb:224:in `rescue in migrate_proposal'

### DIFF
--- a/chef/data_bags/crowbar/migrate/pacemaker/201_add_ring_array.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/201_add_ring_array.rb
@@ -1,24 +1,30 @@
-def upgrade(ta, td, a, d)
+def upgrade_to_multi_ring(corosync)
+  # skip if already upgraded
+  return if corosync["rings"]
+
+  # initialize ring with prior, hard-coded network
   ring = {
     "network" => "admin"
   }
-  if a["corosync"]["bind_addr"]
-    ring["bind_addr"] = a["corosync"]["bind_addr"]
-    a["corosync"].delete "bind_addr"
+
+  # move flat list of attributes into first ring
+  ring_attributes = [
+    "bind_addr",
+    "mcast_addr",
+    "mcast_port",
+    "members"
+  ]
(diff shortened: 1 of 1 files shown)


chef/data_bags/crowbar/migrate/pacemaker/201_add_ring_array.rb
```diff
@@ -30,7 +33,7 @@ def upgrade(ta, td, a, d)
   unless d.key?("required_post_chef_calls")
     begin
       role = Chef::Role.load(d["config"]["environment"])
-      role[:corosync][:rings] = [ring]
+      upgrade_to_multi_ring(role.default_attributes["corosync"])
```
vuntz: I'd be super careful about this, as I'm not sure it works as intended. I think I've been hit in the past when doing something like:

```ruby
a = role.default_attributes["corosync"]
a["rings"] = "foo"
```

instead of 

```ruby
role.default_attributes["corosync"]["rings"] = "foo"
```

This is due to how `role.default_attributes` is implemented.

vuntz: Also another issue: we must not do this if `role.default_attributes["corosync"]["rings"]` already exists. It will be the case if a user upgrades from stable/4.0 with the patch.

scottwulf: Nice catch. It is too bad the schema-revision values are different across branches.

scottwulf: Yes. Thank you for the test steps. This was explicitly confirmed.

PR #268, opened 2017-10-31
crowbar-framework: remove useless metadata

Itxaka:
recommends and suggests are useless in the metadata files as they dont
do anything and are deprecated[0]

[0] https://chef.github.io/chef-rfc/rfc085-remove-unused-metadata.html


Itxaka:
> Fine with it, even though I feel it provides some info even if unused in code.

Disagree there. There is only 2 possible states for cookbooks, is either required or its not. Having a recommends where the cookbook is not needed it makes no sense (plus it breaks some tools like berkshelf when dealing with dependencies, which may be helpful to use with test-kitchen for some cookbook deep testing 😉 )

PR #269, opened 2017-11-01
Haproxy attrs

jhesketh:
Add support for downinter and tune rise/fall params.

### DIFF
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -100,8 +100,8 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
         inter = " inter #{server[:inter] || 2000}"
         fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : ""
         downinter = server[:downinter] ? " downinter #{server[:downinter]}" : ""
-        rise = " rise #{server[:rise] || 2}"
-        fall = " fall #{server[:fall] || 5}"
+        rise = " rise #{server[:rise] || 5}"
+        fall = " fall #{server[:fall] || 2}"
         backup = server[:backup] ? " backup" : ""
       %>
         server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check<%= ssl %><%= inter %><%= fastinter %><%= downinter %><%= rise %><%= fall %><%= backup %>
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -99,11 +99,12 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
         ssl = content[:use_ssl] ? " check-ssl verify none" : ""
         inter = " inter #{server[:inter] || 2000}"
         fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : ""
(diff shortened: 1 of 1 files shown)


chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -99,8 +99,9 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
...
         inter = " inter #{server[:inter] || 2000}"
         fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : ""
-        rise = " rise #{server[:rise] || 2}"
-        fall = " fall #{server[:fall] || 5}"
+        downinter = server[:downinter] ? " downinter #{server[:downinter]}" : ""
+        rise = " rise #{server[:rise] || 5}"
+        fall = " fall #{server[:fall] || 2}"
         backup = server[:backup] ? " backup" : ""
       %>
         server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> check<%= ssl %><%= inter %><%= fastinter %><%= rise %><%= fall %><%= backup %>
```
jsuchome: Shouldn't you add `downinter` to `server` line?

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -99,8 +99,9 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
         ssl = content[:use_ssl] ? " check-ssl verify none" : ""
         inter = " inter #{server[:inter] || 2000}"
         fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : ""
-        rise = " rise #{server[:rise] || 2}"
-        fall = " fall #{server[:fall] || 5}"
+        downinter = server[:downinter] ? " downinter #{server[:downinter]}" : ""
+        rise = " rise #{server[:rise] || 5}"
+        fall = " fall #{server[:fall] || 2}"
```
vuntz: Can you elaborate on the rise/fall change: the commit message is only describing the change and doesn't explain why we should do the change. I get the fall part, but the rise part is a bit unclear to me.

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -99,6 +99,7 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
         ssl = content[:use_ssl] ? " check-ssl verify none" : ""
         inter = " inter #{server[:inter] || 2000}"
         fastinter = server[:fastinter] ? " fastinter #{server[:fastinter]}" : ""
+        downinter = server[:downinter] ? " downinter #{server[:downinter]}" : ""
```
aspiers: @jsuchome is right - this new variable isn't actually getting used by anything :)

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
jhesketh: Ah yes, I missed that. Thanks.

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
jhesketh: I'll expand in the commit as well. It is somewhat arbitrary, but basically my thinking is that we should be quick to change to another healthy host and slow to consider a returning host healthy.

jsuchome:
@aspiers do you still request a change?

PR #270, opened 2017-11-02
crowbar-pacemaker: Reset sync-marks for all nodes

rhafer:
Up to now we only reset the synchronisation marks for the founder node
when re-applying a barlcamp. The reset needs to happen for all nodes of
a cluster. Otherwise we will run into synchronisation issues if e.g. the
first deployment of a barclamp fails and all nodes except of the founder
still have the sync marks set.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
@@ -79,7 +79,11 @@ end
 
 if CrowbarPacemakerHelper.is_cluster_founder?(node) && node[:pacemaker][:reset_sync_marks]
   # we can't reset sync marks if the pacemaker stack is not set up...
-  CrowbarPacemakerSynchronization.reset_marks(node) if node[:pacemaker][:setup]
+  if node[:pacemaker][:setup]
+    CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
+      CrowbarPacemakerSynchronization.reset_marks(cluster_node)
+    end
+  end
   # ... but we don't want to block the other nodes forever
   node.set[:pacemaker][:reset_sync_marks] = false
   dirty = true


chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -78,9 +78,11 @@
...
 
 if CrowbarPacemakerHelper.is_cluster_founder?(node) && node[:pacemaker][:reset_sync_marks]
-  # we can't reset sync marks if the pacemaker stack is not set up...
-  CrowbarPacemakerSynchronization.reset_marks(node) if node[:pacemaker][:setup]
-  # ... but we don't want to block the other nodes forever
+  CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
+    # we can't reset sync marks if the pacemaker stack is not set up...
+    CrowbarPacemakerSynchronization.reset_marks(cluster_node) if cluster_node[:pacemaker][:setup]
+    # ... but we don't want to block the other nodes forever
+  end
```
vuntz: It's good, but the comments make no sense where they live now :-) Let me suggest:

```ruby
if CrowbarPacemakerHelper.is_cluster_founder?(node) && node[:pacemaker][:reset_sync_marks]
  # we can't reset sync marks if the pacemaker stack is not set up...
  if cluster_node[:pacemaker][:setup]
    CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
      CrowbarPacemakerSynchronization.reset_marks(cluster_node)
    end
  end
  # ... but we don't want to block the other nodes forever
  node.set[:pacemaker][:reset_sync_marks] = false
  dirty = true
end
```

rhafer: > but the comments make no sense where they live now

I must admit, they didn't make a whole lot of sense to me even before my patch. But please don't tell anyone :)

rhafer: > Let me suggest

Thanks. And adjusted.

chef/cookbooks/crowbar-pacemaker/recipes/default.rb
```diff
@@ -79,7 +79,9 @@
 
 if CrowbarPacemakerHelper.is_cluster_founder?(node) && node[:pacemaker][:reset_sync_marks]
   # we can't reset sync marks if the pacemaker stack is not set up...
-  CrowbarPacemakerSynchronization.reset_marks(node) if node[:pacemaker][:setup]
+  CrowbarPacemakerHelper.cluster_nodes(node).each do |cluster_node|
+    CrowbarPacemakerSynchronization.reset_marks(cluster_node) if cluster_node[:pacemaker][:setup]
+  end
```
vuntz: Heh, okay, slightly better, but... still confusing to me:

 - `# we can't reset sync marks if the pacemaker stack is not set up...` goes with `if cluster_node[:pacemaker][:setup]`
 - (I dislike having the `if` statement inside the loop because the loop is useless if the statement is fals)

rhafer: Thanks, I completely overlooked that part of you suggestion. Hope I got it right this time.

PR #272, opened 2017-11-04
corosync: remove nonsensical ring default

dirkmueller:
The default attributes are getting merged with the default from the
crowbar barclamp proposal, rather than overwritten, so we always
have two entries, one broken one. For upgraded installs this is
basically desastrously breaking the install, as found out in a live
demo. All debugging thanks to Adam.

### DIFF
--- a/chef/cookbooks/corosync/attributes/default.rb
+++ b/chef/cookbooks/corosync/attributes/default.rb
@@ -16,13 +16,7 @@
 default[:corosync][:cluster_name] = "hacluster"
 
 default[:corosync][:transport] = "udpu"
-default[:corosync][:rings] = [{
-  "network" => "admin",
-  "bind_addr" => "192.168.124.0",
-  "mcast_addr" => "239.1.2.3",
-  "mcast_port" => 5405,
-  "members" => ["192.168.124.0"]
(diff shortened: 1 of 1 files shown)


aspiers:
I didn't realise you were also fixing this, so for completeness and future reference here's the commit message I wrote:

> Chef merges default attributes which are Arrays with their override
> counterparts, so specifying a default ring means that it will appear
> in the node attributes alongside whatever ring is configured in the
> proposal.  We don't ever want this default ring, especially since it
> only has one member, which causes an error about a mismatch between
> the number of members in each ring.  Even if that wasn't an issue, the
> IP refers to a whole network rather than a single node, so that would
> also cause an error.
> 
> In Chef 11 it seems that the Array merging behaviour may have changed:
> 
> https://coderanger.net/arrays-and-chef/
> 
> but in any case this merging has been proven to be a problem with the
> current codebase, so remove the default ring.

aspiers:
CI failed on aodh proposal, everything before that was fine so I think this change did not cause the failure.

rsalevsky:
Backport: https://github.com/crowbar/crowbar-ha/pull/273

PR #276, opened 2017-11-13
crowbar: Improve UI after addition of rings

vuntz:
Corosync rings can be slightly better integreated in the UI:

 - adding clear sections for corosync and pacemaker, to avoid confusion
   for the settings
 - not using sections for "existing rings" and "new rings"
 - moving the "Add" button for the new ring in the header of the list
   item
 - adding placeholder text for the ring entries

https://github.com/crowbar/crowbar-core/pull/1429 is required for correctly aligning the "add" button.

Old UI:
crowbar-ha old2

New UI:
crowbar-ha new

### DIFF
--- a/crowbar_framework/config/locales/pacemaker/en.yml
+++ b/crowbar_framework/config/locales/pacemaker/en.yml
@@ -46,6 +46,7 @@ en:
             registering public endpoints, for instance). Any name specified
             here should already exist in the upstream DNS zones.'
         gui_header: 'Pacemaker GUI'
+        corosync_rings_header: 'Corosync Rings'
         corosync:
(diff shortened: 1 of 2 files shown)


crowbar_framework/config/locales/pacemaker/en.yml
```diff
@@ -66,13 +67,16 @@ en:
             "true" for two-nodes cluster, and to "false" otherwise.'
           loading_text: 'Loading rings...'
           listheader: 'Corosync Rings'
```
scottwulf: Delete listheader as it is no longer in use

vuntz: Good catch, done

PR #279, opened 2017-11-21
haproxy: make maxconn parameter configurable (bsc#1068962)

dirkmueller:
It seems at least on the concurrent Galera DB connection side we
can easily hit the default limit of 2000 parallel connections. Raise
it to the global connection limit.

### DIFF
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -90,6 +90,10 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
 	option <%= option %>
     <% end -%>
 
+    <% unless content[:max_connections].nil? -%>
+	maxconn <%= content[:max_connections] %>
+    <% end -%>
+
     <% unless content[:default_server].nil? -%>
 	default-server <%= content[:default_server] %>
     <% end -%>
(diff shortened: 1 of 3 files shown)


vuntz:
I'm mixed on this; isn't galera the only use case where we really need to bump this? I'm also slightly concerned that one service could eat all connections, leaving all others with no resource.

dirkmueller:
@vuntz yeah, its a bit of a mixed back. the main issue I see with this approach is that we allow haproxy to exceed the underlying galera/mysql max connection limit, which causes it to stop accepting connections, and pacemaker then assuming the process being dead and doing stupid stuff..

I am going to rework this to set the haproxy limit to db connections - small number, so that some slots remain available for pacemaker and haproxy isn't limiting the db connections like it does now.

dirkmueller:
@rhafer I think it doesn't matter, the global limit will limit the max_connections.

chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -90,6 +90,10 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
 	option <%= option %>
     <% end -%>
 
+    <% unless content[:max_connections].nil? -%>
+	maxconn <%= content[:max_connections] %>
+    <% end -%>
+
```
vuntz: When this is not defined, you're adding an empty line to the config file, which adds to the previous empty line and looks ugly + causes a useless restart (or reload, I think for haproxy).

PR #280, opened 2017-11-23
pacemaker: deal with nil op attributes

abel-navarro:
When flatting out the list of 'op' parameters in pacemaker primitive
resource, that list can be nil instead of empty. This change addresses
that situation.


abel-navarro:

PR #282, opened 2017-11-24
pacemaker: Allow having multiple op settings for master/slave

vuntz:
It turns out that we can have multiple entries for the same data type.
For instance, when using ms, we need to define multiple "op monitor"
with different roles:
  https://clusterlabs.org/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/_monitoring_multi_state_resources.html

Also, add support for missing op operations (notify will be used).


Itxaka:
[RabbitMQ restarted by pacemaker from time to time](https://trello.com/c/unGNGzw0/209-rabbitmq-restarted-by-pacemaker-from-time-to-time)

PR #284, opened 2017-12-05
pacemaker: Alert resource

skazi0:
Pacemaker supports alert feature which can be used to attach scripts
to various cluster events. This commit exposes basic alerts to chef as
resources. Creating, updating and deleting is supported.

### DIFF
--- a/chef/cookbooks/pacemaker/spec/fixtures/alert.rb
+++ b/chef/cookbooks/pacemaker/spec/fixtures/alert.rb
@@ -1,0 +1,18 @@
+require_relative "../../libraries/pacemaker/alert"
+
+class Chef
+  module RSpec
+    module Pacemaker
+      module Config
+        ALERT = ::Pacemaker::Alert.new("alert1")
+        ALERT.handler = "handler.sh"
+        ALERT.receiver = "receiver-id"
+        ALERT.attrs_authoritative
+        ALERT_DEFINITION = <<'PCMK'.chomp
+alert alert1 handler.sh \
+         to receiver-id
+PCMK
+      end
+    end
+  end
+end
--- a/chef/cookbooks/pacemaker/libraries/pacemaker/alert.rb
+++ b/chef/cookbooks/pacemaker/libraries/pacemaker/alert.rb
@@ -1,0 +1,36 @@
+require "shellwords"
+
+require_relative "cib_object"
+
+module Pacemaker
+  class Alert < Pacemaker::CIBObject
+    register_type :alert
+
+    attr_accessor :handler, :receiver
+
+    def self.attrs_to_copy_from_chef
+      %w(handler receiver)
+    end
+
+    def parse_definition
+      unless @definition =~ /\A#{self.class.object_type} (\S+) (\S+)[\\\s]+to (\S+)/
+        raise Pacemaker::CIBObject::DefinitionParseError, \
+              "Couldn't parse definition '#{@definition}'"
+      end
+      self.name = Regexp.last_match(1)
+      self.handler = Regexp.last_match(2)
+      self.receiver = Regexp.last_match(3)
+
+      attrs_authoritative
+    end
+
+    def definition_from_attributes
+      str = "#{self.class.object_type} #{name} #{handler}"
+      str << continuation_line("to #{receiver}") unless receiver.empty?
+    end
+
+    def self.description
+      "alert"
+    end
+  end
+end
--- a/chef/cookbooks/pacemaker/providers/alert.rb
+++ b/chef/cookbooks/pacemaker/providers/alert.rb
@@ -1,0 +1,63 @@
+# Cookbook Name:: pacemaker
+# Provider:: alert
+#
+# Copyright:: 2017, SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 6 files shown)


skazi0:
Tests to be added.

skazi0:

skazi0:
@aspiers can you take a second look at this? the tests were added and some hound-polish was applied.

chef/cookbooks/pacemaker/libraries/pacemaker/resource/alert.rb
```diff
@@ -0,0 +1,31 @@
...
+require_relative "../resource"
+require_relative "../mixins/resource_meta"
+
+class Pacemaker::Resource::Alert < Pacemaker::Resource
+  register_type :alert
+
+  attr_accessor :handler, :receiver
+
+  def self.attrs_to_copy_from_chef
+    %w(handler receiver)
```
vuntz: Reading the doc, it seems we can have some `meta`, which means that this should be listed here and I guess we're missing a `include Pacemaker::Mixins::Resource::Meta`

It's useful to specify a timeout.

chef/cookbooks/pacemaker/libraries/pacemaker/resource/alert.rb
```diff
@@ -0,0 +1,31 @@
+require "shellwords"
+
+require_relative "../resource"
+require_relative "../mixins/resource_meta"
+
+class Pacemaker::Resource::Alert < Pacemaker::Resource
```
vuntz: Stupid question: is it really a resource? I'm thinking it's something else, a new type alongside contraints and resources.

aspiers: I don't think it is, because AFAIK you can't start or stop alerts.  `Pacemaker::Resource` provides instance methods `#running?`, `#crm_start_command`, and `#crm_stop_command` which IIUC don't apply here, therefore this class should instead inherit from `Pacemaker::CIBObject`.

chef/cookbooks/pacemaker/libraries/pacemaker/resource/alert.rb
aspiers: Right.  To save others having to google it, here are [the docs on alert meta-attributes](https://clusterlabs.org/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/_alert_meta_attributes.html).

aspiers:
@skazi0 Still -1 but only for the reasons @vuntz already mentioned :-)

chef/cookbooks/pacemaker/libraries/pacemaker/resource/alert.rb
skazi0: Looking at the TOC here: https://clusterlabs.org/doc/en-US/Pacemaker/1.1-pcs/html-single/Pacemaker_Explained/index.html , alerts are on the same level as resources and constraints. I got confused looking at the code. The specs define something called "non-runnable resource" which seemed to fit the alerts. I will reorganize things to make it clear that alerts are not resources.

chef/cookbooks/pacemaker/libraries/pacemaker/resource/alert.rb
skazi0: Let me first rework the non-resource part. IMO, meta can be added as a separate PR. This would speed things up in case we wanted to backport it to Cloud7 (i.e. meta is not really needed for our usecase).

chef/cookbooks/pacemaker/libraries/pacemaker/resource/alert.rb
skazi0: More confusion comes from "resource" having additional meaning in chef context.

skazi0: @aspiers @vuntz please take a second look at the current state of this PR. Alerts should now be independent of resources. Note that I have not tested the new version yet (only unit tests).

skazi0:
@vuntz @aspiers I have tested the modified version and it works as before. I will check the meta part tomorrow morning and if it's not too complicated I will add it to this PR. Otherwise I will create a trello card for that.

chef/cookbooks/pacemaker/spec/fixtures/alert.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class Chef
+  module RSpec
+    module Pacemaker
+      module Config
+        ALERT = ::Pacemaker::Alert.new("alert1")
+        ALERT.handler = "handler.sh"
+        ALERT.receiver = "receiver-id"
+        ALERT.attrs_authoritative
+        ALERT_DEFINITION = "alert alert1 handler.sh to receiver-id".freeze
```
aspiers: Presumably you meant to drop this duplicate definition?

It's causing a couple of new warnings in Travis:

```
/home/travis/build/crowbar/crowbar-ha/chef/cookbooks/pacemaker/spec/fixtures/alert.rb:12: warning: already initialized constant Chef::RSpec::Pacemaker::Config::ALERT_DEFINITION
/home/travis/build/crowbar/crowbar-ha/chef/cookbooks/pacemaker/spec/fixtures/alert.rb:11: warning: previous definition of ALERT_DEFINITION was here
```

aspiers:
@skazi0 Looks great, thanks!  Only need to drop that duplicate line and I think we're good to merge.

chef/cookbooks/pacemaker/spec/fixtures/alert.rb
skazi0: Uh, not sure how it got there. Must be some refactoring leftover. Removed.

PR #285, opened 2017-12-05
crowbar-pacemaker: Deploy ssh keys before joining

skazi0:
Authorized keys need to be deployed to nodes before joining cluster
so that the founder can ssh to the new node from alert handler.

Requires https://github.com/crowbar/crowbar-core/pull/1451


skazi0:

aspiers:
I do love it, because it provides more [evidence](https://github.com/search?q=toomanyreviewers&type=Issues) that [I'm right](http://mailman.suse.de/mlarch/SuSE/cloud-devel/2017/cloud-devel.2017.08/msg00207.html) :-)

PR #286, opened 2017-12-08
2-node clusters should have two_node: 1

matthewoliver:
This fixes https://bugzilla.suse.com/show_bug.cgi?id=981056

On pacemaker clusters of 2-nodes, we should set the
corosync quorum conf to use `two_node: 1` rather than
force the use of `no-quorum-policy=ignore` in pacemaker.

The corosync `two_nodes:1` tells corosync that it only
needs 1 vote for quorum instead of 2.
Before this patch we'd force these clusters to use the ignore
policy when the cluster was being defined via the
crowbar web ui (via javascript) which lead to
bugs like https://bugzilla.suse.com/show_bug.cgi?id=981056.

This patch is removing the javascript function that is
enforcing this behaviour completely as now that 2-node
clusters behave as expected, users should get a choice in
what no-quorum-policy they want to use.

### DIFF
--- a/chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
+++ b/chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
@@ -129,4 +129,10 @@ quorum {
 	# Enable and configure quorum subsystem (default: off)
 	# see also corosync.conf.5 and votequorum.5
 	provider: corosync_votequorum
+  <% if @members_v2.length <= 2 -%>
+	expected_votes: <%= @members_v2.length %>
+    <% if @members_v2.length == 2 -%>
+	two_node: 1
+    <% end -%>
+  <% end -%>
 }
--- a/chef/cookbooks/pacemaker/attributes/default.rb
+++ b/chef/cookbooks/pacemaker/attributes/default.rb
@@ -51,7 +51,7 @@ end
 default[:pacemaker][:founder] = nil
 default[:pacemaker][:is_remote] = false
 default[:pacemaker][:crm][:initial_config_file] = "/etc/corosync/crm-initial.conf"
-default[:pacemaker][:crm][:no_quorum_policy] = "ignore"
+default[:pacemaker][:crm][:no_quorum_policy] = "stop"
 # Should be longer than the systemd timeouts (defaults to 90s) so that
 # pacemaker only reacts when systemd is not helping anymore
 default[:pacemaker][:crm][:op_default_timeout] = 120
--- a/crowbar_framework/app/assets/javascripts/barclamps/pacemaker/application.js
+++ b/crowbar_framework/app/assets/javascripts/barclamps/pacemaker/application.js
@@ -225,7 +225,7 @@ function update_no_quorum_policy(evt, init) {
     if (evt.type == 'nodeListNodeUnallocated') { members -= 1; }
   }
 
-  if (members > 2) {
+  if (members > 1) {
     if (was_forced_policy) {
       no_quorum_policy_el.val(non_forced_policy);
       no_quorum_policy_el.removeData('non-forced');


matthewoliver:
OK this is suppose to WIP, but I don't seem to have the ability to add the label :(

chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
```diff
@@ -129,4 +129,9 @@ quorum {
 	# Enable and configure quorum subsystem (default: off)
 	# see also corosync.conf.5 and votequorum.5
 	provider: corosync_votequorum
+  <% if @members_v2.length <= 2 -%>
+	expected_votes: <%= @members_v2.length %>
+	two_node: 1
```
vuntz: Is `two_node: 1` really fine for a one-node cluster?

aspiers:
> OK this is suppose to WIP, but I don't seem to have the ability to add the label :(

Should be fixed now.

chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
aspiers: I'm guessing not ;-)

aspiers:
I'm not sure it's safe to always set `wait_for_all` to `0` - we should check with our local Pacemaker gurus. 
 @gao-yan or @krig please can you review this?

krig:
For a two-node cluster you definitely want to set `wait_for_all: 1`, by default setting `two_node: 1` also enables `wait_for_all: 1` unless `wait_for_all` is explicitly disabled.

With two nodes and `wait_for_all: 0`, there's almost certainly a STONITH race on cluster startup where each node will try to fence the other before they can see each other.

matthewoliver:
So does this mean we could just use `two_node: 1` and leave out the `wait_for_all` seeing as that will also enabled by default with `two_node`?

So from my reading, `wait_for_all` is only applied at initial start (all down to up). So the 2 node cluster would fail to start if a node was down. But that's ok. You'd hope when you initially start the cluster they'd both be there. It waits for `expected_votes` nodes to be available before asking for quorum as best as I can read.

Is this still ok on a 1 node cluster? I don't know why we'd support that. But anyway if it's a 1 node cluster, I assume we'd set  `expected_votes` to 1, having two_node: 1 (only needing 1 vote to make quorum) and `wait_for_all` being on means it too will only wait for 1 node... So I think this is ok, if it goes down, it'll be fenced. (which I guess is why one might have a 1 node cluster, to add fencing if it's something critical and you don't want it in an inconsistent state).

So something like this might be ok for both 1 and 2 node clusters:
  <% if @members_v2.length <= 2 -%>
  expected_votes: <%= @members_v2.length %>
  two_node: 1
  <% end -%>
Or is my thoughts on a 1 node cluster in error?

matthewoliver:
Hmm, when reading the votequorum (5) man page a bit closer, it seems `two_node` will automatically disable itself if there are more then 2 nodes:

> NOTES: enabling two_node: 1 automatically enables wait_for_all. It is still possible to override wait_for_all by explicitly setting it to 0. If more than 2 nodes join the cluster, the two_node option is automatically disabled.

So although more confusing when reading config, we could leave out the template if statement.. but having `two_node` always enabled could be very confusing to users reading the config.

matthewoliver:
Question. Is there an easy crowbar way to recreate the crowbar ui assets? I've been trying to test this patch and using reposync_crowbar which patches the javascript file, but the minified js assets isn't updated, what in crowbar or mkcloud even generates these.. or are they coming from a package somewhere?

I'm assuming I've just missed an obvious command. I just nuked them to make sure the deployment worked (no javascript no_quorum_policy validation). I'm not much of a npm expert, so thought I'd ask first before going down that rabbit hole :)

krig:
In the HA bootstrap configuration, we only set `two_node: 1` if the cluster really has exactly two nodes. So that's the configuration I'm most comfortable recommending. ;)

vuntz:
> Question. Is there an easy crowbar way to recreate the crowbar ui assets? I've been trying to test this patch and using reposync_crowbar which patches the javascript file, but the minified js assets isn't updated, what in crowbar or mkcloud even generates these.. or are they coming from a package somewhere?

`cd /opt/dell/crowbar_framework && RAILS_ENV=production rake assets:precompile` (iirc)

matthewoliver:
On 11 Dec. 2017 7:40 pm, "Kristoffer Grönlund" <notifications@github.com>
wrote:

In the HA bootstrap configuration, we only set two_node: 1 if the cluster
really has exactly two nodes. So that's the configuration I'm most
comfortable recommending. ;)

Great, I'll do that then :)

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/crowbar/crowbar-ha/pull/286#issuecomment-350656822>,
or mute
the thread

matthewoliver:
On 11 Dec. 2017 7:59 pm, "Vincent Untz" <notifications@github.com> wrote:

Question. Is there an easy crowbar way to recreate the crowbar ui assets?
I've been trying to test this patch and using reposync_crowbar which
patches the javascript file, but the minified js assets isn't updated, what
in crowbar or mkcloud even generates these.. or are they coming from a
package somewhere?

cd /opt/dell/crowbar_framework && RAILS_ENV=production rake
assets:precompile (iirc)

Thanks Vincent. Much appreciated!

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/crowbar/crowbar-ha/pull/286#issuecomment-350661350>,
or mute
the thread

matthewoliver:
@vuntz Yup that was exactly what I needed, thanks!

matthewoliver:
Based on what @krig recommended. I'll update the patch to:
 - Set `two_node: 1` and `expected_votes` to 2 only when there are 2 nodes in the cluster.
 - Instead of removing the JS code completely, I'll put it back, but only force `no-quorum-policy=ignore` on a single node cluster.

The 1 node cluster option above isn't what Kris recommended specifically, but is what we currently have, so at least that is expected behaviour from crowbar.

chef/cookbooks/corosync/templates/default/corosync.conf.v2.erb
```diff
@@ -129,4 +129,8 @@ quorum {
 	# Enable and configure quorum subsystem (default: off)
 	# see also corosync.conf.5 and votequorum.5
 	provider: corosync_votequorum
+  <% if @members_v2.length == 2 -%>
+	expected_votes: <%= @members_v2.length %>
```
vuntz: Shouldn't you set `expected_votes` for a one-node cluster too?

matthewoliver:
not if we have the ignore policy set.. or am I mistaken?

vuntz:
> not if we have the ignore policy set.. or am I mistaken?

Well, I don't know :-) Hence my question.

matthewoliver:
Just a rebase and an extra whitespace fix.

aspiers:
Unfortunately it looks like this referenced the wrong bug.  The issue is in the SOC product, not in the SLE HA product, and AFAICS the bug this PR was addressing is actually [bsc#982353 (no-quorum-policy=ignore is no longer recommended with corosync 2, and should never be set)](https://bugzilla.suse.com/show_bug.cgi?id=982353).

matthewoliver:
Yeah, Adam is right, the Bug Adam references mentions but 981056, which I followed to follow the conversation and when making not of the bug it fixes, must have selected the number from the wrong browser tab.
I'm sorry for the confusion.

matthewoliver:
Seems there are a few bugs that are similar. I went back to the SWAT trello board, and the actual bug this referenced was suppose to be: https://bugzilla.suse.com/show_bug.cgi?id=1023981 But it seems it might have also closed https://bugzilla.suse.com/show_bug.cgi?id=982353 (the one Adam mentioned above), as the PR also changes the default no_quorum_policy away from ignore.

aspiers:
@matthewoliver Thanks for the helpful archaeology - please can you also update bugzilla with the appropriate cross-links to eliminate any future confusion? Cheers!

PR #287, opened 2017-12-08
pacemaker: Made alert receiver optional

skazi0:
The receiver/recipient of alerts is not mandatory. Modified wrapper classes to support such cases.


skazi0:

PR #288, opened 2017-12-14
[4.0] pacemaker: Alert resource

skazi0:
backport of #284 and #287


skazi0:

PR #291, opened 2017-12-14
upgrade: fix pacemaker founder searchs

Itxaka:
When we introduced the new syncmarks for pacemaker
the founder was changed from a boolean indicating
that the node is a founder, to the fqdn of the founder

This results into #set_node_as_founder not working as expected
and unable to find the founder and change it to a new founder
properly

This patch refactors #set_node_as_founder to make sure we
find the old_founder and that we set the new_founder fqdn
into all the cluster nodes

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -79,37 +79,32 @@ module Api
           Rails.logger.error("Node #{name} does not have pacemaker setup")
           return false
         end
-        if new_founder[:pacemaker][:founder]
+        if new_founder[:pacemaker][:founder] == new_founder[:fqdn]
           Rails.logger.debug("Node #{name} is already the cluster founder.")
-          return true
         end
 
-        # 2. find the current cluster founder in the same cluster
+        # 2. find the role for this cluster and change the founder
         cluster_env = new_founder[:pacemaker][:config][:environment]
-        old_founder = NodeObject.find(
-          "pacemaker_founder:true AND pacemaker_config_environment:#{cluster_env}"
-        ).first
+        cluster_role = RoleObject.find_role_by_name(cluster_env)
 
-        if old_founder.nil?
-          Rails.logger.warning("No cluster founder found. Making #{name} the new founder anyway.")
-        else
-          old_founder[:pacemaker][:founder] = false
-          if old_founder[:drbd] && old_founder[:drbd][:rsc]
-            old_founder[:drbd][:rsc].each do |res, _|
-              old_founder[:drbd][:rsc][res][:master] = false
+        cluster_role.default_attributes["pacemaker"]["founder"] = new_founder[:fqdn]
+        cluster_role.save
+
+        # 3. change drdb master in all nodes from this cluster
+        ::Node.find("pacemaker_config_environment:#{cluster_env}").each do |node|
+          # we need to set the new founder on the cluster nodes anyway. This is because
+          # even if we have the new founder on the role, that will only be applied to the
+          # node during a chef run, but could lead to issues on node searches before that chef-run
+          # so for peace of mind, we update it on all nodes always.
+          node["pacemaker"]["founder"] = new_founder[:fqdn]
+          if node[:drbd] && node[:drbd][:rsc]
+            node[:drbd][:rsc].each_key do |res|
+              # if this is the new founder set master to true, false if its any other node
+              node[:drbd][:rsc][res][:master] = (node[:fqdn] == new_founder[:fqdn])
             end
           end
-          old_founder.save
+          node.save
         end
-
-        # 3. mark given node as founder
-        new_founder[:pacemaker][:founder] = true
-        if new_founder[:drbd] && new_founder[:drbd][:rsc]
-          new_founder[:drbd][:rsc].each do |res, _|
-            new_founder[:drbd][:rsc][res][:master] = true
-          end
-        end
-        new_founder.save
       end
 
       def repocheck


crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -79,37 +79,39 @@ def set_node_as_founder(name)
...
-          "pacemaker_founder:true AND pacemaker_config_environment:#{cluster_env}"
-        ).first
+        cluster_nodes = NodeObject.find(
+          "pacemaker_founder:* AND pacemaker_config_environment:#{cluster_env}"
+        )
+        old_founder = NodeObject.load(cluster_nodes.first[:pacemaker][:founder]) rescue nil
 
         if old_founder.nil?
           Rails.logger.warning("No cluster founder found. Making #{name} the new founder anyway.")
-        else
```
skazi0: what happened to this part?

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -79,37 +79,39 @@ def set_node_as_founder(name)
...
         end
 
-        # 2. find the current cluster founder in the same cluster
+        # 2. find all cluster nodes and the old_founder
         cluster_env = new_founder[:pacemaker][:config][:environment]
-        old_founder = NodeObject.find(
-          "pacemaker_founder:true AND pacemaker_config_environment:#{cluster_env}"
-        ).first
+        cluster_nodes = NodeObject.find(
+          "pacemaker_founder:* AND pacemaker_config_environment:#{cluster_env}"
```
jsuchome: `pacemaker_founder:* ` not needed

Itxaka: +1

crowbar_framework/app/models/api/pacemaker.rb
Itxaka: totally refactored the code as we cant longer just modify only the old_founder and new_founder but instead need to change the founder name in all cluster nodes, so I just redid it instead of trying to merge the old code with the new stuff.

looked easier to me :)

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -79,37 +79,41 @@ def set_node_as_founder(name)
...
+        # 3. set the new founder and drbd master in all the cluster nodes
+        cluster_nodes.each do |node|
+          node[:pacemaker][:founder] = new_founder[:fqdn]
+          if node[:fqdn] == new_founder[:fqdn] && node[:drbd] && node[:drbd][:rsc]
+            # set this node as drbd master
+            node[:drbd][:rsc].each_key do |res|
+              node[:drbd][:rsc][res][:master] = true
+            end
+          end
+          if !old_founder.nil? && node[:fqdn] == old_founder[:fqdn] && node[:drbd] && node[:drbd][:rsc]
```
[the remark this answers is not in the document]
Itxaka: You will have to live with that doggy, because  Im not changing that line for 3 chars to make it more difficult to read

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -79,37 +79,41 @@ def set_node_as_founder(name)
...
         end
 
-        # 3. mark given node as founder
-        new_founder[:pacemaker][:founder] = true
-        if new_founder[:drbd] && new_founder[:drbd][:rsc]
-          new_founder[:drbd][:rsc].each do |res, _|
-            new_founder[:drbd][:rsc][res][:master] = true
+        # 3. set the new founder and drbd master in all the cluster nodes
+        cluster_nodes.each do |node|
+          node[:pacemaker][:founder] = new_founder[:fqdn]
```
vuntz: So... that's slightly dangerous: in theory, you're not supposed to set the attribute on the node itself, but on the role for the proposal. In `pacemaker_service.rb`, we do:

```
role.default_attributes["pacemaker"]["founder"] = founder_name
```

Itxaka: ahhh thanks! You just saved me a trip to memory lane, because when I pushed I was wondering why this was done against the node as I remember the migration being done against the role. So now I dont have to search for it :D

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -79,37 +79,28 @@ def set_node_as_founder(name)
...
+        cluster_role.default_attributes["pacemaker"]["founder"] = new_founder[:fqdn]
+        cluster_role.save
 
-        # 3. mark given node as founder
-        new_founder[:pacemaker][:founder] = true
-        if new_founder[:drbd] && new_founder[:drbd][:rsc]
-          new_founder[:drbd][:rsc].each do |res, _|
-            new_founder[:drbd][:rsc][res][:master] = true
+        # 3. change drdb master in all nodes from this cluster
+        Node.find("role:#{cluster_env}").each do |node|
```
jsuchome: `::Node` (or `NodeObject`)

jsuchome:
There is still something wrong here. Both my cluster nodes are marked as founder (you can check at crowbar.vi7 - now without repeating any step)

Itxaka:
Well, drbd is still failing so Im not sure if the limbo status of both nodes being founders does not stem from there.

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -79,37 +79,27 @@ def set_node_as_founder(name)
...
+        cluster_role.default_attributes["pacemaker"]["founder"] = new_founder[:fqdn]
+        cluster_role.save
 
-        # 3. mark given node as founder
-        new_founder[:pacemaker][:founder] = true
-        if new_founder[:drbd] && new_founder[:drbd][:rsc]
-          new_founder[:drbd][:rsc].each do |res, _|
-            new_founder[:drbd][:rsc][res][:master] = true
+        # 3. change drdb master in all nodes from this cluster
+        ::Node.find("role:#{cluster_env}").each do |node|
```
jsuchome: I think adding

`node["pacemaker"]["founder"] = new_founder[:fqdn]`

at this place should help

Itxaka: indeed because the role attributes are not passed to the node unless there is a chef client, according to my tests. So we need to remove this or just add the new founder fqdn to all nodes.

jsuchome: Also, I think you need this search:
```
::Node.find("pacemaker_config_environment:#{cluster_env}").each do |node|
```

Itxaka:
aha! @jsuchome this totally worked! Founder was set up correctly at the same time across all nodes+role.

but now it fails to sync drbd :(

```
modinfo: ERROR: Module drbd not found.
modinfo: ERROR: Module drbd not found.
modprobe: FATAL: Module drbd not found.
Failed to modprobe drbd (No such file or directory)
Command 'drbdsetup status' terminated with exit code 20
command status exited with code 20
```

jsuchome:
@Itxaka this probably means that the HA repo is not present on the node, am I right? IMHO this is caused by applying your 2 patches too late (after nodes repocheck)

Itxaka:
> @Itxaka this probably means that the HA repo is not present on the node, am I right? IMHO this is caused by applying your 2 patches too late (after nodes repocheck)

Ahh yes, you mentioned that. Oh well, finally we can move this forward :)

Itxaka:
> 

>Let's move forward; but I think we will need some code, somewhere, to remove the founder attribute on each node.
>Reason for this is that if the founder changes later on for any reason, this would fail because of these attributes that will never get updated.

Sounds good, I will test that and have a follow up patch to try and fix that

PR #292, opened 2017-12-15
pacemaker: fix migration number

Itxaka:
There were 2 migrations with the same number


vuntz:
> There were 2 migrations with the same number

It's actually fine, they will both run. Or did you hit issues with that?

Itxaka:
| It's actually fine, they will both run. Or did you hit issues with that?

No issues that I could see, but I dont like having duplicates and the schema-migration being out of order!

Itxaka:
rebased!

PR #293, opened 2017-12-15
pacemaker: fix migrations

Itxaka:
There were 2 migrations with the same number


aspiers:

PR #296, opened 2018-01-04
pacemaker: Add meta support to alerts

skazi0:
Pacemaker alerts support some meta attributes (e.g. timeout). This
change adds support for alert-level meta. Pacemaker also supports
recipient meta attributes (e.g. timestamp-format) but these are not
supported in pacemaker cookbooks/libraries yet.

To avoid code duplication, common functions were moved from Resource
up to CIBObject. Alert definition parsing was extended and improved
using these.

Quotes were added around handler path to match crm behavior.

Small bug in error message in extract_hash_from_one was fixed.


skazi0:
[Add `meta` support to alert resource in pacemaker cookbook](https://trello.com/c/UkxeNKZs/221-add-meta-support-to-alert-resource-in-pacemaker-cookbook)

skazi0:
@aspiers please re-review

PR #297, opened 2018-01-04
pacemaker: added support for op defaults

AbelNavarro:
**Note:** this is a rework from https://github.com/crowbar/crowbar-ha/pull/262

on-fail" action.

other resourced.

be extended later on.

### DIFF
--- a/chef/cookbooks/pacemaker/providers/primitive.rb
+++ b/chef/cookbooks/pacemaker/providers/primitive.rb
@@ -69,6 +69,12 @@ def create_resource(name)
   standard_create_resource
 end
 
+def get_cluster_op_defaults(node)
+  return nil unless !node[:pacemaker].nil? && !node[:pacemaker][:op_defaults].nil?
+
+  node[:pacemaker][:op_defaults].to_hash
+end
+
 def update_resource(name)
   current_agent = @current_resource.agent
   unless current_agent.include? ":"
@@ -90,12 +96,56 @@ def update_resource(name)
   maybe_modify_resource(name)
 end
 
+# Merge in op defaults from barclamp to those from pacemaker
+# Returns a *single* Hash with merged ops for 'monitor'. Should 'monitor' be
+# an array of hashes, it has to be dealt externally to this function.
+def merge_cluster_op_monitor_defaults(op_monitor, op_defaults)
+  op_defaults ||= {}
+  op_defaults["monitor"] ||= {}
+  op_defaults["monitor"]["on-fail"] ||= ""
+
+  if !op_defaults["monitor"]["on-fail"].empty?
+    # Merge defined 'on-fail' default with already existing ops
+    default_op_monitor = { "on-fail" => op_defaults["monitor"]["on-fail"] }
+    default_op_monitor.merge(op_monitor)
+  else
+    # 'on-fail' default not defined, return original ops
+    op_monitor
+  end
+end
+
 def maybe_modify_resource(name)
   deprecate_target_role
 
   Chef::Log.info "Checking existing #{@current_cib_object} for modifications"
 
   cmds = []
+
+  # We don't want to modify data from the recipe, so ensure we have a copy as a hash
+  if new_resource.op.nil?
+    ops = {}
+  elsif new_resource.op.is_a? Hash
+    ops = new_resource.op.clone
+  else
+    ops = new_resource.op.to_hash
+  end
+
+  cluster_op_defaults = get_cluster_op_defaults(node)
+
+  ops["monitor"] ||= {}
+
+  # The ops["monitor"] section is presented in two flavors:
+  # - Single op value: Hash
+  # - Multiple op values (i.e. for master vs slave): Array of Hashes
+  if ops["monitor"].is_a?(Array)
+    ops["monitor"].map! do |op_monitor|
+      merge_cluster_op_monitor_defaults(op_monitor, cluster_op_defaults)
+    end
+  elsif ops["monitor"].is_a?(Hash)
+    ops["monitor"] = merge_cluster_op_monitor_defaults(ops["monitor"], cluster_op_defaults)
+  end
+
+  new_resource.op(ops)
 
   desired_primitive = cib_object_class.from_chef_resource(new_resource)
 
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -54,14 +54,19 @@
         "admin_name": "",
         "public_name": ""
       },
-      "clone_stateless_services": false
+      "clone_stateless_services": false,
+      "op_defaults": {
+        "monitor": {
+          "on-fail": ""
+        }
+      }
     }
   },
   "deployment": {
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 202,
+      "schema-revision": 203,
       "element_states": {
         "pacemaker-cluster-member"    : [ "readying", "ready", "applying" ],
         "hawk-server"                 : [ "readying", "ready", "applying" ],
--- a/chef/data_bags/crowbar/migrate/pacemaker/203_op_defaults.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/203_op_defaults.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["op_defaults"] = ta["op_defaults"] unless a.key? "op_defaults"
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("op_defaults") unless ta.key? "op_defaults"
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-pacemaker.schema
+++ b/chef/data_bags/crowbar/template-pacemaker.schema
@@ -141,7 +141,20 @@
                 "public_name": { "type": "str", "required": true }
               }
             },
-            "clone_stateless_services": { "type": "bool", "required": true }
+            "clone_stateless_services": { "type": "bool", "required": true },
+            "op_defaults": {
+              "type": "map",
+              "required": true,
+              "mapping": {
+                "monitor": {
+                  "type": "map",
+                  "required": true,
+                  "mapping": {
+                    "on-fail": { "type": "str", "required": true, "pattern": "/^ignore$|^block$|^stop$|^restart$|^fence$|^standby$|^$/"}
+                  }
+                }
+              }
+            }
           }
         }
       }


chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -97,6 +104,25 @@ def maybe_modify_resource(name)
 
   cmds = []
 
+  ops = new_resource.op
+  ops = ops.to_hash unless ops.is_a? Hash
+  op_defaults = op_defaults(node)
+
+  # If op_defaults is defined and not nil, set the value
+  # otherwise remove it
+  if !op_defaults.nil? &&
```
aspiers: The comment here is a code smell.  You can get rid of it by simplifying the huge conditional expression via "extract temporary variable" refactorings, and using clearly named temporary variables, e.g. (untested)

    on_fail_default_specified =
        !op_defaults.nil? &&
        op_defaults.key?("monitor") &&
        op_defaults["monitor"].key?("on-fail") &&
        !op_defaults["monitor"]["on-fail"].nil? &&
        !op_defaults["monitor"]["on-fail"].empty?

    monitor_op_needs_default =
        ops.key?("monitor") &&
        ops["monitor"].is_a?(Hash) &&
        !ops["monitor"].key?("on-fail")

    if on_fail_default_specified && monitor_op_needs_default
      ...
    end

Notice also how I avoided the need for parentheses like you had around this:

    (ops["monitor"].is_a? Hash)

Small cosmetic style tweak but I think my way is a bit more consistent with the other parts of the conditional, and hence very slightly easier on the eyes.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -97,6 +104,25 @@ def maybe_modify_resource(name)
 
   cmds = []
 
+  ops = new_resource.op
+  ops = ops.to_hash unless ops.is_a? Hash
```
aspiers: I think the motivation for ensuring it's a `Hash` (and thereby avoiding accidentally changing any C::N::A as a side-effect) is obscure enough that it warrants a comment explaining it.

AbelNavarro: agreed. Comment will be included in the next commit update.

chef/cookbooks/pacemaker/providers/primitive.rb
AbelNavarro: Agree with the cosmetic changes. I have implemented those partially, mainly because the original motivation behind `(ops["monitor"].is_a? Hash)`. That exception was in place for the recent rabbitmq changes that allow two different monitor configurations per service, mainly to differentiate between master and slave roles. In that case, `monitor` is an array with the two configurations (Hash) instead of a plain Hash. The new code should deal with this situation.

AbelNavarro:

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +97,48 @@ def update_resource(name)
...
+    !op_defaults.nil? &&
+    op_defaults.key?("monitor") &&
+    op_defaults["monitor"].key?("on-fail") &&
+    !op_defaults["monitor"]["on-fail"].nil? &&
+    !op_defaults["monitor"]["on-fail"].empty?
+
+  if on_fail_default_specified && ops.key?("monitor")
+    if ops["monitor"].is_a?(Array)
+      ops["monitor"].map! do |op_monitor|
+        op_monitor = set_op_default(op_monitor, op_defaults)
```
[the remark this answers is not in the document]
aspiers: Please fix this warning, which is correct.

AbelNavarro:
@aspiers fixed that map assignment.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -69,6 +69,13 @@ def create_resource(name)
   standard_create_resource
 end
 
+def op_defaults(node)
+  return nil unless node[:pacemaker][:founder]
```
vuntz: Why check for founder? I think it's actually wrong since it could lead to inconsistent results if something is run on founder and also on a non-founder node.

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -54,7 +54,12 @@
...
-      "clone_stateless_services": true
+      "clone_stateless_services": true,
+      "op_defaults": {
+        "monitor": {
+          "on-fail": ""
+        }
+      }
     }
   },
   "deployment": {
```
vuntz: You need to bump the schema revision in this file (and adjust the migration filename). I guess there was a migration added since you initially started working on this.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +97,48 @@ def update_resource(name)
...
+
+  on_fail_default_specified =
+    !op_defaults.nil? &&
+    op_defaults.key?("monitor") &&
+    op_defaults["monitor"].key?("on-fail") &&
+    !op_defaults["monitor"]["on-fail"].nil? &&
+    !op_defaults["monitor"]["on-fail"].empty?
+
+  if on_fail_default_specified && ops.key?("monitor")
+    if ops["monitor"].is_a?(Array)
```
vuntz: Wondering out loud: can it really be an `Array`? Do you have an example?

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +97,48 @@ def update_resource(name)
...
+  op_defaults = op_defaults(node)
+
+  on_fail_default_specified =
+    !op_defaults.nil? &&
+    op_defaults.key?("monitor") &&
+    op_defaults["monitor"].key?("on-fail") &&
+    !op_defaults["monitor"]["on-fail"].nil? &&
+    !op_defaults["monitor"]["on-fail"].empty?
+
+  if on_fail_default_specified && ops.key?("monitor")
```
vuntz: Don't we want to still set the default even if the primitive definition in the recipe doesn't contain anything about `"monitor"`?

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +97,48 @@ def update_resource(name)
...
+  ops = ops.to_hash unless ops.is_a? Hash
+
+  op_defaults = op_defaults(node)
+
+  on_fail_default_specified =
+    !op_defaults.nil? &&
+    op_defaults.key?("monitor") &&
+    op_defaults["monitor"].key?("on-fail") &&
+    !op_defaults["monitor"]["on-fail"].nil? &&
+    !op_defaults["monitor"]["on-fail"].empty?
```
vuntz: I somehow would move all this to `set_op_default`, see other comment.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +97,48 @@ def update_resource(name)
   maybe_modify_resource(name)
 end
 
+def set_op_default(op_monitor, op_defaults)
+  if !op_monitor.key?("on-fail")
+    op_monitor.merge("on-fail" => op_defaults["monitor"]["on-fail"])
+  else
+    op_monitor
+  end
```
vuntz: First thing I wonder: wouldn't this work in all cases (so no need for if/else):

```ruby
default_op_monitor = { "on-fail" => op_defaults["monitor"]["on-fail"] }
default_op_monitor.merge(op_monitor)
```

vuntz: Also, as said in a comment below, I'd move all the test on whether op_defaults contain something useful to this method.

So it'd be more like:

```ruby
# these three lines could be moved to the `op_defaults` method
op_defaults ||= {}
op_defaults["monitor"] ||= {}
op_defaults["monitor"]["on-fail"] ||= ""
if op_defaults.nil? || op_defaults["monitor"]["on-fail"].empty?
  op_monitor
else
  <current logic>
end
```

chef/cookbooks/pacemaker/providers/primitive.rb
AbelNavarro: Right, removing that. The reasoning for that is that we don't want the default to be set in different places. But this is already taken care of on the pacemaker_primitive invocation if needed: i.e.:

```
pacemaker_primitive smtp_resource do
  agent node[:pacemaker][:notifications][:agent]
  action [:stop, :delete]
  only_if "crm configure show #{smtp_resource}"
  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
end
```

chef/cookbooks/pacemaker/providers/primitive.rb
AbelNavarro: Tested that and works like a charm. Nice Ruby writing.

AbelNavarro: Agreed, will move it to `set_op_default`.

chef/cookbooks/pacemaker/providers/primitive.rb
AbelNavarro: Agreed, will move it to `set_op_default` as well.

chef/cookbooks/pacemaker/providers/primitive.rb
AbelNavarro: Yes, it can be an array. This was introduced at https://github.com/crowbar/crowbar-ha/pull/282

chef/data_bags/crowbar/template-pacemaker.json
AbelNavarro: Right. I had that on the other PR but didn't port it here.

chef/cookbooks/pacemaker/providers/primitive.rb
AbelNavarro: Right, will try to find some nice Ruby syntax to make it clean.

AbelNavarro:
CI keeps failing with `crowbarctl proposal commit database default` which is totally unrelated to this change. Rebuilding (again).

AbelNavarro:
@vuntz green light. Time for another review please.

AbelNavarro:
@aspiers mind having another look at this PR? danke!

AbelNavarro:
@vuntz right, fixing that.

AbelNavarro:
@vuntz @aspiers green light again. Mind having another look?

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +96,49 @@ def update_resource(name)
   maybe_modify_resource(name)
 end
 
+def set_op_default(op_monitor, op_defaults)
```
aspiers: I find this quite hard to understand, despite having been involved in its design.  I think a big part of the problem is in the variable naming:

- You have a method and a local variable both called `op_defaults`, which is really confusing.
- The variable names don't clearly distinguish between global `op` settings which comes from the Pacemaker proposal (via the node) and local `op` settings which come from the cookbook invoking `pacemaker_primitive`.  The whole point of this code is to take a copy of the global defaults and then merge in the per-cookbook settings without accidentally mutating any attributes along the way, but that's not obvious here.

In other words a lot of local knowledge is assumed, without any explanation.  Adding comments to explain is one possible solution, but that's often a bad smell.  If you can make the code clearer without adding (a lot of) comments, that would be much better.  So I suggest trying to rename the variables / methods to things like `global_op_defaults` and `local_op_defaults`, and adding a comment explaining why we are carefully avoiding mutation of data.  Then maybe it will be easier for others to follow.

AbelNavarro: @aspiers thanks for the feedback. I understand it can be a problem to get the full context just reading the code. I made slight changes renaming vars in this commit: https://github.com/crowbar/crowbar-ha/pull/297/commits/4687e59ef3d9aa35ad0c6605060bfda15ff8ea59

I'm hesitant to do more changes because the code took long time to get it right. I'm open to more comments or even more vars/functions renaming though. How does it look like in this new approach?

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -69,6 +69,12 @@ def create_resource(name)
   standard_create_resource
 end
 
+def get_barclamp_op_defaults(node)
```
aspiers: Thanks, this is definitely a step in the right direction, but the name is still vague because it looks like it could be referring to any barclamp (e.g. the one containing the currently running recipe), but in fact it refers specifically to the `pacemaker` barclamp.  That's why I previously suggested `global_op_defaults` to distinguish cluster-wide defaults from op values specific to a particular recipe/resource.  Having said that, it's per `pacemaker` proposal, which is actually per cluster rather than global like I originally thought.  So I think `get_cluster_op_defaults` would be better here.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +96,50 @@ def update_resource(name)
   maybe_modify_resource(name)
 end
 
+# Merge in op defaults from barclamp to those from pacemaker
+def set_op_default(op_monitor, op_defaults)
```
aspiers: This method is actually not setting defaults, nor is it setting for all ops, but only for monitor.  So I suggest `merge_cluster_op_monitor_defaults`.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +96,50 @@ def update_resource(name)
...
+  ops = new_resource.op
+
+  # Convert to Hash the ops structure in case is in Chef::Node::Attribute form
+  # This is to avoid accidentally changing the original value
+  ops = ops.to_hash unless ops.is_a? Hash
+
+  ops ||= {}
+  ops["monitor"] ||= {}
+
+  barclamp_op_defaults = get_barclamp_op_defaults(node)
```
aspiers: cluster_op_defaults = get_cluster_op_defaults(node)

AbelNavarro:
Thanks @aspiers for the naming proposals. Those actually look much better than the ones I suggested. I think I've been too involved with this code to come up with with naming. I've added your suggestions in the last amended commit.

Btw, this extra commit to address the new naming with be squashed. For now it's much easier for me to have it separate.

aspiers:
> I've added your suggestions in the last amended commit.

I don't see that :-/

> Btw, this extra commit to address the new naming with be squashed. For now it's much easier for me to have it separate.

Why is it easier?  It actually seems a lot harder to me, because it subverts GitHub's mechanism for marking comments as outdated.  Please just [squash and force-push every time](https://blog.adamspiers.org/2017/08/16/squash-merging-and-other-problems-with-github/).  If that is difficult for you then I suspect you are missing some simple git trick and I'd be happy to try to help make it easier for you.

AbelNavarro:
@aspiers sorry, I forgot to push the changes yesterday, so should be no uploaded and squashed.

AbelNavarro:
testbuild fails because the package from this PR (https://github.com/crowbar/crowbar-ha/pull/300) still isn't in the ISO. Will trigger a rebuild tomorrow.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +96,50 @@ def update_resource(name)
...
 
   Chef::Log.info "Checking existing #{@current_cib_object} for modifications"
 
   cmds = []
 
+  ops = new_resource.op
+
+  # Convert to Hash the ops structure in case is in Chef::Node::Attribute form
+  # This is to avoid accidentally changing the original value
+  ops = ops.to_hash unless ops.is_a? Hash
```
vuntz: I think this is slightly wrong: if we were given a hash, we would modify it here, and the recipe would not expect that.

So I'd do something like:

```ruby
# We don't want to modify data from the recipe, so ensure we have a copy as a hash
if new_resource.op.nil?
  ops = {}
if new_resource.op.is_a? Hash
  ops = new_resource.op.clone
else
  ops = new_resource.op.to_hash
end
```

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +96,50 @@ def update_resource(name)
...
+
+  ops ||= {}
```
vuntz: That line could be dropped with my suggested code above.

chef/cookbooks/pacemaker/providers/primitive.rb
aspiers: Yep, that looks good to me.

aspiers: > when you get it with new_resource.op you are getting a copy of it

Hrm, how do you know that?  And are you 100% sure it's a deep copy, rather than the values pointing to mutable data structures which are actually owned elsewhere?  If a caller passes in values from a cookbook, e.g.

```
  pacemaker_primitive ... do
    agent "..."
    op({
        start: { timeout: 120 },
        monitor: nova[:ha][:op][:monitor]
    })
    action :update
    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
  end
```

then I would expect `new_resource.op` to contain a `Hash` which has a `Chef::Node::Attribute` nested inside it, and that is not necessarily safe.

Admittedly I can't find any code where we actually do it like that now, but it's not safe to assume that it will never be written like that in the future.

AbelNavarro: I can only talk from empiric experience: these hash values could not be modified. But let's agree to do that copy so we can be sure in other situations.

Notice though that the case you propose (hash containing Chef::Node::Attribute) is also error prone, because the hash.clone will just do a shallow copy on the Chef::Node::Attribute. This is the problem we originally had (I was doing Chef::Node::Atribute.clone and .dup) and the object was duplicated, but not the references. I hope we don't step into that case.

aspiers: > I can only talk from empiric experience

Relying only on empirical evidence won't take care of all the cases :-)  You wouldn't have seen modifiable hash values because currently we don't nest `C::N::A` inside `Hash` anywhere.  We only construct hashes in places like https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/postgresql/recipes/ha_storage.rb#L78 and in those cases they don't nest mutable objects.

> because the hash.clone will just do a shallow copy on the Chef::Node::Attribute

Sure, that's why I recommended a deep clone a few weeks / months ago when we were originally discussing how to handle this correctly.  [This StackOverflow article](https://stackoverflow.com/questions/8206523/how-to-create-a-deep-copy-of-an-object-in-ruby) might come in useful here.

AbelNavarro:
@vuntz @aspiers your comments have been addressed. Build is green. Mind another review?

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +96,51 @@ def update_resource(name)
...
 
+# Merge in op defaults from barclamp to those from pacemaker
+def merge_cluster_op_monitor_defaults(op_monitor, op_defaults)
+  return op_monitor if op_defaults.nil?
+
+  op_defaults ||= {}
+  op_defaults["monitor"] ||= {}
+  op_defaults["monitor"]["on-fail"] ||= ""
+
+  if !op_defaults["monitor"]["on-fail"].empty?
```
aspiers: This seems needlessly complex.  Can't you just replace the whole of this method with these two lines?

    op_defaults ||= {}
    (op_defaults["monitor"] || {}).merge(op_monitor)

AbelNavarro: The problem I see with this approach is that the merge is done at `op_defaults["monitor"]` level, while the original function did the merge at `op_defaults["monitor"]["on-fail"]` level. The intention for that was that the merge was doable for Hash and Array of Hashes. Although I agree the function is verbose I don't see any way to clean it up.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +96,51 @@ def update_resource(name)
...
+  elsif new_resource.op.is_a? Hash
+    ops = new_resource.op.clone
+  else
+    ops = new_resource.op.to_hash
+  end
+
+  cluster_op_defaults = get_cluster_op_defaults(node)
+
+  ops["monitor"] ||= {}
+  if ops["monitor"].is_a?(Array)
```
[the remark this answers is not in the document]
AbelNavarro: Agree, adding a comment here.

chef/cookbooks/pacemaker/providers/primitive.rb
```diff
@@ -90,13 +96,51 @@ def update_resource(name)
   maybe_modify_resource(name)
 end
 
+# Merge in op defaults from barclamp to those from pacemaker
+def merge_cluster_op_monitor_defaults(op_monitor, op_defaults)
+  return op_monitor if op_defaults.nil?
+
+  op_defaults ||= {}
```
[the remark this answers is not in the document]
AbelNavarro: Agree, fixing that.

chef/cookbooks/pacemaker/providers/primitive.rb
aspiers: OK I think I see what you mean - here `op_defaults["monitor"]` will always be a `Hash` or `nil`, whereas `op_monitor` could be a `Hash` or an `Array`?  If so, again this is non-obvious and deserves a comment.

AbelNavarro: Agree with that. I hope the added message will clarify things up.

PR #299, opened 2018-01-16
upgrade: fix pacemaker founder searchs

Itxaka:
When we introduced the new syncmarks for pacemaker
the founder was changed from a boolean indicating
that the node is a founder, to the fqdn of the founder

This results into #set_node_as_founder not working as expected
and unable to find the founder and change it to a new founder
properly

This patch refactors #set_node_as_founder to make sure we
set the new founder in the proper role for this cluster and
we set the drbd master (if exists) to the new_founder

(cherry picked from commit 7be1e3fc8ebaeca5c6ad1a5df4c1765801904c19)

Forward-port-of: https://github.com/crowbar/crowbar-ha/pull/291

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -22,7 +22,7 @@ module Api
         unless repocheck["ha"]["available"]
           return { errors: [I18n.t("api.pacemaker.ha_not_installed")] }
         end
-        founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+        founders = NodeObject.find("pacemaker_config_environment:*")
         founders.empty? ? { errors: [I18n.t("api.pacemaker.ha_not_configured")] } : {}
       end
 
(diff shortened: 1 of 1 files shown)


Itxaka:
[Port Itxaka's fixes for cluster founder to master](https://trello.com/c/lbXPqsNz/236-port-itxakas-fixes-for-cluster-founder-to-master)

Itxaka:
> It might be better to do the additional changes (that were not needed in the original PR) as a separate commit.... but it's probably fine, if you don't wanna change it now

I thought that when doing a forward/back-port the original PR was adapted to the new branch?

Maybe this is more of an adaptation and its extending it, but I kind of thing that it should be together in case of a revert or looking for where an issue came from it makes more sense to have it together...I think.

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -44,7 +44,13 @@ def health_report
...
         failed_actions = {}
 
-        founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+        # get unique list of founder names across all clusters
+        cluster_founders_names = NodeObject.find(
+          "run_list_map:pacemaker-cluster-member"
+        ).map! do |node|
+          node[:pacemaker][:founder]
+        end.uniq
+        founders = cluster_founders_names.map { |name| NodeObject.find_by_name(name) }
```
jsuchome: That's ineffective to do the search for all nodes first, filter out the names and than do the new (limited) search by names...
I would let this be, but just in case you have better solution quickly enough, we could use it instead

Itxaka: not really, at least not a performant one :(

jsuchome: OK. Truth is, this method is normally called only once (or few times) for an upgrade...

Itxaka: we may be able to cache the founder list somewhere maybe? Then invalidate it when changing the founder, but for the rest of the upgrade it should be the same list of founder names I guess?

PR #300, opened 2018-02-01
pacemaker: Change default for clone_stateless_services to false

vuntz:
As a reminder, with this option set to false, pacemaker will not manage
stateless services such as OpenStack API services -- instead, systemd
will restart them on crash.

This works rather well, and since we never went to the point where we
had some OCF resource agent for such services, and only use a systemd
resource agent, the benefit of pacemaker is too limited. Reducing
complexity, and converging with the Ardana deployment is therefore more
interesting here.

### DIFF
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -54,7 +54,7 @@
         "admin_name": "",
         "public_name": ""
       },
-      "clone_stateless_services": true
+      "clone_stateless_services": false
     }
   },
   "deployment": {


vuntz:
Actually, one question is whether we still want to have the setting in the UI, and whether we even want to allow the old behavior for new deployments... I'd be in favor of disallowing the old behavior for new deployments.

Thoughts?

nicolasbock:
Following along your lines of argument to reduce complexity, I suggest to remove this setting from the UI. The less the user can fiddle with, the less complex the product is :wink:

vuntz:
I'll do a followup for the additional change I suggested.

PR #301, opened 2018-02-01
Disallow DRBD for new deployments

vuntz:
DRBD is only useful for PostgreSQL HA and unclustered RabbitMQ HA, which
are both features we want to disallow for new deployments. So DRBD has
no real use in new deployments.

Depends on https://github.com/crowbar/crowbar-openstack/pull/1545

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -718,9 +718,17 @@ class PacemakerService < ServiceObject
     end
 
     if proposal["attributes"][@bc_name]["drbd"]["enabled"]
-      validation_error I18n.t(
-        "barclamp.#{bc_name}.validation.drbd"
-      ) if members.length != 2
+      proposal_id = proposal["id"].gsub("#{@bc_name}-", "")
+      proposal_object = Proposal.where(barclamp: @bc_name, name: proposal_id).first
+      if proposal_object.nil? || !proposal_object.active_status?
+        validation_error I18n.t(
+          "barclamp.#{@bc_name}.validation.no_new_drbd"
+        )
+      else
+        validation_error I18n.t(
+          "barclamp.#{bc_name}.validation.drbd"
+        ) if members.length != 2
+      end
     end
 
     nodes = NodeObject.find("roles:provisioner-server")
--- a/crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
@@ -119,15 +119,6 @@
 
     %fieldset
       %legend
-        = t(".drbd_header")
-
-      .alert.alert-info
-        = t(".drbd.info")
-
-      = boolean_field %w(drbd enabled)
-
-    %fieldset
-      %legend
         = t(".haproxy_header")
 
       .alert.alert-info
--- a/crowbar_framework/config/locales/pacemaker/en.yml
+++ b/crowbar_framework/config/locales/pacemaker/en.yml
@@ -25,10 +25,6 @@ en:
   barclamp:
     pacemaker:
       edit_attributes:
-        drbd_header: 'DRBD'
-        drbd:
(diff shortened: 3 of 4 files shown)


crowbar_framework/config/locales/pacemaker/en.yml
```diff
@@ -146,6 +142,7 @@ en:
         smtp_server: 'Invalid SMTP server for mail notifications.'
         sender_address: 'Invalid sender address for mail notifications.'
         recipient_address: 'Invalid recipient address for mail notifications.'
+        no_new_drbd: 'DRBD is not allowed for new deployments since nothing will use it.'
```
jsuchome: What about something like _no service can be configured to use it_ ?

s-t-e-v-e-n-k: I'd suggest "DRBD is disallowed for fresh deployments since services can not be configured to use it."

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -718,9 +718,16 @@ def validate_proposal_after_save proposal
     end
 
     if proposal["attributes"][@bc_name]["drbd"]["enabled"]
-      validation_error I18n.t(
-        "barclamp.#{bc_name}.validation.drbd"
-      ) if members.length != 2
+      proposal_id = proposal["id"].gsub("#{@bc_name}-", "")
+      if RoleObject.find_role_by_name("#{@bc_name}-config-#{proposal_id}").nil?
```
aspiers: Seeing this repeated several times between this and the other related PR makes me think that there should be a helper for checking proposal existence to facilitate DRY.  And I wonder if that helper would belong more to `Proposal` than `RoleObject` ...

crowbar_framework/config/locales/pacemaker/en.yml
aspiers: As before this should encourage operators to read the release notes (well, or documentation, whichever is more appropriate).

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -718,9 +718,16 @@ def validate_proposal_after_save proposal
...
+        validation_error I18n.t(
+          "barclamp.#{@bc_name}.validation.no_new_drbd"
+        )
+      else
+        validation_error I18n.t(
+          "barclamp.#{bc_name}.validation.drbd"
+        ) if members.length != 2
```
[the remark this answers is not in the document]
aspiers: Nope, bad idea Hound.

crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
```diff
@@ -117,15 +117,6 @@
...
         = string_field %w(notifications smtp to)
 
-    %fieldset
-      %legend
-        = t(".drbd_header")
-
-      .alert.alert-info
-        = t(".drbd.info")
-
-      = boolean_field %w(drbd enabled)
```
aspiers: There's a Javascript function `update_drbd_enabled` corresponding to this which should also presumably be deleted.

crowbar_framework/app/models/pacemaker_service.rb
vuntz: I checked back then, and didn't see anything. And I checked again today, and there's a method for this. So I was being stupid :-)

crowbar_framework/app/views/barclamp/pacemaker/_edit_attributes.html.haml
vuntz: Good catch, thanks!

crowbar_framework/config/locales/pacemaker/en.yml
vuntz: @aspiers not disagreeing, but that would be a downstream patch (in our package).

aspiers: OK, approving this on the assumption that there will be a downstream patch in our package for this.

aspiers:

PR #302, opened 2018-03-09
pacemaker: add failure nodes to sync fail message (bsc#1083721)

matthewoliver:

issue.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
@@ -64,28 +64,36 @@ module CrowbarPacemakerSynchronization
       return
     end
 
-    Chef::Log.info("Checking if cluster founder has set #{mark}...")
-
     founder_name = CrowbarPacemakerHelper.cluster_founder_name(node)
+    cluster_name = CrowbarPacemakerHelper.cluster_name(node)
 
+    Chef::Log.info("Checking if #{cluster_name} cluster founder #{founder_name} has set #{mark}...")
     begin
       Timeout.timeout(timeout) do
         loop do
           if CrowbarPacemakerCIBAttribute.get(founder_name, "#{prefix}#{mark}", "0") != "0"
-            Chef::Log.info("Cluster founder has set #{mark}.")
+            Chef::Log.info("Cluster founder #{founder_name} has set #{mark} on cluster " \
+              "#{cluster_name}.")
             break
           end
-          Chef::Log.debug("Waiting for cluster founder to set #{mark}...")
+          Chef::Log.debug("Waiting for cluster founder #{founder_name} to set #{mark} " \
+            "on cluster #{cluster_name}...")
           sleep(5)
         end # loop
       end # Timeout
     rescue Timeout::Error
       if fatal
-        message = "Cluster founder didn't set #{mark}!"
+        message = \
+          "Cluster founder #{founder_name} didn't set #{mark} on cluster #{cluster_name}! " \
+          "Timed out while waiting for the founder; please check either " \
+          "'/var/log/crowbar/chef-client/#{founder_name}.#{node[:domain]}.log' " \
+          "on the admin node or '/var/log/chef/client.log' on #{founder_name} to see " \
+          "what went wrong."
         Chef::Log.fatal(message)
         raise message
       else
-        message = "Cluster founder didn't set #{mark}! Going on..."
+        message = "Cluster founder #{founder_name} didn't set #{mark} on cluster " \
+          "#{cluster_name}! Going on..."
         Chef::Log.warn(message)
       end
     end
@@ -97,11 +105,12 @@ module CrowbarPacemakerSynchronization
     return unless CrowbarPacemakerHelper.is_cluster_founder?(node)
 
     attribute = "#{prefix}#{mark}"
+    founder_name = CrowbarPacemakerHelper.cluster_founder_name(node)
 
     if CrowbarPacemakerCIBAttribute.get(node[:hostname], attribute, "0") != "0"
-      Chef::Log.info("Synchronization cluster mark #{mark} already set.")
+      Chef::Log.info("Synchronization cluster mark #{mark} already set on #{founder_name}.")
     else
-      Chef::Log.info("Setting synchronization cluster mark #{mark}.")
+      Chef::Log.info("Setting synchronization cluster mark #{mark} on #{founder_name}.")
       CrowbarPacemakerCIBAttribute.set(node[:hostname], attribute, "1")
     end
   end
@@ -111,25 +120,30 @@ module CrowbarPacemakerSynchronization
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
 
     attribute = "#{prefix}#{mark}"
+    cluster_name = CrowbarPacemakerHelper.cluster_name(node)
+    cluster_nodes = CrowbarPacemakerHelper.cluster_nodes_names(node)
+    nodes_with_mark_set = []
 
     # non-founders simply set the mark and then wait for the founder to set the
     # mark
     unless CrowbarPacemakerHelper.is_cluster_founder?(node)
-      Chef::Log.info("Setting synchronization cluster mark #{mark}.")
-      CrowbarPacemakerCIBAttribute.set(node[:hostname], "#{prefix}#{mark}", "1")
+      Chef::Log.info("Setting synchronization cluster mark #{mark} on #{node[:hostname]} " \
+        "for cluster #{cluster_name}.")
+      CrowbarPacemakerCIBAttribute.set(node[:hostname], attribute, "1")
       return wait_for_mark_from_founder(node, mark, fatal, timeout)
     end
 
     # founder waits for the mark to be set on all non-founders, and then sets
     # its mark; if the mark is already set, we can skip everything
     if CrowbarPacemakerCIBAttribute.get(node[:hostname], attribute, "0") != "0"
-      Chef::Log.info("Synchronization cluster mark #{mark} already set.")
+      Chef::Log.info("Synchronization cluster mark #{mark} already set on #{node[:hostname]} " \
+        "for cluster #{cluster_name}.")
       return
     end
 
     if CrowbarPacemakerHelper.being_upgraded?(node)
       Chef::Log.debug("Node is being upgraded." \
-        "Skipping wait loop for all other cluster nodes.")
+        "Skipping wait loop for all other cluster nodes for cluster #{cluster_name}.")
       return
     elsif !CrowbarPacemakerCIBAttribute.cib_up_for_node?
       if fatal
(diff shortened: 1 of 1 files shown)


jsuchome:
I agree we should do something about the error messages.

However I'm not sure if it makes sense to gather all errors once we already know of the first one. Typical case is that while a node has some real failure, others fail while waiting for this one to sync. It does not make much sense imho to wait after all nodes fail with the timeout error; it would make sense to print more prominently the real error that was behind.

vuntz:
> However I'm not sure if it makes sense to gather all errors once we already know of the first one.

Agreed, I'm concerned that this would really slow down things when an error occurs, since we may actually wait for the full timeout for all other nodes to finally return.

aspiers:
@jsuchome Sadly it wasn't recently, and it didn't get merged because someone rewrote the whole mechanism while it was under review, so now it would need a re-rewrite :-(  @matthewoliver see #202.

aspiers:
But yes, a huge +1 for improving the error message.  This urgently needs to be done, because currently it's pretty much the biggest FAQ in the product...

aspiers:
Also I agree with the concerns about speed stated by @jsuchome and @vuntz.

matthewoliver:
so 100 pull requests off (202 - 302) kinda link that symmetry. 
Cool, thanks for the reviews. and thanks Adam for pointing out 202. I've started to take a look. Will continue tomorrow. I'll see what either reviving 202 involves or merging the ideas into 302 depending on all the changes that have been made since and access.

matthewoliver:
He's the first attempt of pulling @aspiers 202 forward. Will merge the 2 commits before landing... if we get that far :)

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -144,25 +156,34 @@ def self.synchronize_on_mark(node, mark, fatal = false, timeout = 60)
...
+        Chef::Log.info("Checking if all other cluster nodes have set #{mark} " \
+          "on cluster #{cluster_name}...")
 
         Timeout.timeout(timeout) do
-          CrowbarPacemakerHelper.cluster_nodes_names(node).each do |name|
-            next if name == node[:hostname]
+          cluster_nodes.each do |name|
+            if name == node[:hostname]
+              nodes_with_mark_set << name
+              next
```
aspiers: Missing `end` here, as shown by the syntax error in the Travis run.

matthewoliver:
Lol, opps, sorry, been off with gastro, and this was the last patch I
pushed before I gave up.

The pregnant wife now has it, so am currently in a hospital as she gets
rehydrated. Fun times... I'll get back and fix this when I get a chance.

Matt

aspiers:
@matthewoliver Sorry to hear that - don't worry I just pushed a hopefully fixed version.  Hope you both get better soon!

matthewoliver:
@aspiers thanks man!

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -64,28 +64,34 @@ def self.wait_for_mark_from_founder(node, mark, fatal = false, timeout = 60)
...
           sleep(5)
         end # loop
       end # Timeout
     rescue Timeout::Error
       if fatal
-        message = "Cluster founder didn't set #{mark}!"
+        message = \
+          "Cluster founder #{founder_name} didn't set #{mark} on cluster #{cluster_name}! " \
+          "Timed out while waiting for the founder; Please check the chef-client logs on " \
+          "#{founder_name} to see what went wrong."
```
aspiers: The chef-client logs in question do not live on the founder; they live on the admin node under `/var/log/crowbar/chef-client`.  This is a common misunderstanding of operators so I think we could tweak the error message to help clarify that.

The `/var/log/chef/client.log` files on each non-admin node are only for the periodic `chef-client` runs which AFAIK do not perform synchronisation, or at least didn't used to - I've just pinged @vuntz to check about this since the mechanism got changed quite a bit since I last looked at it closely (the marks are now stored in the CIB and are just boolean rather than numeric).

BTW at some point in the future the periodic runs may be ditched entirely.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -144,25 +156,36 @@ def self.synchronize_on_mark(node, mark, fatal = false, timeout = 60)
...
               sleep(5)
             end
           end # each
         end # Timeout
       rescue Timeout::Error
+        remaining = cluster_nodes - nodes_with_mark_set
         if fatal
-          message = "Some cluster nodes didn't set #{mark}!"
+          message = "Some cluster nodes didn't set #{mark} on cluster #{cluster_name}: " \
+            "#{remaining.join(" ")}. Please check chef-client logs for remaining nodes."
```
aspiers: Again it would be good to clarify exactly where to look for these logs because it's not obvious.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -64,28 +64,34 @@ def self.wait_for_mark_from_founder(node, mark, fatal = false, timeout = 60)
...
+            "on cluster #{cluster_name}...")
           sleep(5)
         end # loop
       end # Timeout
     rescue Timeout::Error
       if fatal
-        message = "Cluster founder didn't set #{mark}!"
+        message = \
+          "Cluster founder #{founder_name} didn't set #{mark} on cluster #{cluster_name}! " \
+          "Timed out while waiting for the founder; Please check the chef-client logs on " \
```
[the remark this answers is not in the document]
matthewoliver: Done, cheers

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
matthewoliver: OK, updated. Make reference to 'crowbar chef-client logs' to make sure they are differentiated from the normal chef-logs and to help direct to '/var/log/crowbar/chef-client'. I guess I could have just given the path. But meh.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -144,25 +156,37 @@ def self.synchronize_on_mark(node, mark, fatal = false, timeout = 60)
...
+            end
             loop do
-              break if CrowbarPacemakerCIBAttribute.get(name, attribute, "0") != "0"
-              Chef::Log.debug("Waiting for all other cluster nodes to set #{mark}...")
+              if CrowbarPacemakerCIBAttribute.get(name, attribute, "0") != "0"
+                nodes_with_mark_set << name
+                break
+              end
+              Chef::Log.debug("Waiting for all other cluster nodes to set #{mark} " \
+                "on cluster #{cluster_name}...")
```
jsuchome: In this loop you're just waiting for an attribute state of specific node,  so I think it would make sense to add node's `name` into the log message.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
```diff
@@ -144,25 +156,37 @@ def self.synchronize_on_mark(node, mark, fatal = false, timeout = 60)
...
               sleep(5)
             end
           end # each
         end # Timeout
       rescue Timeout::Error
+        remaining = cluster_nodes - nodes_with_mark_set
```
jsuchome: looking at the way the loop is written, It's actually possible that from nodes `A,B,C`, `A` went fine, loop timed out while waiting on node `B`, while node `C` could be correctly set, but there was no time to check. So it's not _exactly correct_ to say all remaining nodes do not have the cluster marks set.

Anyway, I would probably let this go, your PR is already making current state better.

aspiers:
> Adam Spiers 4:29 PM is there still something preventing synchronization in chef-client periodic runs?
> Vincent Untz 7:07 AM well, the fact that the chef runs may not run at the same time?
> Adam Spiers 11:25 AM No, I mean is there something preventing synchronization checks happening, not is there something preventing them succeeding 🙂
> Adam Spiers 11:25 AM e.g. if a proposal fails and leaves nodes unsynchronised, what happens at the next periodic run?
> Vincent Untz 12:31 PM ah, yes, they will slowly get synchronized
> Vincent Untz 12:32 PM possibly as soon as with the next chef-periodic client run
> Vincent Untz 12:33 PM the only difference between the "apply" chef-client and the periodic one is really "running on all nodes in parallel" vs "running but no idea what happens for the other members of the cluster"
> Adam Spiers 12:33 PM hrm
> Adam Spiers 12:33 PM that's not very good
> Adam Spiers 12:33 PM I thought we used to skip synchronisation on periodic runs
> Vincent Untz 12:34 PM if the nodes have been synchronized, then they will be skipped, yes
> Vincent Untz 12:34 PM they will only be used if the nodes were not synchronized
> Vincent Untz 12:34 PM (or if the sync marks have been reset)
> Vincent Untz 12:35 PM the reset is something that only happens when applying, though
> Vincent Untz 12:35 PM so what this really means is that periodic chef runs will only block on this if "apply" failed

So the error message *also* needs to cater for the possibility that:

- Applying the proposal failed
- A periodic `chef-client` run subsequently happened, and failed due to periodic runs not being run in a fashion which is orchestrated across nodes (they just run semi-randomly roughly every 900 seconds)

In this case the synchronisation failure will appear in `/var/log/chef/client.log` on the failing node, not in `/var/log/crowbar/chef-client` on the admin node.

matthewoliver:
thanks guys, I'm playing with attempting to figure out if it's an apply or a periodic run. I playing with inserting some meta-data from the crowbar side in pre_chef_call and removing it in post_chef_call, and I can then use to direct the message. Though I might be over thinking it. I'm new to the actual crowbar framework side, so lets see how this goes.

If all else fails I'll just add both locations to the error message ;)

aspiers:
@matthewoliver Sounds like a great idea, thanks for trying that out.  I agree that if it proves impossible (which sounds unlikely) then just adding both locations to the message would be an acceptable workaround.

aspiers:
I just noticed this was not in SOC8 yet and looked here to find out why, only to discover it's apparently my fault for not approving it yet.  Bah ;-)  We really really should get this in an update ASAP, since it significantly improves the helpfulness of the product's most common failure case!

aspiers:
[**@aspiers**](https://github.com/aspiers) commented on [25 Apr 2018, 14:00 BST](https://github.com/crowbar/crowbar-ha/pull/302#issuecomment-384277729 "2018-04-25T13:00:27Z - Replied by Github Reply Comments"):
> @matthewoliver Sounds like a great idea, thanks for trying that out.  I agree that if it proves impossible (which sounds unlikely) then just adding both locations to the message would be an acceptable workaround.

I suggest that you don't allow yourself to fall victim to my well-intentioned but misplaced perfectionism.  Better to add both locations to the message than for this PR to never get merged ;-)  Is this something you could do soon so we can quickly merge it?  We can always refine later.

matthewoliver:
This is a rebase _and_ simply adding both possible locations of the logs.. because Adam is right, it's better we correct and get something in that's better then spend too much time engineering the perfect solution, which at the moment I don't have to time for.

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
aspiers: @matthewoliver I think @jsuchome is right here and this could significantly boost the helpfulness of the message - any chance you could address?

aspiers:
mkcloud failed for unrelated reasons (issue getting the ISO). Have triggered a rebuild

chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
matthewoliver: Done

matthewoliver:
Thanks @jsuchome and @aspiers, fixed the debug message that Jiri pointed out.

aspiers:
Caused: java.io.IOException: Backing channel 'JNLP4-connect connection from mkchl.cloud.suse.de/10.162.191.72:48406' is disconnected.

Triggered a rebuild

aspiers:
@jsuchome @nicolasbock Let's get this merged ASAP as the existing error message continues to regularly confuse users!! Please can you re-review?

jsuchome:
sorry it took so long...

aspiers:
Any volunteers to backport this?

matthewoliver:
Just posted some backports:

stable/5.0 (crowbar 8): https://github.com/crowbar/crowbar-ha/pull/358
stable/4.0 (crowbar 7): https://github.com/crowbar/crowbar-ha/pull/359

Unless I have my stable versions mixed.

PR #303, opened 2018-03-12
Update corosync configuration correctly and in time when adding a new cluster node

jsuchome:
When a new node is added to the cluster, waiting for corosync reload after the end of chef-client run (`:delayed` action) is too late.

With current implementation of pacemaker sync marks, crm_attribute calls are used and they fail when corosync does not have the full information about all its nodes.

See https://github.com/crowbar/crowbar-ha/pull/305 for (partially) alternative take


jsuchome:
Closing in favor of https://github.com/crowbar/crowbar-ha/pull/305

PR #304, opened 2018-03-13
haproxy: active-active mode

ilausuch:
Deploys the haproxy in active-active mode. All nodes in cluster will have the haproxy.

Allows the migration from previous mode active-pasive

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
@@ -30,6 +30,21 @@ include_recipe "haproxy::setup"
 
 cluster_name = CrowbarPacemakerHelper.cluster_name(node)
 
+if node[:pacemaker][:haproxy][:clusters].key?(cluster_name) && node[:pacemaker][:haproxy][:clusters][cluster_name][:enabled]
+  nonlocal_bind_file = "/etc/sysctl.d/50-haproxy-nonlocal_bind.conf"
+  cookbook_file nonlocal_bind_file do
+    source "sysctl_nonlocal_bind.conf"
+    mode "0644"
+  end
+
+  # we need to reload immediately, as otherwise haproxy would fail to start
+  bash "reload nonlocal_bind-sysctl" do
+    code "/sbin/sysctl -e -q -p #{nonlocal_bind_file}"
+    action :nothing
+    subscribes :run, resources(cookbook_file: nonlocal_bind_file), :immediately
+  end
+end
+
 # Wait for all nodes to reach this point so we know that all nodes will have
 # all the required packages installed before we create the pacemaker
 # resources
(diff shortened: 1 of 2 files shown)


ilausuch:
[Make haproxy active/active](https://trello.com/c/7oHIy01J/271-make-haproxy-active-active)

chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
```diff
@@ -56,17 +74,13 @@
...
-    cib_objects lazy { transaction_objects }
-    # note that this will also automatically start the resources
-    action :commit_new
+  # Compatibility with existing deployment: we need to drop the group to create
+  # the clone
+  group_name = "g-#{service_name}"
+  pacemaker_group group_name do
+    action [:stop, :delete]
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+    only_if "crm configure show group_name"
```
AbelNavarro: `#{group_name}`

ilausuch: Done

AbelNavarro:
The tests show that when applying this patch, the connection to the API socket is down for about 4 seconds. The API socket is hold by haproxy, not by the actual API service.

Connections to the API service are down for 20 seconds when applying this patch, so it is a major change. Trials to reduce this time by including the deletion of the haproxy group with the haproxy clone in the same transaction have been unsuccessful, but anyway the improvement would have been in the best case minor.

chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
```diff
@@ -43,6 +43,24 @@
 end
 
 if node[:pacemaker][:haproxy][:clusters].key?(cluster_name) && node[:pacemaker][:haproxy][:clusters][cluster_name][:enabled]
+  directory "create /etc/sysctl.d for nonlocal_bind" do
```
aspiers: This shouldn't be necessary; it's provided by both `systemd` and `filesystem` rpms.

vuntz:
Can we get the commit message fixed? :-)

chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
ilausuch: Done

rhafer:
> Afaik, the ha_galera script completes before this line and there are sync marks assuring mysql should be running at this point, so I'm puzzled. Feedback is welcome.

@AbelNavarro  Have you looked into /var/log/messages if the galera resource agent is logging something helpful (it should if the fail count of the galera resource is > 0). Also /var/log/mysql/* might contain some helpful information. Do you still have a system where this is reproducable? We could do a debugging session.

Itxaka:
@aspiers re-review pls?

vuntz:
(trusting the reviews, merging to unblock)

PR #305, opened 2018-03-14
Update corosync configuration correctly and in time when adding a new cluster node

jsuchome:
When a new node is added to the cluster, waiting for corosync reload after the end of chef-client run (:delayed action) is too late.

With current implementation of pacemaker sync marks, crm_attribute calls are used and they fail when corosync does not have the full information about all its nodes. So we must somehow extend the first synchronization that happens before the sync marks mechanism is used.

~Alternative solution to https://github.com/crowbar/crowbar-ha/pull/303 , this one looks cleaner~

Also required for adding cluster node: https://github.com/crowbar/crowbar-openstack/pull/1583

### DIFF
--- a/chef/cookbooks/pacemaker/recipes/default.rb
+++ b/chef/cookbooks/pacemaker/recipes/default.rb
@@ -53,13 +53,36 @@ if (platform_family?("suse") && node.platform_version.to_f >= 12.0) || platform_
   end
 end
 
+cluster_size = node[:pacemaker][:elements]["pacemaker-cluster-member"].length
+nodes_names = node[:pacemaker][:elements]["pacemaker-cluster-member"].map do |n|
+  n.gsub(/\..*/, "")
+end
+
 ruby_block "wait for cluster to be online" do
   block do
     require "timeout"
     begin
       Timeout.timeout(60) do
-        cmd = "crm_mon -1 | grep -qi online"
-        while ! ::Kernel.system(cmd)
+        loop do
+          # example of 'crm_node -l' output:
+          # 1084813649 d52-54-77-77-01-02 member
+          # 1084813652 d52-54-77-77-01-01 member
+          crm_node_cmd = Mixlib::ShellOut.new("crm_node -l").run_command
+          if crm_node_cmd.exitstatus != 0
+            Chef::Log.warn("Problems when executing 'crm_node -l': #{crm_node_cmd.stderr}")
+            next
+          end
+          crm_nodes = crm_node_cmd.stdout
+          crm_names = crm_nodes.split("\n").map { |l| l.split(" ")[1] }
+          crm_mon_cmd = Mixlib::ShellOut.new("crm_mon -1").run_command
+          if crm_mon_cmd.exitstatus != 0
+            Chef::Log.warn("Problems when executing 'crm_mon -1': #{crm_mon_cmd.stderr}")
+            next
+          end
+          crm_mon = crm_mon_cmd.stdout
+          break if crm_mon.include?("#{cluster_size} nodes configured") &&
+              crm_mon.include?("Online:") &&
+              crm_names.sort == nodes_names.sort
           Chef::Log.debug("cluster not online yet")
           sleep(5)
         end
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -384,9 +384,11 @@ class PacemakerService < ServiceObject
     member_nodes = members.map { |n| NodeObject.find_node_by_name n }
     remotes = attributes["elements"]["pacemaker-remote"] || []
     remote_nodes = remotes.map { |n| NodeObject.find_node_by_name n }
+    old_members = []
 
     founder_name = nil
     founder = nil
+    cluster_members_changed = false
 
     # elect a founder
     unless members.empty?
@@ -397,13 +399,16 @@ class PacemakerService < ServiceObject
         old_founder_name = old_role.default_attributes["pacemaker"]["founder"]
         founder_name = old_founder_name if members.include?(old_founder_name)
 
+        old_members = old_attributes["elements"]["pacemaker-cluster-member"]
+
         # the founder from the old role is not there anymore; let's promote
         # another node to founder, so we get the same authkey
         if founder_name.nil?
-          old_members = old_attributes["elements"]["pacemaker-cluster-member"]
           old_members = old_members.select { |n| members.include? n }
           founder_name = old_members.first
         end
+
+        cluster_members_changed = members.sort != old_members.sort
       end
 
       # Still nothing, there are two options:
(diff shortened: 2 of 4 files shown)


jsuchome:
[P1: Bug 1083427 - Failed to add additional node to cluster with DRBD disabled](https://trello.com/c/QyyNGNpM/718-p1-bug-1083427-failed-to-add-additional-node-to-cluster-with-drbd-disabled)

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -397,13 +399,16 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+
         # the founder from the old role is not there anymore; let's promote
         # another node to founder, so we get the same authkey
         if founder_name.nil?
-          old_members = old_attributes["elements"]["pacemaker-cluster-member"]
           old_members = old_members.select { |n| members.include? n }
           founder_name = old_members.first
         end
+
+        cluster_size_changed = members.sort != old_members.sort
```
aspiers: This variable name is misleading - you are not only testing whether the size changed, but also whether members changed.  Presumably we also need the same behaviour if one member is replaced by another.  So I'd recommend changing the variable to `cluster_members_changed`.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -462,6 +467,8 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
       member_node[:drbd] ||= {}
       member_node[:drbd][:local_node_id] = is_founder ? 0 : 1
       member_node[:drbd][:remote_node_id] = is_founder ? 1 : 0
+      member_node[:crowbar_wall][:cluster_size_changed] =
```
aspiers: Same comment here about the attribute name.

chef/cookbooks/pacemaker/recipes/default.rb
```diff
@@ -53,13 +53,16 @@
...
+
 ruby_block "wait for cluster to be online" do
   block do
     require "timeout"
     begin
       Timeout.timeout(60) do
-        cmd = "crm_mon -1 | grep -qi online"
-        while ! ::Kernel.system(cmd)
+        cmd1 = "crm_mon -1 | grep -qi online"
+        cmd2 = "crm_mon -1 | grep -qi '#{cluster_size} nodes configured'"
```
aspiers: Since we are being more thorough here (which is good!), is it worth going a bit further and also checking the node names, to make sure we have the members we expect?

crowbar_framework/app/models/pacemaker_service.rb
aspiers: This variable name is misleading - you are not only testing whether the size changed, but also whether members changed.  Presumably we also need the same behaviour if one member is replaced by another.  So I'd recommend changing the variable to `cluster_members_changed`.

crowbar_framework/app/models/pacemaker_service.rb
aspiers: Same comment here about the attribute name.

chef/cookbooks/pacemaker/recipes/default.rb
jsuchome: Hm, good point. I'll  try to test it in my new deployment.

I'm just bit concerned that the check might get too complicated (or less readable)...

crowbar_framework/app/models/pacemaker_service.rb
jsuchome: True

jsuchome:
@aspiers so I've added another check for members, what do you think now?

chef/cookbooks/pacemaker/recipes/default.rb
```diff
@@ -53,13 +53,20 @@
...
+        cmd3 = "[ #{cluster_size} == $(crm node list | grep #{greps} | wc -l) ]"
```
aspiers: Please don't use `grep` here at all - it introduces a security hole in the web UI, and anyway it's easy to do searching strings and counting lines in native Ruby without needing to use extra shell pipelines.  Also you can just run `crm_mon -1` or `crm status` a single time via backticks and capture the output into a `String` for examination, which is more efficient than calling them 3 times in a row.  I think this will make the code more readable too, since currently it's not immediately obvious what `greps` interpolates to.

chef/cookbooks/pacemaker/recipes/default.rb
```diff
@@ -53,13 +53,26 @@
...
       Timeout.timeout(60) do
-        cmd = "crm_mon -1 | grep -qi online"
-        while ! ::Kernel.system(cmd)
+        loop do
+          # example of 'crm_node -l' output:
+          # 1084813649 d52-54-77-77-01-02 member
+          # 1084813652 d52-54-77-77-01-01 member
+          crm_nodes = Mixlib::ShellOut.new("crm_node -l").run_command.stdout
+          crm_names = crm_nodes.split("\n").map { |l| l.split(" ")[1] }
+          crm_mon = Mixlib::ShellOut.new("crm_mon -1").run_command.stdout
```
aspiers: Thanks, definitely looks better than before.

If for some reason `crm_node` or `crm_mon` are not found on the path then `crm_nodes` or `crm_mon` will be `nil`, causing a `NoMethodError` on `nil`.  Maybe it's an unlikely scenario but it's almost always better to have defensive programming which gracefully handles the worse case.

aspiers: BTW I don't see any problem with using `Mixlib::ShellOut` but I'm curious why you chose that over normal Ruby backticks?

Itxaka: Im guessing chef style, which recommends shellout over everything: https://docs.chef.io/ruby.html#shelling-out

aspiers: Thanks for the link @Itxaka!  I wish that document would explain the rationale for that advice though, rather than expecting everyone to blindly trust it.  I'm sure there is a good reason, I just don't know what it is.

Itxaka: If I had to guess I would say that the list of options that it provides makes it useful to use it over backticks[0], plus its easier to mock in tests :)

You also get a really nice output in case of an exception, which is awesome to try and reproduce the same command[1]

[0] https://github.com/chef/mixlib-shellout/blob/master/lib/mixlib/shellout.rb#L58
[1] https://github.com/chef/mixlib-shellout/blob/master/lib/mixlib/shellout.rb#L229

aspiers: @Itxaka Thanks, useful info, but as usual I wish it was documented somewhere more discoverable than buried deep inside a PR review. Not a criticism of you at all, just a comment about Chef documentation...

Itxaka: Agreed on both cases. Should be documented somewhere and chef docs are the worst.

PR #306, opened 2018-03-29
[4.0] pacemaker: added support for op defaults

AbelNavarro:
backport of https://github.com/crowbar/crowbar-ha/pull/297

on-fail" action.

other resourced.

be extended later on.

(cherry picked from commit e3a6e4c1445986e9688c50fc065afea5a56de71e)

### DIFF
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -54,14 +54,19 @@
         "admin_name": "",
         "public_name": ""
       },
-      "clone_stateless_services": true
+      "clone_stateless_services": true,
+      "op_defaults": {
+        "monitor": {
+          "on-fail": ""
+        }
+      }
     }
   },
   "deployment": {
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 104,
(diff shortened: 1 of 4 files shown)


AbelNavarro:

aspiers:
@AbelNavarro Please can you follow [the established process for handling backports](https://etherpad.nue.suse.com/p/cloud-PR-porting)?  (Yes, I know we need to turn this into a proper document in GitHub. Volunteers welcome.)

AbelNavarro:
@aspiers thanks for the feedback. I made the changes as suggested by the backport guidelines.

Notice that there are no broken URLs in the description :-P

chef/data_bags/crowbar/template-pacemaker.json
```diff
@@ -54,14 +54,19 @@
         "admin_name": "",
         "public_name": ""
       },
-      "clone_stateless_services": true
+      "clone_stateless_services": false,
```
vuntz: Careful: you're changing the default here! It's different in stable/4.0.

AbelNavarro: nice catch, fixing that.

AbelNavarro:
@aspiers mind having another look at this PR?

PR #308, opened 2018-04-09
[4.0] Fix a search for cluster founders.

jsuchome:
[:pacemaker][:founder] was changed from boolean to string recently

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -22,8 +22,8 @@ module Api
         unless repocheck["ha"]["available"]
           return { errors: [I18n.t("api.pacemaker.ha_not_installed")] }
         end
-        founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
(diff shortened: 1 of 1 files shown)


Itxaka:
@jsuchome maybe we want to have the same in both 4.0 as master:

https://github.com/crowbar/crowbar-ha/commit/54c9d4204dcc063d9aa094b0c4213be68d2f47b2#diff-e7b8b7bb6c0a774ed453e208a7d19635R25

And 

https://github.com/crowbar/crowbar-ha/commit/54c9d4204dcc063d9aa094b0c4213be68d2f47b2#diff-e7b8b7bb6c0a774ed453e208a7d19635R47

Doesnt matter to me if those changes are cherry-picked from master to 4.0 or this PR is merged and then pushed to master, but having the same code on both is desired IMO.

jsuchome:
@Itxaka I'll just manually sync this `4.0` code with master one...

bbobrov:
I think that this introduced a regression. I am getting `undefined method `find' for Api::Node:Class` when trying to run prechecks for 7->8 upgrades. Reverting this patch helps.

jsuchome:
@bbobrov you're right, but it's already fixed with https://github.com/crowbar/crowbar-ha/pull/310

PR #311, opened 2018-05-29
pacemaker: allow multiple meta parameters (bsc#1093898)

aplanas:
Based on 4e944918b7df65a25b8737c41ba5ee5558f9c6f1.

There is the chance that multiple 'meta' parameters appear in
a CIB resource. This procude an exception in 'extract_hash`
method.

This patch allows multiple 'meta' paramenter in the result.


aplanas:
Abandoned

aplanas:
@aspiers Sure thing. The use case is when a resource is moved out from maintenance. A new `meta` attribute appears with `maintenance=false`. You can see the exact case in the bug report (looks like that is about neutron-ha-tool replacement)

The problem with this PR is about the test case. Is wrong and I need to redo it.

aspiers:
Thanks.  Please can you provide a concrete real-world example of multiple metas _in the commit message_?  Crowbar is an open project but not everyone has access to bugzilla - and even for those who do, it saves each reviewer having to log into bugzilla and search the comments for the relevant details.

aplanas:
@aspiers I think that I addressed the issues, can you re-review?

aplanas:
[P3 Bug 1093898: Resource with multiple meta entries breaks crowbar](https://trello.com/c/8Q8Yco09/975-p3-bug-1093898-resource-with-multiple-meta-entries-breaks-crowbar)

PR #317, opened 2018-07-26
[4.0] Populate the node names when drbd is enabled

s-t-e-v-e-n-k:
To allow us to set up location constraints for the database and rabbitmq
roles when drbd is enabled and we then grow the cluster to 3 nodes to allow
us to migrate to MariaDB, populate what the nodes are so we can force
DRBD related cluster roles onto those nodes only.

This currently doesn't work -- the intent is [:drbd][:nodes] will be set to the two nodes in the cluster and then not change when a third cluster member is added.

### DIFF
--- a/chef/data_bags/crowbar/migrate/pacemaker/106_populate_drbd_nodes.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/106_populate_drbd_nodes.rb
@@ -1,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  if a["drbd"]["enabled"]
+    a["drbd"]["nodes"] = d["elements"]["pacemaker-cluster-member"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["drbd"].delete("nodes")
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-pacemaker.schema
+++ b/chef/data_bags/crowbar/template-pacemaker.schema
@@ -130,7 +130,8 @@
               "required": true,
               "mapping": {
                 "enabled": { "type": "bool", "required": true },
-                "shared_secret": { "type": "str", "required": true }
+                "shared_secret": { "type": "str", "required": true },
+                "nodes": { "type": "seq", "sequence": [ { "type": "str" } ] }
               }
             },
             "haproxy": {
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -461,12 +461,20 @@ class PacemakerService < ServiceObject
     role.default_attributes["drbd"]["common"]["net"] ||= {}
     role.default_attributes["drbd"]["common"]["net"]["shared_secret"] = \
       role.default_attributes["pacemaker"]["drbd"]["shared_secret"]
+    if member_nodes.length == 2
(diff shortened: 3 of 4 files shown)


crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -467,6 +467,12 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
       member_node[:drbd] ||= {}
       member_node[:drbd][:local_node_id] = is_founder ? 0 : 1
       member_node[:drbd][:remote_node_id] = is_founder ? 1 : 0
+      unless old_members.empty?
```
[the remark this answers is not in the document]
jsuchome: I guess there really should be `if old_members.empty?` here (without switching the branches)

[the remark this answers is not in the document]
jsuchome: Or even better. Maybe you should not check for presence of `old_members`, but rather check if the content is the same as `member_nodes`. 
So the unlikely case of user changing the cluster nodes (and keeping the size) does not break the logic.

chef/data_bags/crowbar/migrate/pacemaker/106_populate_drbd_nodes.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  if a["drbd"]["enabled"]
+    a["drbd"]["nodes"] = d["elements"]["pacemaker-cluster-member"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["drbd"].delete("nodes") if a["drbd"]["enabled"]
```
jsuchome: This condition looks weird. It should be deleted if it is present, regardless the state of drbd enablement.

jsuchome:
Looks good. Would be nice to see relevant changes that actually use these values.

chef/data_bags/crowbar/template-pacemaker.schema
```diff
@@ -130,7 +130,8 @@
               "required": true,
               "mapping": {
                 "enabled": { "type": "bool", "required": true },
-                "shared_secret": { "type": "str", "required": true }
+                "shared_secret": { "type": "str", "required": true },
+                "nodes": { "type": "seq", "sequence": [ { "type": "str" } ] }
```
jsuchome: What about `drbd[:nodes]` in template-pacemaker.json?

s-t-e-v-e-n-k: Why? It isn't marked as required, so the JSON doesn't need to specify it.

s-t-e-v-e-n-k:
I am going to close this due to pacemaker always probing.

jsuchome:
This seems to be still useful, to know the nodes which need/do not need proper DRBD setup.

Looks like simple 
```
if node[:pacemaker][:drbd][:enabled] && node[:drbd][:nodes].include? node.name
```
 
in 
https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/recipes/default.rb#L122

jsuchome:
Hm, not enough, we're gonna need some changes in postgresql recipes as well...

jsuchome:
Something like this https://gist.github.com/jsuchome/5cfb91a48f295fc387615ba48269bef0 (would need a cleanup)

PR #318, opened 2018-08-01
[4.0] DRBD: fix when reinstall node

ilausuch:
When a node is reinstalled the DRBD resources don't are configured
because of the node attribute drbd.rsc.<resource_name>.configured
is marked as true, and the setup is skiped.

This solution store a counter instead of true and false. In each
reinstalation this counter is increased by crowbar. So checking
if the current reinstalation number is equal to the configured
stored in the node attributes it can skip when it's necesary
and setup the DRBD resources when it's needed.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/providers/drbd_create_internal.rb
+++ b/chef/cookbooks/crowbar-pacemaker/providers/drbd_create_internal.rb
@@ -45,7 +45,7 @@ action :create do
       end.run_action(:install)
     end
 
-    next if resource["configured"]
+    next if resource["configured"] == node.crowbar.state_debug.installed
 
     lvm_logical_volume resource_name do
       group new_resource.lvm_group
@@ -64,7 +64,7 @@ action :create do
     end.run_action(:create)
 
     # we know we have to save due to the "next" earlier on
-    node.set["drbd"]["rsc"][resource_name]["configured"] = true
+    node.set["drbd"]["rsc"][resource_name]["configured"] = node.crowbar.state_debug.installed
     node.save
   end
 


ilausuch:
[Reinstalling a DRBD member](https://trello.com/c/l3cikeSq/225-reinstalling-a-drbd-member)

chef/cookbooks/crowbar-pacemaker/providers/drbd_create_internal.rb
```diff
@@ -64,7 +64,7 @@
     end.run_action(:create)
 
     # we know we have to save due to the "next" earlier on
-    node.set["drbd"]["rsc"][resource_name]["configured"] = true
+    node.set["drbd"]["rsc"][resource_name]["configured"] = node.crowbar.state_debug.installed
```
nicolasbock: Could you point out where this attribute is set and incremented? Sorry I might not have had sufficient caffeine to find it myself :wink:

Itxaka: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/crowbar_service.rb#L143

A bit hidden :)

nicolasbock: :smile: Ah, ok, I had seen that code but had mentally added quotes around the `state`. But it's `["state_debug"][state]` and not `["state_debug"]["state"]`. :man_facepalming:

ilausuch: Thanks

PR #320, opened 2018-08-08
pacemaker: Fix version comparsion on cib schema

dirkmueller:
The schema version is a multi digit number that shouldn't be
compared as a string, because otherwise "2.10" < "2.9".

### DIFF
--- a/chef/cookbooks/pacemaker/recipes/setup.rb
+++ b/chef/cookbooks/pacemaker/recipes/setup.rb
@@ -49,7 +49,10 @@ execute "upgrade CIB syntax" do
   command "crm configure upgrade force"
   not_if do
     cib_meta = Mixlib::ShellOut.new(cib_cmd).run_command.stdout.chomp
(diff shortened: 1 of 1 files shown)


chef/cookbooks/pacemaker/recipes/setup.rb
```diff
@@ -44,12 +44,18 @@
 # <cib epoch="41" num_updates="9" admin_epoch="0" validate-with="pacemaker-1.2" ...>
 cib_cmd = "cibadmin -Q -n"
 
+cib_meta = Mixlib::ShellOut.new(cib_cmd).run_command.stdout.chomp
+cib_meta.gsub!(/.*validate-with="[^"]*-([^-"]*)".*/, "\\1")
+
+cib_version = cib_meta.split(".").map{|s|s.to_i}
```
aspiers: Maybe use `Gem::Version` instead?

https://medium.com/@edgar/how-to-compare-semantic-version-strings-in-ruby-95fbf067de32

PR #323, opened 2018-08-10
[4.0] drbd: Add switch to short circuit cluster size check

s-t-e-v-e-n-k:
Currently, DRBD enabled clusters are forced to be only two nodes. To
fascitate upgrades from SOC7 to Cloud 8, we need to switching from
PostgreSQL to MariaDB with Galera, which requires three node clusters at
a minimum.

### DIFF
--- a/chef/data_bags/crowbar/migrate/pacemaker/106_allow_larger_cluster.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/106_allow_larger_cluster.rb
@@ -1,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  unless a["drbd"].key? "allow_larger_cluster"
+    a["drbd"]["allow_larger_cluster"] = ta["drbd"]["allow_larger_cluster"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta["drbd"].key? "allow_larger_cluster"
+    a["drbd"].delete("allow_larger_cluster")
+  end
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -48,7 +48,8 @@
       },
       "drbd": {
         "enabled": false,
-        "shared_secret": ""
+        "shared_secret": "",
+        "allow_larger_cluster": false
       },
       "haproxy": {
         "admin_name": "",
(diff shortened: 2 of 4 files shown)


jsuchome:
Isn't there a word missing in your commit message? _we need to switching from ..._

s-t-e-v-e-n-k:
Extra 'ing', sadly. I've pushed up a correction.

chef/data_bags/crowbar/migrate/pacemaker/106_allow_larger_cluster.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["drbd"]["allow_larger_cluster"] = ta["drbd"]["allow_larger_cluster"]
```
nicolasbock: We usually add a guard here in case this attribute is already there:

~~~
a["drbd"]["allow_larger_cluster"] = ta["drbd"]["allow_larger_cluster"] unless a["drdb"].key? "allow_larger_cluster"
~~~

chef/data_bags/crowbar/migrate/pacemaker/106_allow_larger_cluster.rb
```diff
@@ -0,0 +1,9 @@
...
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["drbd"].delete("allow_larger_cluster")
```
nicolasbock: Same here:

~~~
a["drbd"].delete("allow_larger_cluster") unless ta["drdb"].key? "allow_larger_cluster"
~~~

matthewoliver:
Looking good, just running it through a test on a mkcloud host. If all is well I will approve :)

chef/data_bags/crowbar/migrate/pacemaker/106_allow_larger_cluster.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  unless a["drbd"].key? "allow_larger_cluster"
+    a["drbd"]["allow_larger_cluster"] = ta["drbd"]["allow_larger_cluster"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless a["drbd"].key? "allow_larger_cluster"
```
dirkmueller: this should be ta["drbd"].key?

PR #324, opened 2018-08-29
[4.0] crowbar-pacemaker: Improve cluster_nodes lookup

skazi0:
Chef search based lookup doesn't include nodes which will join cluster
in current chef-client run. This is because `roles` field is only updated
when chef-client is run.


skazi0:
too many unwanted side effects

PR #327, opened 2018-09-03
Adding a node into DRBD cluster

jsuchome:
We have to allow user to add a new node into 2-node DRBD cluster, so that such cluster can hold (3-node) galera configuration later.

But as (our version of) DRBD needs exactly 2 nodes, there are some tweaks necessary so that the extra node does not have DRBD configured, although it is part of the pacemaker cluster.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
@@ -37,6 +37,13 @@ module CrowbarPacemakerHelper
     ["prepare-os-upgrade", "done_os_upgrade"].include? upgrade_step
   end
 
+  # Does the node have DRBD setup? This is needed when we want to find out
+  # if a node in drbd-enabled cluster has really DRBD nor not.
+  def self.drbd_node?(node)
+    return false unless cluster_enabled?(node)
+    node[:pacemaker][:drbd] && node[:pacemaker][:drbd][:nodes] && node[:pacemaker][:drbd][:nodes].include?(node[:fqdn])
+  end
+
   # Returns the number of corosync (or non-remote) nodes in the cluster.
   def self.num_corosync_nodes(node)
     return 0 unless cluster_enabled?(node)
--- a/chef/data_bags/crowbar/migrate/pacemaker/107_populate_drbd_nodes.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/107_populate_drbd_nodes.rb
@@ -1,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  if a["drbd"]["enabled"]
+    a["drbd"]["nodes"] = d["elements"]["pacemaker-cluster-member"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["drbd"].delete("nodes")
+  return a, d
+end
--- a/chef/data_bags/crowbar/migrate/pacemaker/107_populate_drbd_nodes.rb
+++ b/chef/data_bags/crowbar/migrate/pacemaker/107_populate_drbd_nodes.rb
@@ -1,7 +1,15 @@
 def upgrade(ta, td, a, d)
   if a["drbd"]["enabled"]
     a["drbd"]["nodes"] = d["elements"]["pacemaker-cluster-member"]
+
+    d["elements"]["pacemaker-cluster-member"].each do |name|
+      node = Node.find_by_name(name)
+      next if node.nil? || !node[:pacemaker][:drbd][:enabled]
+      node[:pacemaker][:attributes]["drbd-controller"] = true
+      node.save
+    end
   end
+
   return a, d
 end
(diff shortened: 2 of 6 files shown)


chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -37,6 +37,14 @@ def self.being_upgraded?(node)
     ["prepare-os-upgrade", "done_os_upgrade"].include? upgrade_step
   end
 
+  # Does the node have DRBD setup? Usable to find out
+  # if a node in drbd-enabled cluster has DRBD nor not
+  def self.drbd_node?(node)
+    return false unless cluster_enabled?(node)
+    node[:pacemaker][:drbd] && node[:pacemaker][:drbd][:nodes] &&
+      node[:pacemaker][:drbd][:nodes].include?(node[:fqdn])
```
s-t-e-v-e-n-k: These currently evaluate to nil for any member, no matter what. node[:drbd][:nodes], you don't want [:pacemaker] in there.

jsuchome: Hm, I believe I need `node[:pacemaker]`, at least that's where it is in my tests...

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
```diff
@@ -37,6 +37,14 @@ def self.being_upgraded?(node)
     ["prepare-os-upgrade", "done_os_upgrade"].include? upgrade_step
   end
 
+  # Does the node have DRBD setup? Usable to find out
```
skazi0: useful?

chef/data_bags/crowbar/migrate/pacemaker/107_populate_drbd_nodes.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  if a["drbd"]["enabled"]
+    a["drbd"]["nodes"] = d["elements"]["pacemaker-cluster-member"]
```
skazi0: so we assume that all existing clusters with drbd are 2-node?

jsuchome: There's no other way

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
jsuchome: True

chef/cookbooks/crowbar-pacemaker/libraries/helpers.rb
jsuchome: But you must be right...

chef/data_bags/crowbar/migrate/pacemaker/107_populate_drbd_nodes.rb
```diff
@@ -1,7 +1,16 @@
 def upgrade(ta, td, a, d)
   if a["drbd"]["enabled"]
     a["drbd"]["nodes"] = d["elements"]["pacemaker-cluster-member"]
+
+  end
+
+  d["elements"]["pacemaker-cluster-member"].each do |name|
+    node = Node.find_by_name(name)
+    next if node.nil? || !node[:pacemaker][:drbd][:enabled]
```
rhafer: Isn't it safer to look at `a[drbd][enabled]` here`

rhafer: Why do you need to check `[:pacemaker][:drbd][:enabled]` on the node? It's a role attribute (of the proposal role), isn't it? Or do we overwrite it somewhere at the node level?

jsuchome: I guess I don't need i now when I moved this into the inner block.

I was concerned about the possibility of having more clusters and not all of them drbd-enabled, so I thought extra check on node should not hurt.

crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -461,12 +461,25 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+      drbd_nodes = []
+    end
     # set node IDs for drbd metadata
     member_nodes.each do |member_node|
       is_founder = (member_node.name == founder_name)
       member_node[:drbd] ||= {}
       member_node[:drbd][:local_node_id] = is_founder ? 0 : 1
       member_node[:drbd][:remote_node_id] = is_founder ? 1 : 0
+      member_node[:pacemaker] ||= {}
+      member_node[:pacemaker][:drbd] ||= {}
```
s-t-e-v-e-n-k: Why are we putting this under [:pacemaker]? Everything else in this block uses [:drbd] directly.

PR #330, opened 2018-09-11
Set the value of drbd nodes only for the first deployment.

jsuchome:
Also, do not allow more nodes in drbd cluster in the initial deployment.

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -461,13 +461,16 @@ class PacemakerService < ServiceObject
     role.default_attributes["drbd"]["common"]["net"] ||= {}
     role.default_attributes["drbd"]["common"]["net"]["shared_secret"] = \
       role.default_attributes["pacemaker"]["drbd"]["shared_secret"]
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -463,20 +463,17 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
       role.default_attributes["pacemaker"]["drbd"]["shared_secret"]
     if member_nodes.length == 2
       drbd_nodes = member_nodes.map(&:name)
-    elsif founder[:pacemaker].key?("drbd") && founder[:pacemaker][:drbd].key?("nodes")
-      drbd_nodes = founder[:pacemaker][:drbd][:nodes]
-    else
-      drbd_nodes = []
+    elsif old_role && old_role.default_attributes.key?("drbd")
+      drbd_nodes = old_role.default_attributes["drbd"]["nodes"] || []
```
rhafer: I think you still need an `else drbd_node = []` branch here. Because if this is an initial deployment of a 3 node cluster neither the `if` nor the `elsif` conditions will match and `drbd_nodes` basically undefined then. Unless I am missing something obvious.

PR #331, opened 2018-09-24
[4.0] upgrade: Add a healthcheck for pacemaker node state

jsuchome:
It could end badly if we proceed with the upgrade while some node
is in maintenance state (stopping corosync on such node might lead
to pacemaker's confusion about the services running there).

So let's add a specific check that is executed before the upgrade
and which does not allow user to upgrade such setup.

### DIFF
--- a/crowbar_framework/app/models/api/pacemaker.rb
+++ b/crowbar_framework/app/models/api/pacemaker.rb
@@ -43,6 +43,7 @@ module Api
         ret = {}
         crm_failures = {}
         failed_actions = {}
+        unready_nodes = {}
 
         # get unique list of founder names across all clusters
         cluster_founders_names = ::Node.find(
@@ -68,9 +69,12 @@ module Api
             failed_actions[n.name] = "#{n.name}: #{ssh_retval[:stdout]}"
             failed_actions[n.name] << " #{ssh_retval[:stderr]}" unless ssh_retval[:stderr].blank?
           end
+          ssh_retval = n.run_ssh_cmd('crm status | grep "^Node" | grep -E "maintenance|standby"')
+          unready_nodes[n.name] = ssh_retval[:stdout].chomp if ssh_retval[:exit_code].zero?
         end
         ret["crm_failures"] = crm_failures unless crm_failures.empty?
         ret["failed_actions"] = failed_actions unless failed_actions.empty?
+        ret["unready_nodes"] = unready_nodes unless unready_nodes.empty?
         ret
       end
 


crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -68,9 +69,12 @@ def health_report
             failed_actions[n.name] = "#{n.name}: #{ssh_retval[:stdout]}"
             failed_actions[n.name] << " #{ssh_retval[:stderr]}" unless ssh_retval[:stderr].blank?
           end
+          ssh_retval = n.run_ssh_cmd('crm status | grep "^Node" | grep -E "node_state|standby"')
```
skazi0: What is expected output here?
Here's what I got on my setup:
root@d52-54-77-77-01-01:~ # crm status | grep ^Node
Node d52-54-77-77-01-01: maintenance
Node d52-54-77-77-01-02: maintenance
root@d52-54-77-77-01-01:~ # crm status | grep ^Node  | grep -E "node_state|standby"
root@d52-54-77-77-01-01:~ # crm status | grep ^Node
Node d52-54-77-77-01-01: maintenance
Node d52-54-77-77-01-02: maintenance

jsuchome: Oh, grepping for `node_state` is obviously wrong. Result of some global replace command :-)

skazi0: Is there some "offline" state? What is reported if node is shutdown completely or not reachable?

jsuchome: If it's shutdown completely, other checks will fail, I assume (like the one checking if they are ready - I mean crowbar-ready, not pacemaker-ready)

crowbar_framework/app/models/api/pacemaker.rb
```diff
@@ -68,9 +69,12 @@ def health_report
             failed_actions[n.name] = "#{n.name}: #{ssh_retval[:stdout]}"
             failed_actions[n.name] << " #{ssh_retval[:stderr]}" unless ssh_retval[:stderr].blank?
           end
+          ssh_retval = n.run_ssh_cmd('crm status | grep "^Node" | grep -E "maintenance|standby"')
+          node_state[n.name] = ssh_retval[:stdout].chomp if ssh_retval[:exit_code].zero?
         end
         ret["crm_failures"] = crm_failures unless crm_failures.empty?
         ret["failed_actions"] = failed_actions unless failed_actions.empty?
+        ret["node_state"] = node_state unless node_state.empty?
```
rhafer: IMO `node_state` this is somewhat misleading as it's not necessarily the node in the dict key that is in maintenance/standby state. It can be any (or multiple) nodes of the cluster that the node is a member of.

As the new precheck uses this as a boolean value anyway couldn't we just rename it to `has_unready_nodes` or `all_nodes_ready` something similar?

jsuchome:
@rhafer updated

PR #336, opened 2018-10-10
[5.0] pacemaker: Cleanup 4.0 specific attribute

rhafer:
In the 4.0 branch with schema revision 106 we introduced the
["drbd"]["allow_larger_cluster"] attribute. That attribute is only
needed in 4.0 and was never forward ported to 5.0. So this needs to
be cleaned up during the upgrade from 4.0 to 5.0


skazi0:
@rhafer do we need a port to master?

s-t-e-v-e-n-k:
No, we don't need to port to master, this only landed for 4.0.

PR #345, opened 2019-01-31
haproxy: options re{p,q}_ssl_hello_type are deprecated

sjamgade:
use re{p,q}.ssl_hello_type correspondingly
https://www.haproxy.org/download/1.7/doc/configuration.txt

PR #346, opened 2019-02-01
crowbar-pacemaker: bump sync mark timeout to 300s

jgrassler:
We are currently seeing lots of sync mark timeouts in HA
enabled CI jobs. This commit bumps the sync mark timeout
beyond the maximum of how far the timeouts where observed to
be exceeded.

This should fix the timeout problems we are currently seeing with HA enabled mkcloud jobs.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
@@ -55,7 +55,7 @@ module CrowbarPacemakerSynchronization
   end
 
   # See "Synchronization helpers" documentation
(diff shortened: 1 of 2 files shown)


jgrassler:
> I'm generally against arbitrarily increasing timeouts unless the underlying reason for the current timeout being hit is well understood.

So am I, but I want to have this PR ready to go in case the candidates for fixing the root cause (https://github.com/crowbar/crowbar-openstack/pull/1993 and https://github.com/crowbar/crowbar-ha/pull/345 ) end up not doing the trick.

aspiers:
Thanks.  If those other fixes don't help and this is needed, the commit message will still need to be adjusted to cover the points I noted above.

dirkmueller:
I believe this is currently not needed, closing

PR #347, opened 2019-02-15
haproxy: improve galera HA setup (bsc#1122875)

aojea:
It turns out that haproxy has a known issue with concurrent connections
when using the sticky-table with size 1 as explained in this [link](https://bugzilla.redhat.com/show_bug.cgi?id=1211781)
Also, there is an scenario using haproxy with galera that keeps the connections 
open to a backend that's declared down. It happens when the health check
fails but the backend still keeps the port open.
This patch fixes both issues increasing the size table to 1000 and
adding an option to shut down the connections once a backend is marked
down.

Signed-off-by: aojeagarcia <aojeagarcia@suse.com>

### DIFF
--- a/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
+++ b/chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
@@ -91,7 +91,7 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
 
     <% end -%>
     <% if content[:stick] && content[:stick][:on] && !content[:stick][:on].empty? -%>
-	stick-table type ip size 1
+	stick-table type ip size 1000
 	stick on <%= content[:stick][:on] %>
 
     <% end -%>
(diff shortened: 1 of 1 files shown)


chef/cookbooks/haproxy/templates/default/haproxy.cfg.erb
```diff
@@ -128,8 +128,9 @@ listen  admin-stats <%= node[:haproxy][:stats][:enabled] ? node[:haproxy][:stats
         rise = " rise #{server[:rise] || 5}"
         fall = " fall #{server[:fall] || 2}"
         backup = server[:backup] ? " backup" : ""
+        onmarkdown = server[:onmarkdown] ? " on-marked-down shutdown-sessions" : ""
```
rhafer: I'd suggest to rename this to "on_marked_down_shutdown". Just for the sake for better understanding what it actually does.

We could also just add the possibility to allow appending arbitrary parameters instead. That might make future additions simpler:

```
server_options = server[:server_options].nil? ? " #{server[:server_options]}" : ""
...
server <%= server[:name] %> <%= server[:address] %>:<%= server[:port] %> .... <%= server_options %>
```

But I am fine with either way.

aojea: I'm not a fan of allowing such flexibility, this way we have control over the options and we can track and restrict which options are used and in which version

aojea:
I have to test it locally

PR #351, opened 2019-03-13
crowbar-pacemaker: bump sync mark timeout (SCRD-8179)

jgrassler:
We are currently seeing sync mark timeouts in Monasca enabled
HA clouds. The reason for these turns out to be the other
controller being a fairly consistent 70s behind the cluster
founder. This causes the sync marks to time out at a timeout
value of 60s. Further investigation didn't reveal any single
culprit (various things added a couple of seconds here and
there and in the end that just summed up to 70s).

This commit bumps the sync mark timeout to 120s to give us a
reasonable amount of breathing space. For we might be able to
shave off a few seconds here and there to get back below the
60s threshold, but that would be very vulnerable to sporadic
delays and/or additional small delays introduced by running
extra code on the cluster founder.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
@@ -55,7 +55,7 @@ module CrowbarPacemakerSynchronization
   end
 
   # See "Synchronization helpers" documentation
-  def self.wait_for_mark_from_founder(node, mark, fatal = false, timeout = 60)
+  def self.wait_for_mark_from_founder(node, mark, fatal = false, timeout = 120)
     return unless CrowbarPacemakerHelper.cluster_enabled?(node)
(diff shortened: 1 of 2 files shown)


aspiers:
@rhafer is right.  I have [a script for specifically tracking timings of sync marks](https://github.com/SUSE/supportconfig-utils/blob/master/plugins/SOC/bin/lnav-chef-sync-marks).  It sounds to me like there is a problem in the monasca recipes, or possibly in the cookbook immediately preceding it.  Maybe the sync marks there are wrong, or maybe they really do need a bigger timeout, but I doubt that increasing the global default is warranted.  Please ping me as I'm happy to provide advice and pair debug this.

witekest:
Created https://github.com/crowbar/crowbar-openstack/pull/2071 to bump the sync mark timeout only for problematic barclamps.

witekest:
https://github.com/crowbar/crowbar-openstack/pull/2071 increases timeouts in problematic barclamps, so this one can be closed.

PR #352, opened 2019-03-25
pacemaker: "disabled" stonith mode (SCRD-8462)

stefannica:
Allow "disabled" stonith mode to be configured in crowbar proposals.
This mode needs to be used by the Crowbar ECP CI as a temporary
alternative to SBD, which is not currently reliable.

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -536,6 +536,8 @@ class PacemakerService < ServiceObject
 
   def validate_proposal_stonith stonith_attributes, members
     case stonith_attributes["mode"]
+    when "disabled"
+      true # nothing to do
     when "manual"
       # nothing to do
     when "sbd"


stefannica:
> I'm wary of adding this to the product as is, because [SLE HA does not support clusters with STONITH disabled](https://www.suse.com/documentation/sle-ha-15/book_sleha_guide/data/sec_ha_fencing_recommend.html). Indeed, SBD might be unreliable but Pacemaker without any STONITH is basically guaranteed to be unreliable. At least you are not adding it to the web UI; that would not be acceptable without a clear warning.
> 
> Please can you elaborate on why SBD is not reliable? Perhaps we can help fix it instead. Another option is switching to IPMI.

@aspiers , all this happens in the context of deploying Crowbar in the engineering cloud. I cannot use the stonith `libvirt` mode which mkcloud uses, because I don't have access to the hypervisor, nor IPMI, because then I would need something like virtualBMC enabled on those ECP VMs and that would set me back a few weeks at least. SBD was my one shot at getting this to work cleanly, but in SBD mode I can never get the pacemaker barclamp to apply successfully in the first run - the first one or two attempts always run into syncmark failures. The only other people that have been testing SBD support are the QA team and according to @kbaikov , QA has been experiencing this same SBD issue so they have never been able to fully automate it because of this.

I do plan to create an SBD bug to track this, but first I need to understand the problem better, never having to work so closely with pacemaker and stonith myself. In the meantime, having the option to disable stonith gives me a cheap way out. ...Or so I thought. As it turns out, I'm running into the same type of syncmark issues even with stonith disabled, so now my only option is to deal with it.

jsuchome:
So this new SBD issue might not be related, but you know we have a code in mkcloud for enabling it, right? (`want_sbd`) It should install the right packages and create iscsi source + targets. It's possible that at least packages it was using before are not available or missing...

aspiers:
[**@stefannica**](https://github.com/stefannica) commented on [March 26, 2019 10:19 AM](https://github.com/crowbar/crowbar-ha/pull/352#issuecomment-476559947 "2019-03-26T10:19:30Z - Replied by Github Reply Comments"):
> @aspiers , all this happens in the context of deploying Crowbar in the engineering cloud.

Ah sorry, what a difference two letters make.  I missed "CI" in the description, so I thought you were doing this for ECP itself :man_facepalming:

> I cannot use the stonith `libvirt` mode which mkcloud uses, because I don't have access to the hypervisor, nor IPMI, because then I would need something like virtualBMC enabled on those ECP VMs and that would set me back a few weeks at least.

Right.

> SBD was my one shot at getting this to work cleanly

Actually not, there's another option.  I filed [Bug 1097803 – MR: support new fence_openstack.py fencing agent](https://bugzilla.suse.com/show_bug.cgi?id=1097803) *specifically* for this *exact* use case, as you can see in the description.

I appreciate that this would take longer to implement, so in the mean time we can work on fixing an existing approach, but long term I would like to see fencing via OpenStack supported and tested by our CI (although we would still want to test SBD in CI since it's one of the most common fencing agents customers use).

>but in SBD mode I can never get the pacemaker barclamp to apply successfully in the first run - the first one or two attempts always run into syncmark failures.

OK, so then I'm puzzled why you didn't ask for help?

> The only other people that have been testing SBD support are the QA team and according to @kbaikov , QA has been experiencing this same SBD issue so they have never been able to fully automate it because of this.

If multiple people are seeing it I'm really surprised this is the first time I'm hearing about it.  Again wondering why noone asked if I could help.  Of course I can't guarantee I would be able to solve it, but I'd have been happy to try.  And I'm sure I'm not the only one.

> I do plan to create an SBD bug to track this, but first I need to understand the problem better, never having to work so closely with pacemaker and stonith myself.

You don't necessarily need to understand it better; you can just ask for help.  Noone has to be an expert in everything ;-)

> In the meantime, having the option to disable stonith gives me a cheap way out.

Except it doesn't, because then the CI is testing a broken and unsupportable configuration.

> ...Or so I thought. As it turns out, I'm running into the same type of syncmark issues even with stonith disabled, so now my only option is to deal with it.

Yes please.  The timing here is bad, since I'm now officially on FTO with SUSECON next week.  I shouldn't even be looking at this now TBH :-(  But there are several others around with sync mark/SBD experience.

stefannica:
Abandoned in favor of #353

stefannica:
@aspiers thanks for the help. You were right, this was not the right way to solve my problem.

> 
> Actually not, there's another option. I filed [Bug 1097803 – MR: support new fence_openstack.py fencing agent](https://bugzilla.suse.com/show_bug.cgi?id=1097803) _specifically_ for this _exact_ use case, as you can see in the description.
> 
> I appreciate that this would take longer to implement, so in the mean time we can work on fixing an existing approach, but long term I would like to see fencing via OpenStack supported and tested by our CI (although we would still want to test SBD in CI since it's one of the most common fencing agents customers use).
> 

Many thanks, I wasn't aware of that. I created a [Jira card](https://jira.suse.de/browse/SCRD-8580) for this.

> 
> If multiple people are seeing it I'm really surprised this is the first time I'm hearing about it. Again wondering why noone asked if I could help. Of course I can't guarantee I would be able to solve it, but I'd have been happy to try. And I'm sure I'm not the only one.
> 

I don't know. Obviously, people are afraid of you :)

> > I do plan to create an SBD bug to track this, but first I need to understand the problem better, never having to work so closely with pacemaker and stonith myself.
> 
> You don't necessarily need to understand it better; you can just ask for help. Noone has to be an expert in everything ;-)
> 

I have. In fact, @skazi0 has been instrumental in helping me understand that the real issue wasn't with SBD, but rather a deadlock type situation that I'm trying to correct with #353 .

aspiers:
[**@stefannica**](https://github.com/stefannica) commented on [March 27, 2019 9:36 AM](https://github.com/crowbar/crowbar-ha/pull/352#issuecomment-477184547 "2019-03-27T14:36:12Z - Replied by Github Reply Comments"):
> > Actually not, there's another option. I filed [Bug 1097803 – MR: support new fence_openstack.py fencing agent](https://bugzilla.suse.com/show_bug.cgi?id=1097803) _specifically_ for this _exact_ use case, as you can see in the description.
> > 
> > I appreciate that this would take longer to implement, so in the mean time we can work on fixing an existing approach, but long term I would like to see fencing via OpenStack supported and tested by our CI (although we would still want to test SBD in CI since it's one of the most common fencing agents customers use).
> 
> Many thanks, I wasn't aware of that. I created a [Jira card](https://jira.suse.de/browse/SCRD-8580) for this.

Cool great idea - thanks!

> > If multiple people are seeing it I'm really surprised this is the first time I'm hearing about it. Again wondering why noone asked if I could help. Of course I can't guarantee I would be able to solve it, but I'd have been happy to try. And I'm sure I'm not the only one.
> > 
> 
> I don't know. Obviously, people are afraid of you :)

I know you were joking but I am actually a bit worried that I might be scaring people away a bit, maybe with my mailing list rants or something, I dunno ;-)  But if so, I would really hope that people would be pleasantly surprised when they work with me directly...

> > > I do plan to create an SBD bug to track this, but first I need to understand the problem better, never having to work so closely with pacemaker and stonith myself.
> > 
> > You don't necessarily need to understand it better; you can just ask for help. Noone has to be an expert in everything ;-)
> > 
> 
> I have. In fact, @skazi0 has been instrumental in helping me understand that the real issue wasn't with SBD, but rather a deadlock type situation that I'm trying to correct with #353 .

Great!  And as you saw, you even managed to find an issue in the logic which none of us had spotted before.

PR #353, opened 2019-03-27
pacemaker: don't check cluster members on founder and extend timeout for SBD case  (SCRD-8462)

stefannica:
When applying the pacemaker barclamp, the founder node first
(re)configures the cluster services on the local host and then
runs a [loop](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/pacemaker/recipes/default.rb#L76) waiting for two things to happen:
 - the cluster to be [online](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/pacemaker/recipes/default.rb#L97)
 - cluster membership to [reflect the nodes that were configured](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/pacemaker/recipes/default.rb#L100)
  (or updated) in the barclamp

This cluster check is not critical - it just results in a warning
being logged if the expected cluster state is not reached in the
configured amount of time (60 seconds).

While the founder node is waiting on the cluster to converge, the
non-founder nodes are stuck in the chef compile phase, also running
[a time-boxed while-loop](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/crowbar-pacemaker/recipes/default.rb#L44), waiting for the founder node to complete
the setup. Given that this happens in the compile phase, these nodes
haven't yet had a chance to apply the cluster configuration locally,
which means that the founder cluster check will always take the full
60/120 seconds to run and eventually fail.

The same cluster check that is executed on the founder node is
also executed on the non-founder nodes during convergence phase.
However, the difference is that the cluster configuration is actually
created/updated before this happens.

This commit changes the founder cluster check logic so that the
cluster membership part is only run on the non-founder nodes.
This way, the convergence loop run on the founder node has a chance
to complete ahead of the 60 seconds timeout, therefore reducing
the runtime and decreasing the chance that the non-founder nodes
will timeout while waiting for the founder node to complete cluster
setup.

The second change that this PR makes explicitly targets the SBD mode.
When SBD is configured for STONITH, the founder node needs to
wait for the cluster to come online _twice_: one time after the
initial corosync setup and [a second time](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/pacemaker/recipes/sbd.rb#L114) after configuring
the SBD service. This means that the cluster check mentioned
above is actually executed _twice_. Because of this, non-founder
nodes also need to give the founder node more time to complete
its setup before they time out.

These changes are required for the Crowbar ECP deployments 
(SCRD-8462), which otherwise fail too often because the non-founder
nodes run into timeouts.

### DIFF
--- a/chef/cookbooks/pacemaker/recipes/default.rb
+++ b/chef/cookbooks/pacemaker/recipes/default.rb
@@ -66,6 +66,7 @@ end
 # Extending this timeout unconditionally would cause a deadlock with the "Waiting for
 # cluster founder to be set up" loop in crowbar-pacemaker cookbook.
 online_timeout = node.fetch("crowbar_wall", {})[:cluster_node_added] ? 120 : 60
+is_founder = node[:pacemaker][:founder] == node[:fqdn]
 
 ruby_block "wait for cluster to be online" do
   block do
@@ -90,13 +91,13 @@ ruby_block "wait for cluster to be online" do
           end
           crm_mon = crm_mon_cmd.stdout
           ready = true
-          if !crm_mon.include?("#{cluster_size} nodes configured")
+          if !is_founder && !crm_mon.include?("#{cluster_size} nodes configured")
             ready = false
             Chef::Log.warn("cluster doesn't have #{cluster_size} nodes configured yet")
           elsif !crm_mon.include?("Online:")
             ready = false
             Chef::Log.warn("cluster doesn't have nodes online yet")
-          elsif crm_names.sort != nodes_names.sort
+          elsif !is_founder && crm_names.sort != nodes_names.sort
             ready = false
             Chef::Log.warn("crm_node -l listed nodes #{crm_names.sort}, " \
                            "not #{nodes_names.sort} as expected")
--- a/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
@@ -30,6 +30,11 @@ end
 include_recipe "crowbar-pacemaker::quorum_policy"
 include_recipe "crowbar-pacemaker::stonith"
 
+# If SBD is configured for STONITH, the founder node needs to restart the
+# corosync service and wait for the cluster to come online one more time
+# before it's fully set up. That may take an additional 60 seconds.
+founder_timeout = node[:pacemaker][:stonith][:mode] == "sbd" ? 180 : 120
+
 # We let the founder go first, so it can generate the authkey and some other
 # initial pacemaker configuration bits; we do it in the compile phase of the
 # chef run because the non-founder nodes will look during the compile phase for
@@ -39,7 +44,7 @@ unless CrowbarPacemakerHelper.is_cluster_founder?(node)
   begin
     # we use a long timeout because the wait / attribute-set are in two different
     # phases, and this wait is fatal in case of errors
-    Timeout.timeout(120) do
+    Timeout.timeout(founder_timeout) do
       Chef::Log.info("Waiting for cluster founder to be set up...")
       loop do
         founder = CrowbarPacemakerHelper.cluster_founder(node)


chef/cookbooks/pacemaker/recipes/default.rb
```diff
@@ -66,6 +66,7 @@
 # Extending this timeout unconditionally would cause a deadlock with the "Waiting for
 # cluster founder to be set up" loop in crowbar-pacemaker cookbook.
 online_timeout = node.fetch("crowbar_wall", {})[:cluster_node_added] ? 120 : 60
+is_founder = node[:pacemaker][:founder]
```
jsuchome: I don't believe this is a boolean...

stefannica: Right you are :). Done

chef/cookbooks/pacemaker/recipes/default.rb
```diff
@@ -66,6 +66,7 @@
 # Extending this timeout unconditionally would cause a deadlock with the "Waiting for
 # cluster founder to be set up" loop in crowbar-pacemaker cookbook.
 online_timeout = node.fetch("crowbar_wall", {})[:cluster_node_added] ? 120 : 60
+is_founder = !node[:pacemaker][:founder].nil?
```
jsuchome: Sorry, still wrong :-)
`node[:pacemaker][:founder]` is the string value that points to the founder, and is present on all cluster nodes.

I think you need `is_founder = node[:pacemaker][:founder] == node[:fqdn]`

skazi0:
@stefannica the commit/PR message is not fully correct. The first section sounds like this code is meant to be used only on the founder which is not true. Especially the 120s timeout is (almost?) never for the founder but rather for the newly added non-founder node.

chef/cookbooks/pacemaker/recipes/default.rb
stefannica: In my defense, I was confused by this line: https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/pacemaker/recipes/default.rb#L116 
Should I be worried that the `:founder` attribute is not even set ?

jsuchome: Good point. It might be needed to investigate if that line is there on purpose (= all pacemaker cluster members) or it was forgotten when the meaning of `node[:pacemaker][:founder]` was chaned some time ago (more likely)

jsuchome: `CrowbarPacemakerHelper.is_cluster_founder?(node)` seems to be the right answer.

I'm still curious of that include of `pacemaker::setup` is wrong for non-founder nodes...

stefannica:
> @stefannica the commit/PR message is not fully correct. The first section sounds like this code is meant to be used only on the founder which is not true. Especially the 120s timeout is (almost?) never for the founder but rather for the newly added non-founder node.

@skazi0 I think the description is pretty accurate. I added a section to mention that the code is also running on the non-founder nodes and also included some code links, but I really don't want to write a complete description of the pacemaker configuration synchronization strategy.

chef/cookbooks/pacemaker/recipes/default.rb
stefannica: Great, I used `CrowbarPacemakerHelper.is_cluster_founder?(node)`. That does indeed seem to be the correct way to check for the founder node, even in the pacemaker recipes.

vuntz:
@stefannica can we have a quick mumble about it tomorrow? I can possibly help with the history of this, but a discussion would refresh my memory faster.

skazi0:
> > @stefannica the commit/PR message is not fully correct. The first section sounds like this code is meant to be used only on the founder which is not true. Especially the 120s timeout is (almost?) never for the founder but rather for the newly added non-founder node.
> 
> @skazi0 I think the description is pretty accurate. I added a section to mention that the code is also running on the non-founder nodes and also included some code links, but I really don't want to write a complete description of the pacemaker configuration synchronization strategy.

I didn't want you to add more details but rather correct those which are not fully true (e.g. "120 seconds if the founder node is newly added to the cluster").

chef/cookbooks/pacemaker/recipes/default.rb
skazi0: @jsuchome https://github.com/crowbar/crowbar-ha/commit/6df9deda7d198eaa6a744c223679cb3388c5fd1e this answers some of the questions. 
Running the "setup" recipe multiple times doesn't seem to hurt the cluster either at this early stage. I'm not sure about the cluster extension case.

jsuchome: Well, in that original code, `[:pacemaker][:founder]` was really indicating the founder, so maybe we should change that line for including `setup` adapt as well?

(That said, it's not related to this PR)

stefannica:
> > > @stefannica the commit/PR message is not fully correct. The first section sounds like this code is meant to be used only on the founder which is not true. Especially the 120s timeout is (almost?) never for the founder but rather for the newly added non-founder node.
> > 
> > 
> > @skazi0 I think the description is pretty accurate. I added a section to mention that the code is also running on the non-founder nodes and also included some code links, but I really don't want to write a complete description of the pacemaker configuration synchronization strategy.
> 
> I didn't want you to add more details but rather correct those which are not fully true (e.g. "120 seconds if the founder node is newly added to the cluster").

Ah, I understand now what you meant. I fixed the commit message / PR description.

kbaikov:
Just tested this patch. Pacemaker applied without timeout with the first attempt.

stefannica:
Had to push one more fix to extend the timeout on the non-founder nodes side when SBD is enabled, because the founder node actually runs the cluster check twice when SBD mode is used for stonith.

jsuchome:
Just to make sure - you are testing the case where you're adding a new node to the cluster, right?

stefannica:
> Just to make sure - you are testing the case where you're adding a new node to the cluster, right?

Haven't done that yet, but planning to, yes, hence the "do not merge yet" label. :)

stefannica:
[My Crowbar HA ECP build](https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/137/pipeline/) has finally passed with these changes applied.

kbaikov:
> Just to make sure - you are testing the case where you're adding a new node to the cluster, right?

retested the patch by adding a node to the cluster. All was successful. The scenario steps used:

1. install database and rabbitmq on a dedicated node or dedicated cluster

2. install other openstack services on another dedicated cluster like "Controller" in which members are controller01 and controller02.

   Drbd disabled on this cluster.

3. bootstrap a new fresh node named with like controller3.

4. add this new node to pacemaker cluster of "Controller" and apply

stefannica:
> Just to make sure - you are testing the case where you're adding a new node to the cluster, right?

I also tested the "add node to existing cluster" scenario mentioned in [this bug](https://bugzilla.suse.com/show_bug.cgi?id=1083427), using the SBD mode and an ECP Crowbar cloud environment. It was somewhat painful, because I had to re-apply the pacemaker barclamp a few times to get all the OpenStack services to deploy properly, but it turned out ok in the end.

aspiers:
Great work, I'm +1 if @vuntz is.

PR #354, opened 2019-04-04
[5.0] pacemaker: don't check cluster members on founder and extend timeout for SBD case (SCRD-8462)

stefannica:
(backports #353)

When applying the pacemaker barclamp, the founder node first
(re)configures the cluster services on the local host and then
runs a [loop](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/pacemaker/recipes/default.rb#L76) waiting for two things to happen:
 - the cluster to be [online](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/pacemaker/recipes/default.rb#L97)
 - cluster membership to [reflect the nodes that were configured](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/pacemaker/recipes/default.rb#L100)
  (or updated) in the barclamp

This cluster check is not critical - it just results in a warning
being logged if the expected cluster state is not reached in the
configured amount of time (60 seconds).

While the founder node is waiting on the cluster to converge, the
non-founder nodes are stuck in the chef compile phase, also running
[a time-boxed while-loop](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/crowbar-pacemaker/recipes/default.rb#L44), waiting for the founder node to complete
the setup. Given that this happens in the compile phase, these nodes
haven't yet had a chance to apply the cluster configuration locally,
which means that the founder cluster check will always take the full
60/120 seconds to run and eventually fail.

The same cluster check that is executed on the founder node is
also executed on the non-founder nodes during convergence phase.
However, the difference is that the cluster configuration is actually
created/updated before this happens.

This commit changes the founder cluster check logic so that the
cluster membership part is only run on the non-founder nodes.
This way, the convergence loop run on the founder node has a chance
to complete ahead of the 60 seconds timeout, therefore reducing
the runtime and decreasing the chance that the non-founder nodes
will timeout while waiting for the founder node to complete cluster
setup.

The second change that this PR makes explicitly targets the SBD mode.
When SBD is configured for STONITH, the founder node needs to
wait for the cluster to come online _twice_: one time after the
initial corosync setup and [a second time](https://github.com/crowbar/crowbar-ha/blob/1b749ef337fa3ba935757a142de3bfea94169107/chef/cookbooks/pacemaker/recipes/sbd.rb#L114) after configuring
the SBD service. This means that the cluster check mentioned
above is actually executed _twice_. Because of this, non-founder
nodes also need to give the founder node more time to complete
its setup before they time out.

These changes are required for the Crowbar ECP deployments 
(SCRD-8462), which otherwise fail too often because the non-founder
nodes run into timeouts.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/default.rb
@@ -30,6 +30,11 @@ end
 include_recipe "crowbar-pacemaker::quorum_policy"
 include_recipe "crowbar-pacemaker::stonith"
 
+# If SBD is configured for STONITH, the founder node needs to restart the
+# corosync service and wait for the cluster to come online one more time
+# before it's fully set up. That may take an additional 60 seconds.
+founder_timeout = node[:pacemaker][:stonith][:mode] == "sbd" ? 180 : 120
+
 # We let the founder go first, so it can generate the authkey and some other
 # initial pacemaker configuration bits; we do it in the compile phase of the
 # chef run because the non-founder nodes will look during the compile phase for
@@ -39,7 +44,7 @@ unless CrowbarPacemakerHelper.is_cluster_founder?(node)
   begin
     # we use a long timeout because the wait / attribute-set are in two different
     # phases, and this wait is fatal in case of errors
-    Timeout.timeout(120) do
+    Timeout.timeout(founder_timeout) do
       Chef::Log.info("Waiting for cluster founder to be set up...")
       loop do
         founder = CrowbarPacemakerHelper.cluster_founder(node)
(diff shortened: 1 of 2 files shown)


stefannica:
Testing with an ECP Crowbar CI HA setup: https://ci.suse.de/blue/organizations/jenkins/cloud-crowbar8-job-ha-x86_64/detail/cloud-crowbar8-job-ha-x86_64/3/pipeline

stefannica:
> Testing with an ECP Crowbar CI HA setup: https://ci.suse.de/blue/organizations/jenkins/cloud-crowbar8-job-ha-x86_64/detail/cloud-crowbar8-job-ha-x86_64/3/pipeline

Previous job failed because the ECP Crowbar scenario file was incorrect. Re-testing with: https://ci.suse.de/blue/organizations/jenkins/cloud-crowbar8-job-ha-x86_64/detail/cloud-crowbar8-job-ha-x86_64/6/pipeline

PR #357, opened 2019-05-09
temp debug for sync_marks

skazi0:


aspiers:
@skazi0 Just checking you know about https://github.com/SUSE/supportconfig-utils/blob/master/plugins/SOC/bin/lnav-chef-sync-marks ?  I appreciate that it won't replace what you are doing here, but maybe a combination of the two techniques would help.

skazi0:
@aspiers thanks! I will check that for sure. My problem is with resetting syncmarks which doesn't happen sometimes so I need some detailed debug info on what's going on inside.

PR #358, opened 2019-05-30
pacemaker: add failure nodes to sync fail message (bsc#1083721)

matthewoliver:

issue.

(cherry picked from commit 90bf282f955ed7ceb34afe2d65f189b6b1ceb1a9)


nicolasbock:
I restarted the CI job

toabctl:
CI failure is unrelated. merging...

PR #359, opened 2019-05-30
[4.0] pacemaker: add failure nodes to sync fail message (bsc#1083721)

matthewoliver:

issue.

(cherry picked from commit 90bf282f955ed7ceb34afe2d65f189b6b1ceb1a9)

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
+++ b/chef/cookbooks/crowbar-pacemaker/libraries/synchronization.rb
@@ -64,28 +64,36 @@ module CrowbarPacemakerSynchronization
       return
     end
 
-    Chef::Log.info("Checking if cluster founder has set #{mark}...")
-
     founder_name = CrowbarPacemakerHelper.cluster_founder_name(node)
+    cluster_name = CrowbarPacemakerHelper.cluster_name(node)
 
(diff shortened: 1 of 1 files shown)


AbelNavarro (approved):

aspiers (approved):
The CI failure looks unrelated, so assuming that this is a straight cherry-pick, I'm +1 (actually +10000) for backporting this significant UX improvement, which should help reduce the most common FAQ we get from customers.

nicolasbock:
I restarted the CI job

AbelNavarro:
I also just restarted the CI job. The error looked unrelated, but you could actually see the extended debug info in place which is useful. Please don't let this backport die.

aspiers:
I guess this is the extract from the CI failures @AbelNavarro was talking about (or one of them at least), which proves this is working:

```
[2019-06-07T13:29:43+00:00] FATAL: Cluster founder d52-54-77-77-01-01 didn't set keystone_certificates on cluster data! Timed out while waiting for the founder; please check either '/var/log/crowbar/chef-client/d52-54-77-77-01-01.vm1.cloud.suse.de.log' on the admin node or '/var/log/chef/client.log' on d52-54-77-77-01-01 to see what went wrong.
```

I suggest we wait for one more reviewer to approve and then merge.

JanZerebecki:
This branch should now only receive important changes. This looks like it is not. Please reopen if you feel otherwise.

PR #360, opened 2019-06-13
Make default sync_mark timeout configurable + scalable

skazi0:
This change introduces a new databag item "sync_mark" in "crowbar-config"
databag. It can contain "default_timeout" value and that will be used if
available. Otherwise previous default of 60s will be used.

This is useful in environments where node performance is not consistent.
Such cases were observed in virtualized deployments running on top of cloud
which hosted various other workloads and sometimes non-founders were much
faster than the founder. This lead to situation where some non-founder can
leave "sync window" (with timeout) before the founder even reaches the
"wait" sync_mark.

Another reason for such "inhomogeneous" performance could be (semi)broken
hardware (e.g. RAID cache batteries).

Increasing the default sync_mark timeout in above cases is a workaround
which will "relax" the timing constrains and could possibly make more
of those deployments successfull.

----------------------------------------------------------------

Configurable default timeout value doesn't help in cases where
timeout is cusomized for given resource. Such cases are better covered
with scaling of all timeouts with a multiplier.
E.g.
base:
some_resource/timeout = 60
customized_resource/timeout = 100

with default_timeout = 120:
some_resource/timeout = 120
customized_resource/timeout = 100

with timeout_multipler = 2:
some_resource/timeout = 120
customized_resource/timeout = 200

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/providers/sync_mark.rb
+++ b/chef/cookbooks/crowbar-pacemaker/providers/sync_mark.rb
@@ -15,6 +15,12 @@
 #
 
 def get_options resource
+  sync_mark_config = begin
+    Chef::DataBagItem.load("crowbar-config", "sync_mark")
+  rescue Net::HTTPServerException
+    {}
+  end
+  timeout = sync_mark_config.fetch("default_timeout", 60)
   action = nil
   mark = nil
 
@@ -33,15 +39,21 @@ def get_options resource
     mark = new_resource.mark
   end
 
+  unless new_resource.timeout.nil?
+    timeout = new_resource.timeout
+  end
+
   raise "Missing mark attribute" if mark.nil?
 
-  [action, mark]
+  Chef::Log.info("Using timeout #{timeout} for sync_mark #{mark}")
+
+  [action, mark, timeout]
 end
 
 action :wait do
-  _, mark = get_options(new_resource)
+  _, mark, timeout = get_options(new_resource)
   CrowbarPacemakerSynchronization.wait_for_mark_from_founder(
-    node, mark, new_resource.fatal, new_resource.timeout
+    node, mark, new_resource.fatal, timeout
   )
 end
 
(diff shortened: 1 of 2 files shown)


AbelNavarro:
I think we are all hitting timeouts on sync-marks in our developer environments. I was fixing it for one particular sync-mark that was giving me headaches (https://github.com/crowbar/crowbar-openstack/pull/2150) but maybe your approach is even better.

skazi0:
@aspiers Of course, there you are: ECP has a mix of slow and faster hardware. Slowness comes mostly from broken batteries in RAID controllers (root cause, fixing ongoing for past 1,5 years) but also from varying workloads. The spread of I/O performance is at least x5 between the slowest and and the fastest host.

Our sync_mark approach is based largely on the assumption that cluster members are similar in terms of performance. With the mix we're given, what often happens is that some of the non-founders are faster than the founder and are already at the end of "sync window" before the founder even gets to the "wait" sync_mark. The host performance can vary throughout the deployment process so the timeouts are happening randomly at various places (i.e. not possible to fix by increasing specific timeout).

The parameter I propose here is supposed to "relax" the sync logic to cover for such "inhomogeanous" cases.

aspiers:
[**@skazi0**](https://github.com/skazi0) commented on [June 13, 2019 5:16 PM](https://github.com/crowbar/crowbar-ha/pull/360#issuecomment-501772629 "2019-06-13T16:16:47Z - Replied by Github Reply Comments"):
> @aspiers Of course, there you are: ECP has a mix of slow and faster hardware. Slowness comes mostly from broken batteries in RAID controllers (root cause, fixing ongoing for past 1,5 years) but also from varying workloads. The spread of I/O performance is at least x5 between the slowest and and the fastest host.

This is a very good point, so please put it in the commit message.  "In some environments it is useful to ..." is a very vague and unconvincing motivation, but this is much better :-)

> Our sync_mark approach is based largely on the assumption that cluster members are similar in terms of performance. With the mix we're given, what often happens is that some of the non-founders are faster than the founder and are already at the end of "sync window" before the founder even gets to the "wait" sync_mark.

Sure.  It's not just that the non-founders are faster, but also that the founder typically has much more to do.

> The host performance can vary throughout the deployment process so the timeouts are happening randomly at various places (i.e. not possible to fix by increasing specific timeout).
> 
> The parameter I propose here is supposed to "relax" the sync logic to cover for such "inhomogeanous" cases.

Yes this makes sense.  Please put it in the commit message and I'll +1, although I would like to see some prominent caveat displayed around this option warning that it should generally only be used to deal with slower hardware.

skazi0:
For our specific case, the `timeout_multiplier` could be enough. Shall I remove the `default_timeout` and only keep the multiplier? wdyt?

skazi0:
@AbelNavarro the crowbar-core PR was updated but I squashed the commits to avoid migration problems if someone picks only one of them.

PR #363, opened 2019-08-20
corosync: Hardcode ruby version for package installation (SOC-10010)

rsalevsky:
Sometimes there is a race condition and ohai didn't collect the ruby
version. to_f evalutes then the version to 0.0 and zypper fails to
install the rubygem `ruby0.0-rubygem-ruby-shadow' not found in package
names`.

### DIFF
--- a/chef/cookbooks/corosync/recipes/service.rb
+++ b/chef/cookbooks/corosync/recipes/service.rb
@@ -60,7 +60,7 @@ end
 # chef-client can only use it immediately if we install it at
 # recipe compile-time, not run-time:
 # from the next run onwards:
-rubygem_ruby_shadow = "ruby#{node["languages"]["ruby"]["version"].to_f}-rubygem-ruby-shadow"
+rubygem_ruby_shadow = "ruby2.1-rubygem-ruby-shadow"
 pkg = package rubygem_ruby_shadow do
   action :nothing
 end


chef/cookbooks/corosync/recipes/service.rb
```diff
@@ -60,7 +60,7 @@
 # chef-client can only use it immediately if we install it at
 # recipe compile-time, not run-time:
 # from the next run onwards:
-rubygem_ruby_shadow = "ruby#{node["languages"]["ruby"]["version"].to_f}-rubygem-ruby-shadow"
+rubygem_ruby_shadow = "ruby2.1-rubygem-ruby-shadow"
```
nicolasbock: Is there no way to fix the race condition instead? By introducing a hard coded version string here, we add tech debt, i.e. a future ruby upgrade would require a code change.

rsalevsky: I didn't invest much time in it as 1) 98% are already hard coded (only 2 packages aren't) and 2) the version will not change in the lifetime of crowbar. So I opted for the quick and dirty fix to make the CI a bit more stable.

nicolasbock: Ah ok, that makes sense. Thanks for the update!

PR #368, opened 2019-09-11
Sync_mark workaround (SOC-10297)

skazi0:
**NOT TESTED!!!**

If crm_attribute from node hits CIB at the same time when other node changes CIB,
the attribute change could get lost and founder will time out waiting for the mark.
Re-setting the mark periodically (while waiting for the founder to detect marks from
all nodes) is a quick workaround for this problem.


stefannica:
I created a periodic job running daily to test this patch: https://ci.suse.de/job/cloud-crowbar9-job-gate-ha-deploy-sync-mark-fix-x86_64/

skazi0:
As the problem turned out to be a regression in `crmsh` package, this workaround is not needed anymore.

PR #369, opened 2019-11-15
[5.0] Drop g-haproxy location before group deletion (bsc#1156914)

skazi0:
When container (group) resource is deleted, it's children are not
and all constraints assigned to the container are re-assigned to one
of the children. In g-haproxy case this leads to incorrect situation
where l-g-haproxy-controller constraint is assigned to
vip-admin-cluster-data.
Removing the constraint before the group should avoid such situations.

### DIFF
--- a/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
+++ b/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
@@ -74,6 +74,12 @@ if node[:pacemaker][:haproxy][:clusters].key?(cluster_name) && node[:pacemaker][
   # Compatibility with existing deployment: we need to drop the group to create
   # the clone
   group_name = "g-#{service_name}"
+  # drop location constraint first as it would get reassigned to some child of the group
+  # otherwise. See: https://github.com/ClusterLabs/crmsh/issues/140
+  pacemaker_location openstack_pacemaker_controller_only_location_for group_name do
+    action :delete
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
   pacemaker_group group_name do
     action [:stop, :delete]
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }


chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
```diff
@@ -74,6 +74,12 @@
   # Compatibility with existing deployment: we need to drop the group to create
   # the clone
   group_name = "g-#{service_name}"
+  # drop location constraint too as it would get reassigned to some child of the group
```
jsuchome: What about _drop location constraint first_ ?

chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
```diff
@@ -74,6 +74,12 @@
...
+  # otherwise.
+  pacemaker_location openstack_pacemaker_controller_only_location_for group_name do
+    action :delete
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
jsuchome: + `only_if "crm configure show location_name"` ?

skazi0: I believe it is not needed here. The `only_if` is for `:stop` in the group resource. See this: https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/pacemaker/providers/location.rb#L41 and this https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/pacemaker/libraries/chef/mixin/pacemaker/runnable_resource.rb#L27 (group uses runnable_resource)

skazi0:
tested during 7-8 upgrade and it seems to do the trick :o)

chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb
jsuchome: good point

PR #370, opened 2019-11-18
Drop g-haproxy location before group deletion (bsc#1156914)

skazi0:
When container (group) resource is deleted, it's children are not
and all constraints assigned to the container are re-assigned to one
of the children. In g-haproxy case this leads to incorrect situation
where l-g-haproxy-controller constraint is assigned to
vip-admin-cluster-data.
Removing the constraint before the group should avoid such situations.

(cherry picked from commit 8540c40fbbd1f67593b467b94aa11cd6f43fcfdd)

port of #369


jsuchome:
The code is correct, I just wonder if it is really needed in SOC9? Could g-haproxy even exist in such deployment?

skazi0:
@jsuchome The group deletion code is still there so we either add this or remove the deletion code.

jsuchome:
I'd prefer dropping the delete group part. There's no way it should exist running this code.

skazi0:
Closed in favor of https://github.com/crowbar/crowbar-ha/pull/371

PR #372, opened 2020-01-10
[4.0] add ssl termination on haproxy (bsc#1149535)

bbobrov:
If the requests are passed in tcp mode, the source ip gets replaced with
the one of the node where haproxy lives, and there is no way to get the
original ip on the services side.

Add an ability to terminate ssl on haproxy. Using `terminate_ssl` and
`pemfile` will make haproxy switch to http mode and make plain https
requests to the service.


bbobrov:
> Why not submit this to master first and then backport?

Because the customer who encountered the issue on soc7 and the solution grew to fit what they had. Do you think i should submit to master first?

PR #376, opened 2020-08-27
pacemaker: remove node on delete (SOC-11240)

sandonovsuse:
On node delete, crowbar doesn't execute pacemaker barclamp due to
default pacemaker proposal not having transitions attribute enabled
for delete. This change adds said feature and the delete transition.

### DIFF
--- a/crowbar_framework/app/models/pacemaker_service.rb
+++ b/crowbar_framework/app/models/pacemaker_service.rb
@@ -955,6 +955,34 @@ class PacemakerService < ServiceObject
     end
   end
 
+  def transition(inst, name, state)
+    @logger.debug("Pacemaker transition: entering: #{name} for #{state}")
+
+    if state == "delete"
+      proposal = Proposal.find_by(barclamp: @bc_name, name: inst)
+      stonith_mode = proposal["attributes"]["pacemaker"]["stonith"]["mode"]
+
+      # remove node from stonith cfg
+      proposal["attributes"]["pacemaker"]["stonith"][
+        stonith_mode]["nodes"].delete(name)
+      if proposal["attributes"]["pacemaker"]["stonith"]["per_node"]["nodes"]
+        proposal["attributes"]["pacemaker"]["stonith"][
+          "per_node"]["nodes"].delete(name)
+      end
+
+      # remove from pacemaker-cluster-member
+      proposal["deployment"]["pacemaker"]["elements"][
+        "pacemaker-cluster-member"].delete(name)
+
+      # Save and commit
+      proposal.save
+      proposal_commit(inst, in_queue: false, validate: false, validate_after_save: false)
+    end
+
+    @logger.debug("Pacemaker transition: exiting: #{name} for #{state}")
+    [200, { name: name }]
+  end
+
   private
 
   def pacemaker_node_name(node, remotes)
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/models/pacemaker_service.rb
```diff
@@ -955,6 +955,36 @@ def prepare_stonith_attributes(role_attributes, remote_nodes, member_nodes, remo
...
 
+  def transition(inst, name, state)
+    @logger.debug("Pacemaker transition: entering: #{name} for #{state}")
+
+    if state == "delete"
+      proposal = Proposal.find_by(barclamp: @bc_name, name: inst)
+      stonith_mode = proposal["attributes"]["pacemaker"]["stonith"]["mode"]
+
+      # remove node from stonith cfg
+      [stonith_mode, "per_node"].each do |key|
```
skazi0: Why do we need it for current mode AND "per_node"? I have SBD STONITH in my env and I don't have "per_node" under "stonith".
Even if we decide to keep it, it should be guarded in case the key doesn't exist.

skazi0 (changes requested):

crowbar_framework/app/models/pacemaker_service.rb
sandonovsuse: Agreed, I had in it my env, didn't think it might be optional. Thanks

skazi0 (dismissed):
Code looks good but I'm not 100% sure if it's safe to commit from inside the transition.

jgrassler:
Thanks for pointing out the "Forget" operation. It's behavior does not appear to be affected though: with this patch in place, all occurences of the forgotten node are purged from the Pacemaker proposal upon saving it post-forget. But the same thing happens without this patch in place. So where _does_ this come into play/what is the expected change in behavior?

skazi0:
@jgrassler the forgotten nodes are removed from the list by UI/CLI and this would be persisted if pacemaker proposal was saved post-forget. The problem here is that if you just "forget" and leave the cloud running, at some point the periodic chef-client runs will start to fail because there is no new founder elected. This change (together with the crowbar-core one) is supposed to make this "save & commit" step trigger automatically after "forget".

skazi0:
@sandonovsuse I think I know why it didn't work for @jgrassler. I also had similar results in my tests. The problem is that you changed proposal schema but didn't add a migration for this. You probably tested with new proposal (after adding your changes) but it would not work for existing proposals.

This should make it usable for existing proposals:
```
diff --git a/chef/data_bags/crowbar/migrate/pacemaker/302_add_delete_transition.rb b/chef/data_bags/crowbar/migrate/pacemaker/302_add_delete_transition.rb
new file mode 100644
index 0000000..f2f0a4d
--- /dev/null
+++ b/chef/data_bags/crowbar/migrate/pacemaker/302_add_delete_transition.rb
@@ -0,0 +1,11 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  deployment["config"]["transition_list"] = template_deployment["config"]["transition_list"]
+  deployment["config"]["transitions"] = template_deployment["config"]["transitions"]
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  deployment["config"]["transition_list"] = template_deployment["config"]["transition_list"]
+  deployment["config"]["transitions"] = template_deployment["config"]["transitions"]
+  return attrs, deployment
+end
diff --git a/chef/data_bags/crowbar/template-pacemaker.json b/chef/data_bags/crowbar/template-pacemaker.json
index 668f102..ca06cbb 100644
--- a/chef/data_bags/crowbar/template-pacemaker.json
+++ b/chef/data_bags/crowbar/template-pacemaker.json
@@ -65,7 +65,7 @@
     "pacemaker": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 301,
+      "schema-revision": 302,
       "element_states": {
         "pacemaker-cluster-member"    : [ "readying", "ready", "applying" ],
         "hawk-server"

